PR #5, opened 2017-12-07
Initialize repo

mgwalker:
Initialize the repo with some stuff:

- Standard 18F files
  - [x] contributing guide
  - [x] license doc/waiver
  - [x] project readme

- Continuous integration config
  - [x] CircleCI configuration
  - [x] CodeClimate configuration
  - [x] CodeCov configuration (handled by CircleCI)

- Web app setup
  - [x] Webpack setup
  - [x] Test setup
  - [x] Basic 3rd-party libraries (React, etc.)

- Docker
  - [x] Dockerfile for web app
  - [x] docker-compose file for whole project (though it doesn't have anything for the API or database yet)

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -1,0 +1,58 @@
+version: 2
+jobs:
+  prepare web dependencies cache:
+    docker:
+      - image: node:8
+    steps:
+      - checkout
+      - restore_cache:
+          key: v1-cms-hitech-apd-{{ checksum "web/package.json" }}
+      - run:
+          name: install dependencies
+          working_directory: ~/project/web
+          command: npm install
+      - save_cache:
+          key: v1-cms-hitech-apd-{{ checksum "web/package.json" }}
+          paths:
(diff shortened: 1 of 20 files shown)


mgwalker:
I put in that `prepare web dependencies cache` job before the lint and test jobs so we wouldn't have two jobs installing dependencies, but it's kind of ugly having it show up in Github like that.  I'd hoped it wouldn't show up...  Now I'm thinking we should just let CircleCI bear the cost of duplicating dependency installs.

We could also orchestrate the workflow so that `test` comes after `lint` and get the cacheing setup that way.  We'd just need to have both jobs always run, otherwise we might not know our tests' status if there's a code linting error, which seems less than ideal.  😛

brendansudol:
fwiw, i'm totally fine with the dependency cache step showing up there -- the more green checkmarks, the better :)

brendansudol:
this is 💯 - thanks @mgwalker

PR #6, opened 2017-12-11
add (and run) prettier

brendansudol:
this PR adds [prettier](https://github.com/prettier/prettier), and subsequently runs prettier across all js files

prettier is a code formatter than ensures consistency and standardization around things like putting a semi-colon at the end of statements, proper indentation, maximum line width.

it can be run across all files via `npm run prettier`; you can also add a prettier extension to your code editor of choice (atom, sublime, etc) so that it auto-formats the file you're working on when you save (optional, though encouraged).

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,1 +1,7 @@
+# dependencies
 node_modules
+
+# misc
+.DS_Store
+
+npm-debug.log*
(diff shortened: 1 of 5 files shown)


brendansudol:
haha guilty! it just makes my life a little bit easier and better :)

Issue #35, opened 2017-12-15
Add token mechanism


mgwalker:
Some kind of bearer token, probably.  Maybe JWT?

mgwalker:
This might be fully covered by #47.  Up for discussion.

#47 uses client sessions, where an authenticated user is serialized and then stored in an encrypted cookie.  In this case, the user is serialized as simply their ID.  When the user requests an API resource, the API receives back the encrypted cookie.  If the API can decrypt the cookie, then we assume it has not been tampered with.  Then we deserialize it, reconstructing the user ID back into a full user object.

brendansudol:
Yep, this is all set -- nice work, @mgwalker

Issue #33, opened 2017-12-15
Add API endpoints to manage local users


mgwalker:
Related to #31.  These should be simple for now since they may not survive long

PR #47, opened 2017-12-16
Authentication scaffolding

### LINKED ISSUE #30: Add passport to the API server

mgwalker:
- fixes a bug in the way default environment variable values were being set
- adds Passport for authentication
  - sets up a local strategy
  - sets up a hard-coded username/password (`hello`/`world`) to get started
  - stores serialized auth info in an encrypted client cookie
  - deserializes auth info before processing API endpoint requests
- adds an API endpoint at `POST: /auth/login` that takes a username and password as form-encoded variables and returns the generated user ID
  - **note**: this isn't strictly necessary; the status code should be sufficient. If the user is logged in, their session cookie will be set and there's no reason to send their ID.  This was mostly just for illustration
- adds a bunch of tests

closes #30
impacts #35 and #33

### DIFF
--- a/api/env.js
+++ b/api/env.js
@@ -5,4 +5,9 @@ const defaults = {
 };
 
 dotenv.config();
-Object.assign(process.env, defaults);
+
+Object.keys(defaults).forEach((env) => {
+  if (!process.env[env]) {
+    process.env[env] = defaults[env];
+  }
+});
--- a/api/env.js
+++ b/api/env.js
@@ -7,9 +7,4 @@ const defaults = {
 };
 
 dotenv.config();
-
-Object.keys(defaults).forEach((env) => {
-  if (!process.env[env]) {
-    process.env[env] = defaults[env];
-  }
-});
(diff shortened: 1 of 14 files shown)


api/env.js
```diff
@@ -1,8 +1,15 @@
...
 const defaults = {
-  PORT: 8000
+  PORT: 8000,
+  SESSION_SECRET: crypto.randomBytes(32).toString('hex')
 };
 
 dotenv.config();
-Object.assign(process.env, defaults);
+
+Object.keys(defaults).forEach((env) => {
```
brendansudol: so we're adding new fields from `defaults` obj but not overriding anything already in `process.env`. is that equivalent to: `process.env = Object.assign({}, defaults, process.env)` ? (though perhaps it's bad practice to reassign `process.env`)

mgwalker: Oh, huh!  I didn't know `Object.assign` could take more than two arguments.  I'll update, I prefer your way.  😄

brendansudol: 👍

PR #50, opened 2017-12-18
Remove sudo in deploy script

mgwalker:
Because our CircleCI config uses the official Node images, we don't need `sudo`.  The official images run as root and don't switch users, so there's no `sudo` installed.

Also adds a Node version to the package.json `engines` property.  Otherwise cloud.gov gives us 4.8, which just plain ol' doesn't work. 😆

### DIFF
--- a/api/package.json
+++ b/api/package.json
@@ -37,6 +37,9 @@
     "dotenv": "^4.0.0",
     "express": "^4.16.2"
   },
+  "engines": {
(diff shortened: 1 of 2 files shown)


mgwalker:
The API still doesn't start, but it does deploy.  We'll have to figure out the problem with it not starting separately, but at least the CircleCI part seems good.  😄

PR #51, opened 2017-12-18
add starter redux bits

brendansudol:
This adds `redux` and related packages (`redux-logger`, `react-redux`), and initializes a file structure and example workflow for managing state (via a hello-world-esque `Counter` example).

Open JS Console to see info re: previous / next states as actions are dispatched.


web/src/app.js
```diff
@@ -1,9 +1,18 @@
...
 
-ReactDOM.render(
-  <HelloWorld name="friendly visitor" />,
+const middleware = [createLogger()];
+const store = createStore(reducer, applyMiddleware(...middleware));
+
+render(
+  <Provider store={store}>
+    <App />
+  </Provider>,
```
mgwalker: 🙌  Thank you for moving the app bootstrap component into its own thing!

.codeclimate.yml
```diff
@@ -1,4 +1,6 @@
 engines:
+  duplication:
+    enabled: false
```
mgwalker: I wish CodeClimate let you specific which files to apply this to.  It's a good idea, but tests are going to be ***FULL*** of duplication, intentionally.

web/src/containers/App.js
```diff
@@ -0,0 +1,27 @@
...
+
+import Counter from '../components/Counter';
+import { decrement, increment } from '../actions';
+
+const App = ({ dispatch, total }) => (
+  <div>
+    <Counter
+      total={total}
+      onIncrement={() => dispatch(increment())}
+      onDecrement={() => dispatch(decrement())}
```
mgwalker: I'm down with this, especially for something so simple.  How do you feel about the `mapDispatchToProps` approach for more complicated dispatch events?

.codeclimate.yml
brendansudol: yep, i totally agree -- i really wish we could keep this on and ignore it for test files...

web/src/containers/App.js
brendansudol: yep, def prefer `mapDispatchToProps` (as well as `bindActionCreators`) once we start going for realsies

brendansudol:
Thx @mgwalker, and agreed about chore issue around test coverage!

PR #54, opened 2017-12-18
add react-router, integrate with redux

brendansudol:
this adds `react-router` (4.x) and `react-router-redux@next`.

`react-router-redux` creates a binding between redux and react-router and keeps them in sync; it also allows for niceties like being able to navigate page/url changes by dispatching actions. this isn't strictly necessary (more info [here](https://github.com/ReactTraining/react-router/blob/master/packages/react-router/docs/guides/redux.md#deep-integration)), but it seems like it will be beneficial.

i opted to not test some of the new dummy components because they're really not going to be around for more than a week or so.

preview:
router-w-redux

**WIP**

Stay tuned...


web/.eslintrc.json
```diff
@@ -2,17 +2,26 @@
...
   "rules": {
     "comma-dangle": [1, "never"],
-    "react/jsx-filename-extension": 0
+    "react/jsx-filename-extension": 0,
+    "jsx-a11y/anchor-is-valid": [
+      2,
+      {
+        "components": ["Link"],
+        "specialLink": ["to"]
+      }
```
mgwalker: Nice.

web/package.json
```diff
@@ -35,10 +35,13 @@
   },
   "homepage": "https://github.com/18F/cms-hitech-apd#readme",
   "dependencies": {
+    "history": "^4.7.2",
     "prop-types": "^15.6.0",
     "react": "^16.2.0",
     "react-dom": "^16.2.0",
     "react-redux": "^5.0.6",
+    "react-router-dom": "^4.2.2",
+    "react-router-redux": "^5.0.0-alpha.9",
```
mgwalker: Should we explicitly declare a dependency on `react-router`?  I know we'll get it implicitly and maybe that's better - fewer direct dependency versions to manage, and let the downstream ones shake themselves out.  🤔   Now I've typed these words and I think the implicit dependency is better. 😛

brendansudol: hehe, cool, i like the implicity (that really should be a word!) in this case, and it's also how they do in their docs / guides

Issue #53, opened 2017-12-18
Add database connection to prototyping space


mgwalker:
Our cloud.gov prototyping space will need a database instance attached.

brendansudol:
<img width="1060" alt="cf" src="https://user-images.githubusercontent.com/1060893/34300438-0b214e6e-e6f6-11e7-8f7d-8bca07c88ff5.png">

PR #55, opened 2017-12-19
Add database, use it for authentication

### LINKED ISSUE #31: Add local database authentication

mgwalker:
Adds database layer ([knex](https://knexjs.org)) and updates authentication to use that.  Also updates tests accordingly.

### Local authentication

Local users are stored in the `users` table in the database.  Users are uniquely identified by an email address (for now, anyway).  There is also an incrementing `id` field.  Finally, their password is stored as a bcrypt hash.

Authentication is performed with a form-encoded `POST` request to the `/auth/login` API endpoint with `username` and `password` parameters.  The authentication steps are pretty simple:

1. query the `users` table for a row where `email` matches the username that was passed in
2. check if the hashed `password` from the database matches the password passed in
3. return an error if the database causes a problem, a `false` user object if no user matches, or a user object that looks like `{ username, id }`
4. respond with...
  a. with an HTTP 401 if the user does not exist or the password does not match
  b. with an HTTP 200 if the user is valid
  c. with an HTTP 500 if there is a database error, or probably anything unhandled

In addition to the HTTP status code, the API also sets a session cookie (called `session`, surprisingly).  The session cookie is encrypted with a secret key that is only accessible to the API and contains enough metadata to identify the user.  Basically after explicit authentication, subsequent requests are *implicitly* authenticated using the session cookie: if it decrypts successfully, it was set by the API and has not been tampered with.  Currently the encrypted cookie contains the user's database `id`.

### The database stuff

knex uses a `knexfile.js` to specify connection information to the database in different environments.  By default, it figures out its environment based on the `NODE_ENV` environment variable.  I have populated it for now with `development`, `test`, and `integration` environments.  I updated the Docker config to make the API container set `NODE_ENV` to `development.

knex also supports migrations.  I created an initial migration to create a very simple `users` table containing email address and password fields.

And finally, knex supports data seeds.  I created a seed file for populating a fake user in the development environment.  knex executes all the files it finds in the `seeds` directory, and I was a little concerned we might accidentally end up seeding the integration or production environments with test data.  It turns out that knex does _*not*_ recurse the `seeds` directory, though, so I created environment-specific seed subdirectories.  The only files directly in the `seeds` directory now are `development.js` and `test.js`.  These files check the `NODE_ENV` and only call their respective seeds if the environment matches.

I went ahead and created a test environment in the knexfile and created a seed for it, but I don't actually have a use for it yet.  I expect it's what we'll use for API testing, though.  Just have to figure out the best way to do it.  🙂

For development, I updated the API `Dockerfile` to run migration and seeding so we should be covered there.  For integration, I added a `Procfile` that runs migrations before starting the API, which should cover cloud.gov.

### Still needs...

- [ ] postgres setup on cloud.gov (#53)
- [ ] postgres linked to API on cloud.gov
- [ ] set `NODE_ENV` on cloud.gov to integration
  - this value and the one in the knexfile should match; right now I'm assuming we'll call it `integration`, but I don't care too much what the name is 😄 

### Tasks
Closes #31


api/auth/authenticate.js
```diff
@@ -1,10 +1,33 @@
...
+module.exports = (
+  db = defaultDB,
+  bcrypt = defaultBcrypt
+) => ((username, password, done) => db('users')
+  .where({
+    email: username
+  })
+  .select()
+  .then((users) => {
+    let user = { password: '' };
```
brendansudol: thoughts on early exiting if no `users` are found (before checking if passwords match)? i guess it means repeating the "failed login" callback; eh, either way is fine i suppose. also not sure how you feel about ternaries ;)

`const user = users.length ? users[0] : { password: '' }`

api/tests/db.js
```diff
@@ -0,0 +1,21 @@
...
+const db = require('../db');
+
+tap.test('database object module', (dbTest) => {
+  const knex = sinon.stub().returns({
+    thisIs: 'a database'
+  });
+
+  const nodeEnv = process.env.NODE_ENV;
+  process.env.NODE_ENV = 'env2';
+  const result = db(knex, { env1: 'environemtn 1', env2: 'environment 2' });
```
brendansudol: `environemtn` spelling  😜

api/auth/authenticate.js
```diff
@@ -8,15 +8,10 @@ module.exports = (
...
-    let user = { password: '' };
-    if (users.length) {
-      user = users[0];
-    }
-
+  .first()
+  .then((user) => {
     // If there is a matching user, return it
-    if (bcrypt.compareSync(password, user.password)) {
+    if (user && bcrypt.compareSync(password, user.password)) {
```
brendansudol: Nice!

PR #56, opened 2017-12-19
CSS option 1: PostCSS & ExtractTextPlugin

brendansudol:
**CSS Architecture Option 1**

`PostCSS` and `ExtractTextPlugin`

This is a more traditional approach where we have various stylesheets, both third-party and custom ones we make, and variables controlling styles in those sheets, and then leverage postcss and webpack to process and bundle that up into one `app.css` file.

This necessitates us to use the `className' property in our React components to add / compose styles and build up our UI library.


brendansudol:
Closing this in favor of option 2 :)

PR #59, opened 2017-12-21
add example new UI component

brendansudol:
Added a new UI component -- an `Alert` box that extends Rebass's `Box` component.

It takes a `kind` parameter that can be `['success', 'warning', 'error', ...]`; these correspond with the USWDS alert names, and the output is a similarly styled element.

Preview:
preview

### DIFF
--- a/web/.eslintrc.json
+++ b/web/.eslintrc.json
@@ -1,6 +1,7 @@
 {
   "extends": "airbnb",
   "rules": {
+    "arrow-parens": ["error", "as-needed"],
     "comma-dangle": [1, "never"],
     "react/jsx-filename-extension": 0,
     "jsx-a11y/anchor-is-valid": [
--- a/web/src/components/Hello.js
+++ b/web/src/components/Hello.js
@@ -1,7 +1,18 @@
-import React from 'react';
 import PropTypes from 'prop-types';
+import React from 'react';
+import { Measure, Subhead } from 'rebass';
 
-const Hello = props => <h1>Hello {props.name}!</h1>;
+import Alert from '../styles/Alert';
+
+const Hello = props => (
+  <div>
+    <h1>Hello {props.name}!</h1>
+    <Alert kind="error">
+      <Subhead>Hey-oh!</Subhead>
+      <Measure>This is a fake alert :)</Measure>
+    </Alert>
+  </div>
+);
 
 Hello.propTypes = {
   name: PropTypes.string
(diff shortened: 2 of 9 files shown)


web/.eslintrc.json
```diff
@@ -1,6 +1,7 @@
 {
   "extends": "airbnb",
   "rules": {
+    "arrow-parens": ["error", "as-needed"],
```
mgwalker: I don't have strong feelings about this rule, except that I've gotten used to the Airbnb default.  Now it looks strange to me to see an arrow function with a blocked body where the argument isn't wrapped in parens.  😛

web/src/components/Hello.js
```diff
@@ -1,7 +1,18 @@
...
+import Alert from '../styles/Alert';
 
-const Hello = props => <h1>Hello {props.name}!</h1>;
+const Hello = props => (
+  <div>
+    <h1>Hello {props.name}!</h1>
+    <Alert kind="error">
+      <Subhead>Hey-oh!</Subhead>
+      <Measure>This is a fake alert :)</Measure>
+    </Alert>
```
mgwalker: I know this is meant to be an example, but my first reaction is along the lines of, "Why do I have to know the component types for the title and message for this alert?"  I.e., would `<Alert kind="error" title="Hey-oh!">This is a fake alert :)</Alert>` make more sense?

Not a blocker for now, just something I wondered about.

web/src/styles/Alert.js
```diff
@@ -0,0 +1,26 @@
...
+    border-left: 0.5rem solid ${colors[1]};
+  `;
+};
+
+const Alert = styled(Box)`
+  ${props => cssRules(props.kind)};
+  padding: 1rem;
+`;
+
+export default Alert;
```
mgwalker: Oh, I see.  So `<Alert>` is used less like a standard component and more like a style class?  That's interesting!

web/.eslintrc.json
brendansudol: Yeah, this is one the casualties of war when using `prettier` -- the paren isn't necessary for single args so it's removed... there aren't too many conflicts between airbnb and prettier but there are a couple, but hopefully none are egregious, and keep me honest if you strongly object

web/src/styles/Alert.js
brendansudol: Yes! I think there are two levels of hierarchy and components we want to support -- these style class components, then one level up shared components that take in content / data props and compose slightly bigger things (using those style class components from below)

web/src/components/Hello.js
brendansudol: i think it all depends on how flexible we want our alerts to be, but yes, i think your Alert component props / API makes more sense for the use cases we will likely support

PR #60, opened 2017-12-22
bind cloud.gov db in manifest.yml

brendansudol:


manifest.yml
```diff
@@ -1,11 +1,13 @@
 ---
 applications:
   - name: hitech-apd-frontend
-    path: web/dist
     memory: 64M
+    path: web/dist
```
brendansudol: consistent order across apps

Issue #45, opened 2017-12-15
Update project README or other documentation


mgwalker:
- document authentication
  - [ ] passport configuration, etc. (#30)
  - [ ] API endpoints (#33)
  - [ ] authorization model (#32)
  - [ ] ~~token (#35)~~ authentication (see #47)

- document how the web app works
  - [ ] how routing works (#28)
  - [ ] how app state management works (#27)
  - [ ] how the app hits the API (#29)
  - [ ] how we're doing CSS (#34)

mgwalker:
For documenting API endpoints, what if we just wrote some Swagger?  I've got some code somewhere that conditionally enables CORS based on a Swagger spec, too, which might be pretty handy.

brendansudol:
I like idea of Swagger if it makes our lives easier

brendansudol:
I'm marking this as complete (with https://github.com/18F/cms-hitech-apd/pull/64 now merged); @mgwalker, if need be, let's open smaller issues for other documentation holes

PR #63, opened 2017-12-22
Add endpoints for local user management

### LINKED ISSUE #33: Add API endpoints to manage local users

mgwalker:
### Create a new user

An HTTP `POST` to `/user` allows an authenticated user to add a new user.  Expects a JSON body with an email address and password.  The password will be hashed with bcrypt before it is stored.  Sends an HTTP status 400 if the body is invalid or if the email address already exists in the database along with an error message token.  Otherwise it sends back an HTTP status 200 and an empty body.  Relies on the `loggedIn` middleware.

### Get a list of all users

An HTTP `GET` to `/users` allows an authenticated user to get a list of all users.  Sends the results back as JSON.  Relies on the `loggedIn` middleware.

### Get a specific user

An HTTP `GET` to `/users/:id` allows an authenticated user to get the user specified by the `:id` parameter (e.g., `/users/57` would get the user whose ID is 57).  Sends an HTTP status 400 if the ID is invalid (e.g., missing or not a number) along with an error message token.  Sends a status 404 and empty body if the user ID is not found.  Otherwise it sends back the user as JSON.  Relies on the `loggedIn` middleware.

## Middleware

To make it easier to lock routes to only authenticated users (and hopefully, down the road, to let us specific authorization requirements), I've added a `loggedIn` middleware.  Inserting this into the list of endpoint handlers will make sure the "main" handler is only called if the user is logged in.  E.g.:

```js
app.post('/user', loggedIn, (req, res) => { ...`
```

In the future, I'm envisioning extending this model so we also have an `authenticated` (or something else; I'm not a fan of that name, but can't think of a better one offhand) middleware:

```js
app.post('/user', authenticated('admin'), (req, res) => { ...`
```

The idea is simply that the various endpoint handlers shouldn't have to check user credentials themselves - that should be centralized.

## Error message tokens

This is me just making stuff up.  Instead of sending back full English strings describing the errors, I've made these endpoints send back token-ish strings like `add-user-email-exists`.  The idea is that it's readable for debugging purposes, but for displaying to users, it can be mapped to localized strings in the client.  This feels like an experiment to me.  Feedback welcomed and encouraged.  😆 

If we decide this is a good approach, we should probably capture all of the error "tokens" in a doc somewhere.  For #45, I was planning to create a `docs` folder to put development documentation in.  That might be a place for it.  However we do it, let's keep it lightweight.  😬 

## Random fixes

* The email field in the users table was not unique.  Oops!  There's a new migration to handle that.
* The API dockerfile tried to helpful migrate and seed the database with dev data on *build*, but that didn't work because the local development volume isn't mounted into the container until the container is *started*.  So knex was complaining that it couldn't find its config file and that was breaking the entire container build.  For now, migrations and seeds will just have to be run manually.  That's not too hard, though: once the containers are running, you can do:
   ```bash
   docker-compose exec api npm run migrate
   docker-compose exec api npm run seed
   ```
   For folks not using docker...  no change.  😉 

## Tasks

* Closes #33

### DIFF
--- a/api/routes/users/post.js
+++ b/api/routes/users/post.js
@@ -1,0 +1,37 @@
+const bcrypt = require('bcryptjs');
+const db = require('../../db')();
+const loggedIn = require('../../auth/middleware').loggedIn;
+
+const ifUserIsNew = (email, res) => db('users')
+  .where({ email }).first()
+  .then((user) => {
+    if (user) {
+      res.status(400).send({ error: 'add-user-email-exists' }).end();
+      return Promise.reject();
+    }
+    return Promise.resolve();
+  });
+
+const insert = (email, password) => {
+  const hashed = bcrypt.hashSync(password);
+  return db('users').insert({ email, password: hashed });
+};
+
+module.exports = (app) => {
+  app.post('/user', loggedIn, (req, res) => {
+    if (req.body.email && req.body.password) {
+      ifUserIsNew(req.body.email, res)
+        .then(() => insert(req.body.email, req.body.password))
+        .then(() => {
+          res.status(200).end();
+        })
+        .catch(() => {
+          // 👆 eventually we should catch errors here and log
+          // them somewhere.  For now... just eating them.
+          res.status(500).end();
+        });
+    } else {
+      res.status(400).send({ error: 'add-user-invalid' }).end();
+    }
+  });
+};
--- a/api/routes/users/post.js
+++ b/api/routes/users/post.js
@@ -2,15 +2,20 @@ const bcrypt = require('bcryptjs');
 const defaultDB = require('../../db')();
 const loggedIn = require('../../auth/middleware').loggedIn;
 
-const ifUserIsNew = (email, res, db) => db('users')
-  .where({ email }).first()
-  .then((user) => {
-    if (user) {
-      res.status(400).send({ error: 'add-user-email-exists' }).end();
-      return Promise.reject();
-    }
-    return Promise.resolve();
-  });
+const ifUserIsNew = (email, res, db) =>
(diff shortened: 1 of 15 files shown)


brendansudol:
so we're using `application/json` body for creating users and `application/x-www-form-urlencoded` for the passport user auth -- how do you feel about that, @mgwalker? it may be nice to standardize, but maybe there's good reasons for having both header types?

api/routes/users/post.js
```diff
@@ -0,0 +1,45 @@
...
+
+const insert = (email, password, db) => {
+  const hashed = bcrypt.hashSync(password);
+  return db('users').insert({ email, password: hashed });
+};
+
+module.exports = (app, db = defaultDB) => {
+  app.post('/user', loggedIn, (req, res) => {
+    if (req.body.email && req.body.password) {
+      ifUserIsNew(req.body.email, res, db)
```
brendansudol: malformed emails might be another good one to explicitly handle, but we can fold those in l8er

mgwalker:
Re: `application/x-www-form-urlencoded`, I agree it'd be nice to standardize on `application/json`.  The reason it's using the form-encoded type now is that I'm using the default Passport local authentication strategy, and that's what it expects.  I think.  Now that I write that out, I'm curious if a JSON POST would work, as long as the JSON body parser runs before the authentication code...  🤔   I'll experiment with that!

mgwalker:
It turns out that `application/json` will work for logging in because the JSON body parser is already run before authentication.  :tada:

brendansudol:
Sweet!

PR #74, opened 2018-01-02
Move API test files around

mgwalker:
This PR puts the API test files next to their corresponding source files, like with the frontend.


api/auth/authenticate.test.js
```diff
@@ -11,7 +11,7 @@ const bcrypt = {
   compareSync: sandbox.stub()
 };
 
-const auth = require('../../auth/authenticate.js')(db, bcrypt);
```
brendansudol: less `../..`s -- love it!

Issue #52, opened 2017-12-18
Setup tests to generate more accurate coverage


mgwalker:
On both sides, we need to make sure that all source files are imported at some point during testing so we get accurate coverage metrics of the entire project.  Without that, we only get coverage metric of sources that are imported during tests, which means we could easily forget to write tests for a new file and our automated tools wouldn't warn us.

- [x] client
  - can exclude `src/app.js` for now since it isn't doing anything that needs testing?
- [x] API
  - can exclude `main.js` for now since it isn't doing anything that needs testing, plus it will start a web server?

PR #77, opened 2018-01-03
import client source files for proper coverage stats

brendansudol:
This is part A of https://github.com/18F/cms-hitech-apd/issues/52 (client side)

Will do API side in a new PR tomorrow :)


web/.eslintrc.json
```diff
@@ -14,7 +14,8 @@
     ]
   },
   "env": {
-    "browser": true
+    "browser": true,
+    "jest": true
```
mgwalker: Do you think this `jest` env should be in the `overrides` section?  Something like...

```json
{
  "overrides": [
    {
      "files": ["**/*.test.js"],
      "env": {
        "jest": true
      }
    }
  ]
}
```

I think we shouldn't need the `globals` in there with the `env` set.  I like this `env` version way better.  😛

web/src/containers/Home.js
```diff
@@ -37,11 +29,11 @@ const mapDispatchToProps = dispatch => ({
...
       increment,
-      decrement,
-      goToHelloPage: () => push('/hello')
+      decrement
     },
     dispatch
   )
 });
 
+export { Home as Pure };
```
mgwalker: Nice!

PR #80, opened 2018-01-04
[WIP] Add API endpoint tests

mgwalker:
This PR adds API endpoint tests.

## The tests

The tests are in the `endpoint-tests` directory.  I chose to structure the directories to match the the HTTP routes (e.g., there's a route at `/auth/login`, so its tests are at `endpoint-tests/auth/login.js`.  For routes that support multiple HTTP verbs, I think something like `endpoint-tests/auth/login/get.js` and `endpoint-tests/auth/login/post.js` would be appropriate.

The tests are using `tap` like the API unit tests.  They use `request` to hit the running API and then check for valid status codes and bodies (if a body is returned).  In the case of the login endpoint tests, it also verifies that a cookie is set, but because the cookie is encrypted, it can't verify the cookie contents.

#### Docker

If you're developing in Docker, there's a handy-dandy shell script at `endpoint-tests/endpoint.sh` that will spin up the proper containers and run all the tests for you.  There's a compose script specifically for endpoint testing, to keep everything tidy and separate from development; that's the `docker-compose.endpoint-tests.yml`.  It handles setting all the right environment variables, putting together a testing database, creating the database structure and seeding it with test data.  Migration and seeding are done with ephemeral containers.  Once the tests have finished, it stops everything.  (Maybe it should `down`?  Not sure...)

#### Standalone

If you're not developing in Docker, you'll need to do a handful of steps:

1. make sure you have a `hitech_apd_test` database in your Postgres instance
2. set `NODE_ENV` environment variable to `test`
3. run `npm run migrate` and `npm run seed` to prepare the test database
4. run the API (either in the background, or in a separate terminal instance)
5. set `TEST_DB_HOST` environment variable to the address of your Postgres server
5. run `npm run test-endpoints`

## CI tests

I renamed the old `api tests` job to `api unit tests` and created a new job called `api endpoint tests`.  We'll need to update the list of required status checks in Github accordingly.

The endpoint tests job includes a 5 second pause before database migration and seeding to make sure the postgres container has started.  It then starts the API as a background job and pauses 5 more seconds to make sure the API is up.  Finally, it runs the endpoint tests against localhost.

## Environment variables

This PR adds three new environment variables:

|name|purpose|default|
|---|---|---|
|TEST_DB_HOST|The address to the database server to be used for running endpoint tests.|`db`|
|API_HOST|The address to the API to be used for running endpoint tests|`localhost`|
|API_PORT|The port that the test API is listening on|the `PORT` variable if defined, else `8000`|

## Still needs...

- [ ] tests for `/users` GET endpoint
- [ ] tests for `/user` POST endpoint
- [ ] update dev documentation


brendansudol:
not sure what's up with the hanging chad...

on the bright side, i get the tests passing locally (no docker) which is neat :)

brendansudol:
great stuff, thanks for laying down this great foundation, @mgwalker

PR #89, opened 2018-01-04
/user POST endpoint tests

mgwalker:
## Endpoint tests for `POST /user`

Adds a bunch of tests for this endpoint.  This should cover all of the endpoints we have right now.

## Database stuff in tests

Knex keeps database connections open indefinitely (probably all other database layers would do the same; doing otherwise would be silly), which is cool.  Tap (our testrunner) runs every test file in its own process, which is also cool (yay, test isolation!).  The problem here is that Tap expects the test processes to end on their own.  If they don't, it generates a timeout error and forcibly kills them, making it look like tests failed.

Tap has an event you can hook when it expects the test process to be finished - that is, all of the registered tests have been marked as done.  In that event handler, if you call `process.exit()`, that will move Tap along.  (It _should_ gracefully close the Knex connections, but also I guess who cares if it doesn't?  The process is gone, regardless.)

I added a helper method to the test utils to fetch a Knex object.  Calling this helper will also register a handler for Tap's `teardown` event that will exit the process.  This way, individual test files that use the database should ***NOT*** have to worry about killing their process.  It should just happen.

## Other stuff

* In `auth/login.js`, I modified it to use the `getFullPath` helper from test utils.  Everything else is the result of running prettier.
* In the `users/get` endpoint, its unit test, and its endpoint test, I addressed a bug where requesting an invalid user ID (i.e., non-numeric) would return a string rather than a JSON object

### DIFF
--- a/api/endpoint-tests/auth/login.js
+++ b/api/endpoint-tests/auth/login.js
@@ -1,11 +1,10 @@
 const tap = require('tap'); // eslint-disable-line import/no-extraneous-dependencies
 const request = require('request'); // eslint-disable-line import/no-extraneous-dependencies
+const { getFullPath } = require('../utils');
 
 tap.test('login endpoint | /auth/login', loginTest => {
-  const url = `http://${process.env.API_HOST || 'localhost'}:${process.env
-    .API_PORT ||
-    process.env.PORT ||
-    8000}/auth/login`;
+  const url = getFullPath('/auth/login');
+  console.log(url);
 
   loginTest.test('with no post body at all', invalidTest => {
     request.post(url, (err, response) => {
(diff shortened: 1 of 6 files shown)


api/endpoint-tests/auth/login.js
```diff
@@ -1,8 +1,10 @@
 const tap = require('tap'); // eslint-disable-line import/no-extraneous-dependencies
 const request = require('request'); // eslint-disable-line import/no-extraneous-dependencies
+const { getFullPath } = require('../utils');
 
 tap.test('login endpoint | /auth/login', loginTest => {
-  const url = `http://${process.env.API_HOST || 'localhost'}:${process.env.API_PORT || process.env.PORT || 8000}/auth/login`;
+  const url = getFullPath('/auth/login');
+  console.log(url);
```
brendansudol: is this log intentional?

brendansudol: removed it (really just so build steps would re-run, because 1 was cancelled)

PR #136, opened 2018-01-24
Switch API to using await/async instead of promise syntax

mgwalker:
Switches from using promises directly, with the sometimes-gnarly `.then()` code, to using the `await` and `async` syntactic sugar.  There are a couple of things worth pointing out specifically:

- all tests are updated to use `async` methods, which always return a promise; this way, we rely on tap's native promise support instead of having to remember to manually call `done()` on the tests
- the `request-promise` library does something weird when I use it with `await`: it prints the response object to the console and doesn't resolve like it's supposed to; as a temporary workaround, I added async wrapper methods to the endpoint test utils class

There's also a CodeClimate error that I think is valid, but I recommend that we ignore it for now because it should be cleaned up when we pick a DB relational layer.

### DIFF
--- a/api/auth/authenticate.test.js
+++ b/api/auth/authenticate.test.js
@@ -13,7 +13,7 @@ const bcrypt = {
 
 const auth = require('./authenticate.js')(db, bcrypt);
 
-tap.test('local authentication', authTest => {
+tap.test('local authentication', async authTest => {
   const doneCallback = sandbox.spy();
   authTest.beforeEach(done => {
     sandbox.reset();
@@ -43,8 +43,6 @@ tap.test('local authentication', authTest => {
       doneCallback.calledWith(sinon.match.truthy),
       'got an error message'
     );
-
-    errorTest.done();
   });
 
   authTest.test('with no valid users', async noUserTest => {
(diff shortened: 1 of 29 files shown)


api/auth/authenticate.test.js
```diff
@@ -22,123 +22,113 @@ tap.test('local authentication', authTest => {
...
-
-      errorTest.equal(doneCallback.callCount, 1, 'called done callback once');
-      errorTest.ok(
-        doneCallback.calledWith(sinon.match.truthy),
-        'got an error message'
-      );
-
-      errorTest.done();
-    });
+    await auth('user', 'password', doneCallback);
```
brendansudol: so clean!

api/endpoint-tests/users/post.js
```diff
@@ -23,111 +24,76 @@ tap.test('users endpoint | POST /user', postUsersTest => {
...
-              body,
-              { error: 'add-user-email-exists' },
-              'sends a token indicating the failure'
-            );
-            invalidTest.done();
-          }
+    authenticatedTests.test(
+      'with existing email address',
+      async invalidTest => {
+        const { response, body } = await request.post(url, {
```
brendansudol: 😍

PR #141, opened 2018-01-26
Use Bookshelf.js ORM

mgwalker:
~~I want to do a quick test run with Objection before we make a decision.  I shouldn't~~
~~have spent so much time getting tests and everything updated for Bookshelf until~~
~~we had looked into Objection more thoroughly...~~ 😬 

Having looked at [how Objection does relationship querying](http://vincit.github.io/objection.js/#fetch-queries), I don't really want to spend time coding it up.  It is almost as verbose as plain SQL would be.  It's more flexible than Bookshelf but not easier than just writing the queries directly via Knex, so I'm pretty comfortable saying we don't want to use Objection.

---

Instead of building and executing queries with knex directly, sets up data relationships with bookshelf and modifies code to use that.

Besides the small mountain of code changes (sorry!) to support this, the biggest impact right now is in the deserialization code, where a user's full list of activities is attached to the user object.  Instead of a lot of SQL joins of different kinds, it now relies on bookshelf's ability to resolves those relationships.  It's pretty fancy.

### Old code:
```js
const users = await db('users')
  .where({ id: userID })
  .select();

const role = users[0].auth_role;

const activityRows = await db('auth_role_activity_mapping')
  .fullOuterJoin(  
  'auth_roles',
    'auth_roles.id',
    'auth_role_activity_mapping.role_id'
  )
  .where({ 'auth_roles.name': role })
  .innerJoin(
    'auth_activities',
    'auth_activities.id',
    'auth_role_activity_mapping.activity_id'
  )
  .select();
activities = activityRows.map(row => row.name);
```

### New code:
```js
const user = await userModel.where({ id: userID }).fetch();
activities = await user.activities();
```

There's a lot of code here to support that fairly simple example, but as the data model grows and the relationships expand and get more complicated, this should save us a BUNCH of code complexity.

# To add a new model

The DB index file at `api/db/index.js` takes care of actually instantiating and cacheing all the models.  The model definitions are in the other module files in the `api/db/` directory.

First, add a new model definition file in the `api/db/` directory, or add a new model to an existing file.  (The [BookshelfJS documentation](http://bookshelfjs.org/#associations) isn't awesome, but it's the best resource for how to define a model.)  Your new module should export an object whose keys are the names of the models you're defining and whose values are the model definitions; for example:

```js
module.exports = {
  myNewAwesomeModel: {
    tableName: 'amazingBestTable'
  },
  myNewPrettyOkayModel: {
    tableName: 'prettyGoodTable'
  }
};
```

Then you need to update the `api/db/index.js` file to add your definitions to the list that it processes.  The `setup()` method takes an argument called `models` - that is the list of models to process.  Add your new definition to the `models` default value array and then you should be good to go:

```js
const thoseNewModels = require('./yourNewModelDefinitionFile');

const setup = (
  knex = defaultKnex,
  bookshelf = defaultBookshelf,
  config = defaultConfig,
  models = [user, authorization, thoseNewModels]
) => {
...
}
```

Finally, your models will be available off the DB index object.  The model object name is derived from the object keys returned by your definition module:

```js
const models = require('../db').models;
models.myNewAwesomeModel.where(...);
```

> **Note:** These instructions for adding a model assume you've also created the necessary knex migrations and run them.

---

Major kudos to @brendansudol for proposing this and investigating alternatives.


api/db/authorization.js
```diff
@@ -0,0 +1,40 @@
...
+      activitiesModel,
+      'auth_role_activity_mapping',
+      'role_id',
+      'activity_id'
+    );
+  }
+};
+
+module.exports = (bookshelf = defaultBookshelf) => {
+  if (!activitiesModel) {
```
brendansudol: i don't understand why these conditional checks are needed (`if (!a) a = bookshelf.extend...`)

api/db/index.js
```diff
@@ -0,0 +1,27 @@
...
+  knex = defaultKnex,
+  bookshelf = defaultBookshelf,
+  config = defaultConfig,
+  models = { user, authorization }
+) => {
+  const db = knex(config[process.env.NODE_ENV]);
+  const orm = bookshelf(db);
+
+  Object.keys(models).forEach(modelName => {
+    exportedModels[modelName] = models[modelName](orm);
```
brendansudol: I'm confused by this bit -- the `exportedModels` object only gets populated when `setup` is called (which prompts the `forEach` loop) but until then (and in the module.exports), it's just an empty `{}`, is that right?

api/db/user.js
```diff
@@ -0,0 +1,31 @@
...
+) => {
+  if (!model) {
+    model = bookshelf.Model.extend({
+      tableName: 'users',
+
+      role() {
+        return this.hasOne(AuthModels().roles, 'name', 'auth_role');
+      },
+
+      async activities() {
```
brendansudol: nice!

api/db/authorization.js
mgwalker: Because the default export a function that returns the two models, so this prevents those models from being recreated every time it's called.  Basically it's to support cacheing.  I'm not convinced this is the best way to accomplish that.  Maybe they should export a `setup` method that creates the models and the default export is just an object or something?  🤔

api/db/index.js
mgwalker: Correct.  It has to modify the existing object because the default export is cached by Node.js the first time it's created.  If you reassign `exportedModels` to a new object, nobody will see it because Node.js has cached off the original value.  If you modify the `exportedModels` object, though, everyone gets it.  Yay, by-reference.  😄 

So basically this is the alternate version of cacheing I described in the other comment.  It only just now occurs to me that I'm doing it two different ways...  Should almost certainly land on just one or the other.

brendansudol: Gotcha. Ok, I understand now; makes sense. Do you have a cacheing approach preference?

api/db/authorization.js
brendansudol: Got it, thanks -- i'm not opposed to this pattern fwiw

api/db/index.js
mgwalker: Not really, but now I want to think about it for a little bit before we merge it.  The good news is that we can change it later without much problem.

mgwalker: Ah, now I remember.  The idea is that to access models, you'd do `require('../db').models.user` (for example) rather than referencing the model directly.  The models would self-cache just in case someone *did* reference them directly, but the intent is that by going through the `db` module, you could get to them as objects and not have to remember to call them as functions (which I know from experience, I would forget *all the time*).

They're still functions under the hood, though, so that their dependencies can be injected for testing.

I think the whole thing can be cleaned up some more, though.  And it's also not entirely consistent right now, to boot - to get the roles model, for example, you'd need to do `require('../db').models.authorization.roles` and that is _*SUPER*_ non-obvious.

brendansudol: Ahh, I see - yeah I like this approach to access models. 👍

mgwalker:
@brendansudol I added a section to the PR description that shows how to add a new model, which hopefully highlights the changes I just made.  I feel a lot better about the state of the code now.

Instead of each set of model modules creating themselves with Bookshelf, the modules just return model definitions, and the `db/index.js` module creates the models with Bookshelf.  This way, the cacheing happens all on its own, and it simplifies the code because the Bookshelf code ends up being primarily in one place.  I'm also using the Bookshelf `registry` plugin, which allows us to register models by name so models that rely on each other can reference each other with a string instead of an object - that should eliminate the risk of circular dependencies (that risk is why the `role` and `activity` model definitions were included in the same `db/authorization.js` file originally; I left them in the same file because it seems like a reasonable grouping).

brendansudol:
updates are 💯 @mgwalker

PR #142, opened 2018-01-26
Our first form

brendansudol:
This PR adds [redux-form](https://github.com/erikras/redux-form), a library that helps manage a lot of the intricacies of working with forms. A few highlights:

* stores form values in our redux store (and abstracts away all the action creators / reducers to add and update those values
* real-time client side validation (upon submission and/or per keystroke)
* customizable input format / styles

There are two other libraries I considered: [react-redux-form](https://github.com/davidkpiano/react-redux-form), which is similar to `redux-form` and quite flexible (i.e., you create your own reducers for form values) but has lower adoption; and [formik](https://github.com/jaredpalmer/formik) which by design does not use redux (says it's unnecessary) and simply focuses on keeping track of your form's state and then exposes it plus a few reusable methods and event handlers to your form via props. I think any of these three libraries would work for us fine, but narrowed in on `redux-form` to start since it seems like the most established and has a ton of examples to draw from. Happy to sub it out if we decide later that it's too cumbersome / not a great fit with our use-case.

I also added a starter form component to kick things off.

Next up (in future PRs):
* better looking inputs
* inputs that actually correspond with the information we want to gather (from content doc)
* more pages / forms

### DIFF
--- a/web/src/components/Form1.js
+++ b/web/src/components/Form1.js
@@ -17,7 +17,7 @@ const Form1 = ({ handleSubmit, pristine, reset, submitting }) => (
       </div>
     </div>
     <div>
-      <label htmlFor="firstName">Last Name</label>
+      <label htmlFor="lastName">Last Name</label>
       <div>
         <Field
           id="lastName"
--- a/web/src/components/Form1.js
+++ b/web/src/components/Form1.js
@@ -1,0 +1,66 @@
+import PropTypes from 'prop-types';
+import React from 'react';
+import { Field, reduxForm } from 'redux-form';
+
+const Form1 = ({ handleSubmit, pristine, reset, submitting }) => (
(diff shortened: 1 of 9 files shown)


web/src/components/Form1.js
```diff
@@ -0,0 +1,66 @@
...
+          id="firstName"
+          name="firstName"
+          component="input"
+          type="text"
+          placeholder="First Name"
+        />
+      </div>
+    </div>
+    <div>
+      <label htmlFor="firstName">Last Name</label>
```
mgwalker: I think this should be `htmlFor="lastName"`?  :)

brendansudol: yes, thx! (and i feel like that duplicate label keys should be a linting check 🙃)

PR #143, opened 2018-01-29
Start state contacts page

brendansudol:
This PR adds a new page (`/state-contacts`) and a new form (`Form2` [should rename soon]).

Also included is a new React component `InputHolder` which allows us to leverage `rebass` styled components while also using the `redux-form` methods and props.

(see https://redux-form.com/7.1.2/docs/api/field.md/#2-a-stateless-function)

This new form also showcases a pattern for how to work with "Field Arrays", which is needed when we don't know how many things a user will provide (i.e., contacts)


web/src/components/Input.js
```diff
@@ -0,0 +1,25 @@
+import PropTypes from 'prop-types';
+import React from 'react';
+import { Box, Input, Label } from 'rebass';
+
+const InputHolder = ({ input, label, type, meta: { touched, error } }) => (
+  <Box mb={3}>
+    <Label>{label}</Label>
+    <Input {...input} type={type} />
```
mgwalker: Maybe getting a little ahead of us here, but I'm wondering if we should generate a unique ID here for the `Input` and then assign the `Label`'s `htmlFor` prop to that value?  It'd be pretty nice to not have to worry about the IDs.

Just threw together an example of how it might work to be sure the `id` and `htmlFor` props would pass through Rebass, and they do.  😄

brendansudol: yes, those definitely should be in here -- will add, thx

web/.eslintrc.json
```diff
@@ -21,6 +21,10 @@
       }
     ],
 
+    "react/forbid-prop-types": ["error", { "forbid": ["any"] }],
```
brendansudol: just reorganizing react rules from jsx rules from all other rules

web/src/components/Input.js
```diff
@@ -0,0 +1,29 @@
...
+import { Box, Input, Label } from 'rebass';
+
+const InputHolder = ({
+  input: { name, ...rest },
+  meta: { touched, error },
+  label,
+  type
+}) => (
+  <Box mb={3}>
+    <Label htmlFor={name}>{label}</Label>
```
brendansudol: baked in the `htmlFor` and `id` of input, @mgwalker

PR #156, opened 2018-02-01
CheckboxGroup component, Activities start

brendansudol:
The main contribution of this PR is a new component, `CheckboxGroup`, which allows us to have a multi-select checkbox form elements that still plays nice with `redux-form`'s way of doing things. Much of this is inspired by the discussion and code bits in [this thread](https://github.com/erikras/redux-form/issues/1037#issuecomment-243003954). As you'll see, before dispatching the action to update the store, we mutate the local state so that values for the checkbox group are in the form of an array. Also you can see how to pre-fill data into form inputs using the `initialValues` prop. For example,

```
const formConfig = {
  form: 'formName',
  initialValues: { activities: ['bar'] }
};

export default reduxForm(formConfig)(FormComponent);
```


web/src/components/CheckboxGroup.js
```diff
@@ -0,0 +1,52 @@
...
+
+      return onChange(vals);
+    };
+
+    return (
+      <Label key={`checkbox-${value}`}>
+        <Checkbox
+          name={`${name}[${idx}]`}
+          value={value}
+          checked={inputVal.includes(value)}
```
mgwalker: Does `Array.includes` get polyfilled or transpiled by Babel?  I know it's not supported by IE, which we probably have to support, so wanted to flag it.

web/src/components/Form4.js
```diff
@@ -0,0 +1,49 @@
...
+
+Form4.propTypes = {
+  handleSubmit: PropTypes.func.isRequired,
+  pristine: PropTypes.bool.isRequired,
+  reset: PropTypes.func.isRequired,
+  submitting: PropTypes.bool.isRequired
+};
+
+const formConfig = {
+  form: 'form3',
```
mgwalker: You know I'm going to call out the `Form4` name, but incidentally, I think this config name is supposed to be `form4`.  😛

web/src/components/CheckboxGroup.js
brendansudol: good call, need to add `babel-polyfill` (https://babeljs.io/docs/usage/polyfill/)

web/src/components/Form4.js
brendansudol: so not only is it a bad name, it's the wrong, bad name -- 🤦‍♂️

brendansudol:
Added my updates @mgwalker -- updated a couple other small things (which I recognize is not good practice 😬) so that it's more demo-able for upcoming meeting

PR #158, opened 2018-02-01
State start page updates

brendansudol:
now form fields and text is more in line with google doc

Preview:
preview


brendansudol:
https://github.com/18F/cms-hitech-apd/pull/153 should be merged first

web/src/components/FormStateStart.js
```diff
@@ -1,37 +1,44 @@
...
+      name="phoneNumber"
+      type="tel"
+      pattern="[0-9]{3}-[0-9]{3}-[0-9]{4}"
+      component={Input}
+      label="Phone number"
+    />
 
-    <div>
-      <label htmlFor="favoriteColor">Favorite Color</label>
+    <SectionHeader>State:</SectionHeader>
```
mgwalker: Let's leave it for now, but I'll be interested to see what design has to say about this header after a little bit of testing.  Not every entity submitted APDs is a state.  😆

web/src/util/index.js
```diff
@@ -0,0 +1,55 @@
...
+  { id: 'tx', name: 'Texas' },
+  { id: 'ut', name: 'Utah' },
+  { id: 'vt', name: 'Vermont' },
+  { id: 'va', name: 'Virginia' },
+  { id: 'wa', name: 'Washington' },
+  { id: 'dc', name: 'Washington, DC' },
+  { id: 'wv', name: 'West Virginia' },
+  { id: 'wi', name: 'Wisconsin' },
+  { id: 'wy', name: 'Wyoming' }
+];
```
mgwalker: While we should let the header go for now, this list should include the District of Columbia, American Samoa, Guam, Northern Mariana Islands, Puerto Rico, and US Virgin Islands.  I think that's all, but we should ask @jeromeleecms or @NAretakis to make sure.

PR #160, opened 2018-02-02
Adds a personnel form

### LINKED ISSUE #150: Initial APD prototype - personnel

mgwalker:
Adds a form for entering state and contracting personnel and computes what those personnel will cost.  This closes #150.

* Adds the `/state-personnel` route
* Updates the APD overview page to go to the personnel page next
* Updates the `Input` component to add an option to turn labels off
  - labels are pushed offscreen so screen readers will still read them
  - we should discuss this approach more; are there cases where the labels should also be  hidden from screen readers?
* Adds a `<Dollars>` component for formatting numbers into dollars for display

This is pretty ugly, IMO, but it's a prototype...  Also a good excuse to play with redux-forms.  😄 

- [x] ~~Tests have been updated (and~~ all tests are passing ~~)~~
  - foregoing ***new*** tests on the prototype UI components

- Design has approved the experience
  - N/A - design will definitely not approve this experience, but that's intentional!
- Product has approved the experience
  - N/A - same

### DIFF
--- a/web/src/components/FormPersonnel.js
+++ b/web/src/components/FormPersonnel.js
@@ -3,7 +3,7 @@ import React from 'react';
 import { Button } from 'rebass';
 import { Field, FieldArray, reduxForm } from 'redux-form';
 
-import Input from './Input';
+import { Input } from './Inputs';
 
 const PersonnelList = ({ fields, meta: { error, submitFailed } }) => (
   <div>
@@ -56,6 +56,7 @@ const PersonnelList = ({ fields, meta: { error, submitFailed } }) => (
                 name={`${person}.nextCompensation`}
                 component={Input}
                 parse={value => Number(value)}
+                format={value => `${value}`}
                 type="number"
               />
             </td>
@@ -66,6 +67,7 @@ const PersonnelList = ({ fields, meta: { error, submitFailed } }) => (
                 name={`${person}.nextTime`}
                 component={Input}
                 parse={value => Number(value)}
+                format={value => `${value}`}
                 type="number"
               />
             </td>
@@ -77,6 +79,7 @@ const PersonnelList = ({ fields, meta: { error, submitFailed } }) => (
                 name={`${person}.nextNextTime`}
                 component={Input}
                 parse={value => Number(value)}
+                format={value => `${value}`}
                 type="number"
               />
             </td>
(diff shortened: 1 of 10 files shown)


web/src/components/FormPersonnel.js
```diff
@@ -0,0 +1,152 @@
...
+                component={Input}
+              />
+            </td>
+            <td>
+              <Field
+                hideLabel
+                label="2019 compensation"
+                name={`${person}.nextCompensation`}
+                component={Input}
+                parse={value => Number(value)}
```
brendansudol: can also do `+value` to convert to value -- more terse than `Number(value)` though also less clear (i have no pref except for consistency)

web/src/components/Inputs.js
```diff
@@ -2,17 +2,25 @@ import PropTypes from 'prop-types';
 import React from 'react';
 import { Box, Input, Label, Textarea } from 'rebass';
 
+const offscreenStyle = {
```
brendansudol: hiding labels (and keeping them still visible for screen readers) is perfectly legit, and your CSS to do this looks good. 18F's accessibility guide offers a slightly modified CSS snippet to accomplish the same thing (https://accessibility.18f.gov/hidden-content/) but they admit "There are multiple ways to accomplish this"

web/src/containers/StatePersonnel.js
```diff
@@ -0,0 +1,111 @@
...
+}
+
+StatePersonnel.propTypes = {
+  goTo: PropTypes.func.isRequired,
+  statePersonnel: PropTypes.array.isRequired,
+  contractingPersonnel: PropTypes.array.isRequired
+};
+
+const mapDispatchToProps = dispatch =>
+  bindActionCreators({ goTo: path => push(path) }, dispatch);
```
brendansudol: we should figure out if there's a better way to abstract this `goTo` method across our connected components so we don't have to repeat it everywhere (I know I started this 😳) but we can still navigate to diff pages with redux actions

brendansudol:
none of my comments are merge blocking, so after reading through, feel free to hit that big green button :)

mgwalker:
* Switched from `Number(value)` to `+value`
* added a global `sr-only` style (using the 18F guide's version) and switched to using that for visually-hidden labels

web/src/containers/StatePersonnel.js
mgwalker: Agreed.  It'd be pretty cool if we mapped out the flow in a global object somewhere and then had a `next` event that advanced to the next page, maybe?  Or something else, I dunno.  lol.  We can brainstorm on a good approach!

PR #165, opened 2018-02-05
add 404 component

brendansudol:
Displays a 404 / No Match component when there is no matching front-end route (as documented at https://reacttraining.com/react-router/web/example/no-match).

Preview:
preview

### DIFF
--- a/web/src/components/App.js
+++ b/web/src/components/App.js
@@ -1,9 +1,10 @@
 import React from 'react';
-import { Link, Route } from 'react-router-dom';
+import { Link, Route, Switch } from 'react-router-dom';
 import { Container, NavLink, Provider, Toolbar } from 'rebass';
 
 import Demo from './Demo';
 import DevProvider from './DevProvider';
+import NoMatch from './NoMatch';
 import ActivitiesStart from '../containers/ActivitiesStart';
 import ApdOverview from '../containers/ApdOverview';
 import StateContacts from '../containers/StateContacts';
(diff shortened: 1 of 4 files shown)


web/src/components/App.js
```diff
@@ -24,12 +25,15 @@ const App = () => (
     </Toolbar>
 
     <Container is="main">
-      <Route exact path="/" component={Demo} />
-      <Route path="/activities-start" component={ActivitiesStart} />
-      <Route path="/apd-overview" component={ApdOverview} />
-      <Route path="/state-start" component={StateStart} />
-      <Route path="/state-contacts" component={StateContacts} />
-      <Route path="/state-personnel" component={StatePersonnel} />
+      <Switch>
```
mgwalker: This is a cool component.  Having just read a little about it, I kind of feel like maybe `<Switch>` should always be at the root of page-level routing.  😛

web/src/components/NoMatch.js
```diff
@@ -0,0 +1,15 @@
...
+
+const NoMatch = () => (
+  <Box py={5}>
+    <Heading mb={1} center>
+      Whoops, we can’t find that page.
+    </Heading>
+    <Text center>
+      You might want to double-check your link and try again. (404)
+    </Text>
+  </Box>
```
mgwalker: This basically answers my primary question, which was whether we could set the content and style for the 404.  👍

PR #170, opened 2018-02-06
Add milestone schedule form page

mgwalker:
Adds the milestone schedule page to the form UX prototype.

### DIFF
--- a/web/src/components/FormActivitySchedule.js
+++ b/web/src/components/FormActivitySchedule.js
@@ -1,0 +1,176 @@
+import PropTypes from 'prop-types';
+import React from 'react';
+import { Box, Button, Flex } from 'rebass';
+import { Field, FieldArray, reduxForm } from 'redux-form';
+
+import { Input } from './Inputs';
+import SectionHeader from './SectionHeader';
+
+const AddButton = ({ fields, meta: { error, submitFailed } }) => (
+  <Box>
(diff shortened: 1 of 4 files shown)


web/src/components/FormActivitySchedule.js
```diff
@@ -0,0 +1,176 @@
...
+          <th>Actual end date</th>
+          <th />
+        </tr>
+      </thead>
+      <tbody>
+        <FieldArray name="milestones" component={Milestones} />
+      </tbody>
+    </table>
+
+    <FieldArray name="milestones" component={AddButton} />
```
brendansudol: Not sure I understand this bit -- why is `AddButton` being used as a FieldArray? Is that just so it gets the `fields` prop for `milestones`?

mgwalker: Yeah, that's exactly why.  It was the most obvious way of getting access to the `fields` prop - and faster than digging through redux-form docs.  😛

brendansudol: Ok, i don't love that, but we can figure out a cleaner way and refactor soon

mgwalker: I also don't love it.  I think this falls under the conversation we were having about how to take the `redux-form` state and merge it into our "real" state.

PR #172, opened 2018-02-07
prototype refinement

brendansudol:
small layout and style refinements

Preview:
screen shot 2018-02-07 at 6 00 44 pm

screen shot 2018-02-07 at 5 58 50 pm

### DIFF
--- a/web/src/components/FormActivityGoals.js
+++ b/web/src/components/FormActivityGoals.js
@@ -1,30 +1,33 @@
 import PropTypes from 'prop-types';
 import React from 'react';
-import { Box, Button, Flex } from 'rebass';
+import { Absolute, Box, Button, ButtonOutline, Relative } from 'rebass';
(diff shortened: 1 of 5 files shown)


brendansudol:
I still gotta do that non-full-width input thing (but i'm still noodling on the best way to handle that)

web/src/components/FormActivityGoals.js
```diff
@@ -35,32 +38,31 @@ Objectives.propTypes = {
...
+          name={`${goal}.description`}
+          component={Textarea}
+          label="Description"
+        />
+        <FieldArray name={`${goal}.objectives`} component={Objectives} />
       </Box>
     ))}
     <Box>
-      <Button bg="black" onClick={() => fields.push({})}>
+      <Button bg="black" onClick={() => fields.push({ ...entryShell })}>
```
mgwalker: I like this `entryShell` bit.  👍

Issue #151, opened 2018-01-30
Initial APD prototype - expenses


brendansudol:

nkkl:
@mgwalker is task ready to close?

mgwalker:
I don't think so.  We have code to support it, but I took the opportunity to fiddle with some different routing logic, file structure, etc.  There will be an ongoing conversation in #173 about some of that.

So, we can think of it two ways: yes, the prototype implementation is done and now we're discussing it from a code perspective; or no, the prototype implementation is not done because users can't test it yet.  I lean towards the latter, personally.

nkkl:
Ok, let's keep it open, then.

PR #173, opened 2018-02-08
[WIP/Discussion] Expenses form prototype

mgwalker:
* The expenses section is multiple screens, so instead of having a bunch of new form files in the `components` directory, I created a `components/Form/Expenses` directory to gather all the new form files (there ended up just being two...  but still, pattern)
* Likewise I created a `containers/Expenses` directory for gathering the containers
* The `Expenses` index sets up internal routes to the various pieces of the expenses section

I'm not really thrilled with having the nav buttons spread through the different pieces because it's a fair bit of duplication, but those pieces "know" where they should go next.  That said (and ignoring ugly styling), I'm generally pretty happy with this.

#151


mgwalker:
Regarding the CodeClimate issue, I feel like it's okay.  It's more lines than it needs to be because of `prettier`, which is an A+ use of extra lines of code IMO.  😛

web/src/components/Form/Expenses/index.js
```diff
@@ -0,0 +1,9 @@
+import CommonExpenses from './CommonExpenses';
+import ExpenseCategory from './ExpenseCategory';
+
+export default {
```
brendansudol: why default and named exports for same components?

web/src/containers/Expenses/index.js
```diff
@@ -0,0 +1,59 @@
...
+        <Switch>
+          <Route
+            path={`${root}/list`}
+            component={() => (
+              <CategoryList goTo={goTo} next="/apd-overview" prev={root} />
+            )}
+          />
+          <Route
+            path={`${root}/edit`}
+            component={() => <EditExpense goTo={goTo} next={`${root}/list`} />}
```
brendansudol: so to pass in props to a component for `Route` you have to wrap in the function? not the prettiest but I guess it works!

web/src/components/Form/Expenses/index.js
mgwalker: I don't actually know.  I _think_ it's an approach I adopted once when I wanted to be able to import the parts of the module in different ways (e.g., `import { x } from './module'` and `import mod from './mod'; mod.x();`).  It's nice because you can have everything namespaced into that one imported object.  I don't feel strongly about it at all, though - I think it's a leftover habit from when I was still trying to figure out how imports/exports worked.

web/src/containers/Expenses/index.js
mgwalker: Yeah, we can definitely find a better solution.  This was just get-it-out-the-door code.

web/src/containers/Expenses/Edit.js
```diff
@@ -0,0 +1,22 @@
+import PropTypes from 'prop-types';
```
brendansudol: I feel like the only components that should be in `containers` directory are those that are connected via redux. (that's my understanding of the organizational design behind container vs component directories.)

with non-connected components now in `containers` dir, it seems less clear where to put things

web/src/components/Form/Expenses/ExpenseCategory.js
```diff
@@ -0,0 +1,101 @@
+// import PropTypes from 'prop-types';
+import React from 'react';
+import { Box, Button } from 'rebass';
+import { Field, FieldArray, reduxForm } from 'redux-form';
+
+import { Input } from '../../Inputs';
+import SectionHeader from '../../SectionHeader';
+
+const ExpenseCategory = () => (
+  <Box py={4}>
```
brendansudol: for accessibility, should these be wrapped in a `form` element?

brendansudol:
there's pros and cons to this new pattern. one downside is that you can't easily grok all of the routes in the app in one place (which you could before by just looking at `App.js`), now things are more nested and you have to hunt around. maybe we were going to need that anyway... i also feel like now we're muddying the line between what's a component and what's a container. i'm also finding some of the REST-y naming confusing -- for example, there's `Common` and `Edit` components within `Expenses`, which leads me to believe that only the `Edit` one takes in inputs from the user, but `Common` also does that.

web/src/containers/Expenses/Edit.js
mgwalker: Ah, I didn't know that's where the naming came from.  I was operating under the assumption that a container was a thing that wasn't generally reused; that is, essentially a standalone thing that only exists in one context.

brendansudol: here's the article that influenced my thinking on this - https://medium.com/@dan_abramov/smart-and-dumb-components-7ca2f9a7c7d0 (and redux example apps - https://github.com/reactjs/redux/tree/master/examples)

web/src/components/Form/Expenses/index.js
brendansudol: i'm all for it if it makes things easier, but I'm not sure it's needed yet. perhaps let's keep them as just named exports until housing them under one object helps clean things up.

```
export { default as CommonExpenses } from './CommonExpenses'
export { default as ExpenseCategory } from './ExpenseCategory'
```

web/src/containers/Expenses/Edit.js
mgwalker: So a short summary of the distinction between the two might be:

Components are:
- not connected directly to app state
- do not mutate the state
- receive their data and event handlers via props

Containers are:
- basically every other piece of UI

Is that roughly right?  I can be on board with that.

web/src/components/Form/Expenses/ExpenseCategory.js
mgwalker: No idea.  Probably?  😬

mgwalker:
Setting aside the names for a second (yeah, I agree they could be better...  naming things is hard!), I think the biggest thing this PR introduces is the routing change.  (The other stuff is mostly just wrong on my part.)

I understand what you're saying about having all the routes in one place, and there's merit to that.  My concern is trying to map out nested paths at a high level and essentially imposing that in a top-down fashion instead of letting each container handle its own internal routing.  You end up with a hierarchical route flow that is controlled by hierarchical components.  It might be harder to understand the total route tree, but it'd be pretty quick to find which component goes to each endpoint (though, obviously, it's super quick right now too 😄).

I'm also making the assumption that we want nested routes at all, which may not be true.  I'm imagining a scenario where someone in the state is working on their APD and they want someone else to login and help them with a piece, it'd be cool if they could send them a direct link to whatever section they're looking at.  It may turn out that nobody wants that in practice.

In short, I'm okay with not adopting this routing approach right now, but I wanted to try it out and see what it might look like.  I think we accomplished that.  I'm okay with closing this PR as a reference in case we end up needing a different routing approach and we want this to look at, and I can open another PR that uses our current routing approach (with containers/components/names made a little nicer, hopefully).

web/src/containers/Expenses/Edit.js
brendansudol: yes, i agree!

web/src/components/Form/Expenses/ExpenseCategory.js
brendansudol: i'm not sure it's a accessibility requirement, but it's good html semantics

brendansudol:
thanks for the write-up. i'm really glad we tried this routing approach to see what it looks like and how it feels, and it's very possible that in a week, we're going to agree that it's the way to go.

in the meantime, i think we should also do some research into some bigger react-router projects and see how they structure them. i suspect that after a certain point in complexity, nested routing and a file organization similar to yours is the way to go, but let's see what patterns are out there that we can steal 😏

mgwalker:
Going to replace this PR with a hunk of code more in line with our pre-existing practices.  Closing in favor of #209.

PR #174, opened 2018-02-08
prototype tweaks

brendansudol:
Fixes and tweaks based on walkthrough with @nkkl last night.

* "continue" button takes you back to activity list on approaches page
* removed example PII
* fixed hover/active color bug on `ButtonOutline`

### DIFF
--- a/web/.eslintrc.json
+++ b/web/.eslintrc.json
@@ -27,6 +27,7 @@
 
     // unnecessary due to prettier
     "function-paren-newline": "off",
+    "indent": "off",
     "object-curly-newline": "off"
   },
   "env": {
--- a/web/src/components/FormStateContacts.js
+++ b/web/src/components/FormStateContacts.js
@@ -119,9 +119,9 @@ const formConfig = {
       zip: '05671-1010'
     },
     medicaidDirector: {
-      name: 'Cory Gustafson',
-      email: 'cory.gustafson@vermont.gov',
-      phone: '802-879‐5901'
+      name: 'First Last',
+      email: 'first.last@vermont.gov',
+      phone: '555-123-4567'
     }
   },
   destroyOnUnmount: false
--- a/web/src/components/FormStateContacts.js
+++ b/web/src/components/FormStateContacts.js
@@ -120,7 +120,7 @@ const formConfig = {
     },
     medicaidDirector: {
       name: 'First Last',
-      email: 'first.last@vermont.gov',
+      email: 'first.last@state.gov',
       phone: '555-123-4567'
     }
   },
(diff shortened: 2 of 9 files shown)


web/.eslintrc.json
```diff
@@ -27,6 +27,7 @@
 
     // unnecessary due to prettier
     "function-paren-newline": "off",
+    "indent": "off",
```
mgwalker: Why this one?  😛

web/src/components/FormStateContacts.js
```diff
@@ -119,9 +119,9 @@ const formConfig = {
       zip: '05671-1010'
     },
     medicaidDirector: {
-      name: 'Cory Gustafson',
-      email: 'cory.gustafson@vermont.gov',
-      phone: '802-879‐5901'
+      name: 'First Last',
+      email: 'first.last@vermont.gov',
```
mgwalker: Should this be `@state.gov` too, like in `FormStateStart.js`?  (Amusingly, state.gov is still real.)

web/src/styles/ButtonOutline.js
```diff
@@ -0,0 +1,27 @@
+import styled from 'styled-components';
+import { ButtonOutline, util } from 'rebass';
```
mgwalker: Where did you find docs about this `util` thing?  The rebass docs seem to be missing a lot of stuff that isn't the components, and even with the components, it's basically just "here they are."  😕

web/.eslintrc.json
brendansudol: because prettier handles it and sometimes will conflict with eslint's ideas of where indents are "needed"

web/src/components/FormStateContacts.js
brendansudol: yes, will change

web/src/styles/ButtonOutline.js
brendansudol: found them just by reading the source code... yes, the docs are certainly lacking and i'm also worried that this library is getting a little stale too... i think in general `rebass` has been helpful during this prototyping phase, and i'm not sure i __love__ it / will even __like__ it long-term. we should talk more about this next sprint

mgwalker: Ah good, glad to hear that (sorta...  you know what I mean).  I'm not exactly loving it either.  It seems _okay_, but I feel pretty confident that there are better ways forward.

brendansudol: agreed 👍

brendansudol: i'm happy we tried it, but i think i'm gonna be happy to change it up too

PR #185, opened 2018-02-09
main -> div

brendansudol:
The HTML5 `main` element is not supported by Internet Explorer, which was throwing off the container width and auto margin.

This should fix :)

More details: https://stackoverflow.com/a/35820454

### DIFF
--- a/web/src/components/App.js
+++ b/web/src/components/App.js
@@ -29,7 +29,7 @@ const App = () => (
       </NavLink>
     </Toolbar>
 
-    <Container is="main">
+    <Container>
       <Switch>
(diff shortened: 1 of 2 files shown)


mgwalker:
Do we have a way of doing visual tests in IE other than pulling down virtual machines?  I'm guessing _not_, but that'd be a super cool CI service - push some HTML/JS/CSS to it, have it render screens in multiple browsers, and then post those into your Github PR.

brendansudol:
agreed that that service would be dope!

brendansudol:
we may have to make it ourselves ;)

PR #186, opened 2018-02-12
upgrade styled-components, rebass

brendansudol:
upgrades `styled-components` and `rebass` to latest releases (which fixes some bugs and improves performance).

while we're using these libraries (which may not be for too too much longer ;)), we should at least use the latest and greatest...

### DIFF
--- a/web/package.json
+++ b/web/package.json
@@ -37,7 +37,6 @@
   "dependencies": {
     "axios": "^0.17.1",
     "babel-polyfill": "^6.26.0",
-    "fsevents": "^1.1.3",
     "history": "^4.7.2",
     "lodash.kebabcase": "^4.1.1",
     "numeral": "^2.0.6",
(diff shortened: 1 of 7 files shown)


brendansudol:
the `prepare web dependencies cache` circle task is failing, which makes sense i think, because dependencies are changed, but how should this get resolved @mgwalker?

web/package.json
```diff
@@ -37,6 +37,7 @@
   "dependencies": {
     "axios": "^0.17.1",
     "babel-polyfill": "^6.26.0",
+    "fsevents": "^1.1.3",
```
mgwalker: `fsevents` probably shouldn't be in our list of dependencies.  It should get pulled in by our dependencies on an as-needed basis.  It only works on macOS, so the build is failing (and the deploy would fail too) because we declare that we require it but the build is in a Linux container.

brendansudol: ahhhh, thx!

PR #194, opened 2018-02-13
user login beginnings

brendansudol:
This PR begins work to have users authenticate via our front-end site.

It follows the standard/vanilla redux patterns & organization, and adds `redux-thunk` to handle the async authentication request.

Next up (after this initial chunk is live):

* Adding `ProtectedRoute`(s) that redirect to `Login` if unauthenticated
* Better error handling messaging in UI
* Updating top navbar to have dynamic login / logout links

### DIFF
--- a/web/package.json
+++ b/web/package.json
@@ -8,7 +8,8 @@
     "build": "npm run scaffold && NODE_ENV=production webpack -p",
     "lint": "eslint 'src/**/*.js'",
     "start": "webpack-dev-server --host 0.0.0.0 --port 8001 --history-api-fallback --content-base src/",
-    "test": "jest",
+    "test": "jest --collectCoverage",
+    "test-watch": "jest --watch",
     "prettier": "prettier --single-quote --trailing-comma none --write \"src/**/*.js\""
   },
   "repository": {
@@ -68,12 +69,13 @@
     "eslint-plugin-import": "^2.8.0",
     "eslint-plugin-jsx-a11y": "^6.0.2",
     "eslint-plugin-react": "^7.5.1",
+    "fetch-mock": "^6.0.0",
     "jest": "^21.2.1",
     "jest-styled-components": "^4.9.0",
     "prettier": "^1.9.1",
     "react-test-renderer": "^16.2.0",
     "redux-logger": "^3.0.6",
-    "redux-mock-store": "^1.4.0",
+    "redux-mock-store": "^1.5.1",
     "webpack": "^3.10.0",
     "webpack-dev-server": "^2.9.7"
   },
@@ -85,7 +87,6 @@
     ]
   },
   "jest": {
-    "collectCoverage": true,
     "coverageDirectory": "../coverage",
     "rootDir": "src",
     "setupFiles": [
--- a/web/package.json
+++ b/web/package.json
@@ -50,6 +50,7 @@
     "rebass": "^1.0.4",
     "redux": "^3.7.2",
     "redux-form": "^7.2.1",
+    "redux-thunk": "^2.2.0",
     "styled-components": "^2.3.3"
   },
   "devDependencies": {
--- a/web/package.json
+++ b/web/package.json
@@ -55,6 +55,7 @@
     "styled-components": "^2.3.3"
   },
   "devDependencies": {
+    "axios-mock-adapter": "^1.13.1",
     "babel-core": "^6.26.0",
     "babel-eslint": "^8.1.2",
     "babel-jest": "^21.2.0",
@@ -69,7 +70,6 @@
     "eslint-plugin-import": "^2.8.0",
     "eslint-plugin-jsx-a11y": "^6.0.2",
     "eslint-plugin-react": "^7.5.1",
-    "fetch-mock": "^6.0.0",
     "jest": "^21.2.1",
     "jest-styled-components": "^4.9.0",
     "prettier": "^1.9.1",
--- a/web/src/actions/auth.js
+++ b/web/src/actions/auth.js
@@ -1,0 +1,16 @@
+export const LOGIN_REQUEST = 'LOGIN_REQUEST';
+export const LOGIN_SUCCESS = 'LOGIN_SUCCESS';
+export const LOGIN_FAILURE = 'LOGIN_FAILURE';
+
+export const requestLogin = () => ({ type: LOGIN_REQUEST });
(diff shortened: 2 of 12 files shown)


brendansudol:
the flagged codeclimate complaint is silly

brendansudol:
code coverage slightly up! 💥

web/package.json
```diff
@@ -8,7 +8,8 @@
     "build": "npm run scaffold && NODE_ENV=production webpack -p",
     "lint": "eslint 'src/**/*.js'",
     "start": "webpack-dev-server --host 0.0.0.0 --port 8001 --history-api-fallback --content-base src/",
-    "test": "jest",
+    "test": "jest --collectCoverage",
+    "test-watch": "jest --watch",
```
mgwalker: I could add another docker container that runs jest and lint as it detects changes.  🤔  (It'd probably be an awful experience.)

web/package.json
```diff
@@ -50,9 +51,11 @@
     "rebass": "^1.0.4",
     "redux": "^3.7.2",
     "redux-form": "^7.2.1",
+    "redux-thunk": "^2.2.0",
```
mgwalker: :tada: I love redux-thunk.

web/package.json
```diff
@@ -84,7 +87,6 @@
     ]
   },
   "jest": {
-    "collectCoverage": true,
```
mgwalker: Ah, I see, disabling coverage for the `watch`.  Nifty!

web/src/actions/auth.js
```diff
@@ -0,0 +1,25 @@
+import axios from 'axios';
+
+const API_URL = 'http://localhost:8000';
```
mgwalker: In f6a7cc69, I added a webpack config to replace instances of `process.env.API_URL` with the actual value of that environment variable at build time, so this should work if it's changed to:

```js
const API_URL = process.env.API_URL;
```

The deploy scripts that CircleCI uses are setting the `API_URL` appropriately, based on which environment it's going to.

Feel free to push back on this as a global approach (maybe we should only inject environment variables into something like an `env.js` file?), but we definitely need some kind of solution for setting the API URL.

web/src/actions/auth.js
```diff
@@ -0,0 +1,25 @@
...
+export const requestLogin = () => ({ type: LOGIN_REQUEST });
+export const completeLogin = () => ({ type: LOGIN_SUCCESS });
+export const failLogin = error => ({ type: LOGIN_FAILURE, error });
+export const logout = () => ({ type: LOGOUT_SUCCESS });
+
+export const attemptLogin = (username, password) => dispatch => {
+  dispatch(requestLogin());
+
+  return axios
+    .post(`${API_URL}/auth/login`, { username, password })
```
mgwalker: Should we look into how to use `await/async` with webpack+babel?  It sure is fun.  😛

web/src/actions/auth.test.js
```diff
@@ -0,0 +1,64 @@
...
+
+      const expectedActions = [
+        { type: actions.LOGIN_REQUEST },
+        { type: actions.LOGIN_FAILURE, error: 'foo' }
+      ];
+
+      return store.dispatch(actions.attemptLogin()).then(() => {
+        expect(store.getActions()).toEqual(expectedActions);
+      });
+    });
```
mgwalker: Don't need a special test case for this, but I'm curious how axios handles a 5xx error.  Presumably that would also reject?

web/src/containers/Login.js
```diff
@@ -0,0 +1,78 @@
...
+}
+
+Login.propTypes = {
+  authenticated: PropTypes.bool.isRequired,
+  error: PropTypes.string.isRequired,
+  fetching: PropTypes.bool.isRequired,
+  attemptLogin: PropTypes.func.isRequired
+};
+
+const mapStateToProps = ({ auth }) => ({ ...auth });
```
mgwalker: Someday I'll be able to read these destructuring statements more quickly.

At the moment, though, this pattern (the `{ ...auth }` in particular) bothers me a little bit because it's unclear which properties are coming from the `auth` state and which are expected to be applied directly to the component.  In order to know what's going into the props, you have to go look at the associated reducer.  With `propTypes` at least the linter will tell you if you're using a prop that isn't declared, but that still doesn't help understand where it came from...

What do you think?  Should we be explicit about which properties we're taking from state?

web/src/containers/Login.js
```diff
@@ -0,0 +1,78 @@
...
+const mapDispatchToProps = { attemptLogin };
```
mgwalker: TIL `mapDispatchToProps` can be an object.  Neat!

web/src/reducers/auth.js
```diff
@@ -0,0 +1,29 @@
...
+const initialState = {
+  fetching: false,
+  authenticated: false,
+  error: ''
+};
+
+const auth = (state = initialState, action) => {
+  switch (action.type) {
+    case LOGIN_REQUEST:
+      return { ...state, fetching: true, authenticated: false, error: '' };
```
mgwalker: I _*love*_ this pattern.  I used to use [updeep](https://npm.im/updeep) to do deep cloning and it was kind of a hassle.  This is so much nicer.

mgwalker:
I think product (cc: @jeromeleecms @NAretakis) and design (cc: @quarterback @lauraponce) should be aware of this one.  Once it's up, I imagine both will have thoughts about it.  😄

web/package.json
brendansudol: agreed -- i think it's the simplest and most understandable of the lot

web/src/actions/auth.js
brendansudol: oh cool, that sounds reasonable to me -- will change it over

web/src/containers/Login.js
brendansudol: i agree with you completely -- this is me being lazy and not wanting to type. i appreciate you bringing it up -- i will update!

web/src/reducers/auth.js
brendansudol: 👍 🎉

web/src/actions/auth.test.js
brendansudol: it would, yes

web/src/actions/auth.js
brendansudol: good call!

PR #200, opened 2018-02-13
unwind rebass / styled-components upgrade

brendansudol:
for some reason, the newer `rebass` / `styled-components` is failing on the production build (though working fine on dev...) i can't figure out exactly why, so let's go back to the versions we were using before 🙄


brendansudol:
another reason why we may want to move off of rebass

PR #201, opened 2018-02-14
Fix database seed bug

mgwalker:
The database was being seeded with activities and roles using known IDs.  The reason for that is so we could easily map the activities into roles.  The *problem* is that the ID is an autoincremented field, but when you set the ID yourself, the database does not set the "next autoincrement value" accordingly.  What that means in practice is if you try to insert another row without specifying an ID and the next autoincremented ID happens to be one of the manually set IDs, it will reject the insert.

So instead of setting the IDs manually, this PR allows them to be set by the database.  It then queries the database to find out what the IDs are, and uses that to setup the roles/activities mappings.

### DIFF
--- a/api/seeds/01-roles-and-activities.js
+++ b/api/seeds/01-roles-and-activities.js
@@ -1,37 +1,60 @@
-const setupActivities = async activitiesTable => {
-  await activitiesTable.insert({ id: 1, name: 'view-all-apds' });
-  await activitiesTable.insert({ id: 2, name: 'comment-on-apd' });
-  await activitiesTable.insert({ id: 3, name: 'send-apd-to-state' });
-  await activitiesTable.insert({ id: 4, name: 'view-own-apds' });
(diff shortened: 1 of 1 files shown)


api/seeds/01-roles-and-activities.js
```diff
@@ -1,37 +1,60 @@
...
+// the knex table object.  The reason is that if you have an insert on
+// the knex table object and then try to do an insert, under the hood it
+// attempts the insert again, which is bad of course, but it also causes
+// a key constraint error.
+const insertAndGetIDs = async (knex, tableName, values) => {
+  const insert = Object.keys(values).map(name => ({ name }));
+  await knex(tableName).insert(insert);
+  const asInserted = await knex(tableName).select('*');
+
+  const idMapping = { };
```
brendansudol: for funsies, here's a liner to do this (converting arrays to objects), which i seem to encounter a lot:

`return Object.assign({}, ...asInserted.map(as => ({ [as.name]: as.id })))`

PR #203, opened 2018-02-14
Role management GET route

mgwalker:
Adds a route for getting a list of roles.  By navigating to `https://{api-url}/roles`, you can get a list of roles in JSON format.  For each role, it returns its internal ID, its name, and a list of activities (by name) that are currently associated with that role.  It looks like this:

```json
[{
  "id": 33,
  "name": "admin",
  "activities": [
    "view-roles",
    "create-roles",
    "edit-roles"]
}, {
  "id": 34,
  "name": "cms-reviewer",
  "activities": []
}, {
  "id": 35,
  "name": "state-submitter",
  "activities": []
}]
```

- [ ] Product has approved the experience (cc: @jeromeleecms @NAretakis for approval)

### DIFF
--- a/api/routes/roles/index.test.js
+++ b/api/routes/roles/index.test.js
@@ -15,12 +15,12 @@ tap.test('roles endpoint setup', async endpointTest => {
     getEndpoint.calledWith(app),
     'users GET endpoint is setup with the app'
   );
-  endpointTest.ok(
-    postEndpoint.calledWith(app),
-    'users POST endpoint is setup with the app'
-  );
-  endpointTest.ok(
-    putEndpoint.calledWith(app),
-    'users PUT endpoint is setup with the app'
-  );
+  // endpointTest.ok(
+  //   postEndpoint.calledWith(app),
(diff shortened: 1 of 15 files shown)


api/routes/roles/index.test.js
```diff
@@ -0,0 +1,26 @@
...
+  const postEndpoint = sinon.spy();
+  const putEndpoint = sinon.spy();
+
+  rolesIndex(app, getEndpoint, postEndpoint, putEndpoint);
+
+  endpointTest.ok(
+    getEndpoint.calledWith(app),
+    'users GET endpoint is setup with the app'
+  );
+  // endpointTest.ok(
```
brendansudol: what's the thinking here - to uncomment these once the post and put routes are filled in?

mgwalker: Yeah, I originally starting writing the code for all the endpoints in here, but then I realized I should probably do them one at a time.  So...  it's not as clean as I'd like, but it will be soon.  😛

PR #207, opened 2018-02-15
Add protected routes

brendansudol:
This PR adds protected routes to the front-end. If a user tries to go to a "protected" page and they have not authenticated, it redirects to login page and upon successful login takes you back to that page.

In time, we may want to add some custom messaging when this happens (i.e., "You need to log in to see that page...") 

### DIFF
--- a/web/.eslintrc.json
+++ b/web/.eslintrc.json
@@ -4,6 +4,7 @@
   "rules": {
     "arrow-parens": ["error", "as-needed"],
     "comma-dangle": [1, "never"],
+    "no-confusing-arrow": 0,
 
     "jsx-a11y/anchor-is-valid": [
       2,
--- a/web/src/components/App.js
+++ b/web/src/components/App.js
@@ -12,6 +12,7 @@ import ActivityGoals from '../containers/ActivityGoals';
 import ActivityOverview from '../containers/ActivityOverview';
 import ApdOverview from '../containers/ApdOverview';
 import Login from '../containers/Login';
+import PrivateRoute from '../containers/PrivateRoute';
 import StateContacts from '../containers/StateContacts';
 import StateStart from '../containers/StateStart';
 import StatePersonnel from '../containers/StatePersonnel';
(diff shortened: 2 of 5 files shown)


web/.eslintrc.json
```diff
@@ -4,6 +4,7 @@
   "rules": {
     "arrow-parens": ["error", "as-needed"],
     "comma-dangle": [1, "never"],
+    "no-confusing-arrow": 0,
```
mgwalker: Why would we disable this rule?  We're adding a lot of customization to our default lint ruleset.  I'm not a big fan of that.

brendansudol:
do any of our customizations specifically concern you @mgwalker? i'm not sure the total number of changes is a very important metric so long as each change is valid and reasonable

web/.eslintrc.json
brendansudol: keep in mind that most of these are because prettier is better than eslint :)

mgwalker: That, sir, is an opinion.  😉

brendansudol: re: `no-confusing-arrow`, this comes up in the render prop here:

```
  <Route
    {...rest}
    render={props =>
      authenticated ? (
        <Component {...props} />
      ) : (
        <Redirect
          to={{
            pathname: '/login',
            state: { from: props.location }
          }}
        />
      )
    }
  />
```

by disabling, we avoid having to include repetitive `return` statements and because prettier formats it nicely, i don't believe this is "confusing".

mgwalker:
Yes, several of them.

* Why are we disabling the check on confusing arrow statements?
* Why are we disabling indentation consistency checks?
* Why are we disabling parentheses placement consistency checks?
* ...etc...

I get that `prettier` "fixes" these things, but prettier is a formatter, not a linter.  If prettier really does fix the "confusing arrow statements," for example, then there shouldn't be any reason to disable the linter rule.  The linter is for verifying consistency within the code and (ideally) enforcing best practices.

There are a lot of quirky cases where Javascript becomes hard to read (and these quirks are why we always use semicolons, personal preferences aside), and the threshold for "hard to read" is entirely subjective.  I'm in favor of a more conservative approach that errs on the side of making the code more explicit.  While I agree that the arrow function you pointed out is not especially confusing, I honestly don't think it's glanceable, either.  Wrapping the entire function body in parens should satisfy the lint rule, not introduce those extra return statements, and, in my opinion, be more clear.

brendansudol:
the problem with that suggestion (adding a parens) is that prettier will remove it because it's not necessary, and while i may agree that would perhaps look more clear, subjectively speaking, i much prefer the objectiveness of prettier's decision-making (which frees us up to focus on other things)

brendansudol:
in answer to your questions, we're disabling those rules because those rules actually add some subjectivity and lack of consistency that can be rooted out by disabling and using prettier instead

nkkl:
It sounds like the main risk here is that a future developer on this project might have a hard time understanding what's happening. I think it's an important sustainability thing for whoever we hand off to to be able to onboard to the codebase and understand it easily – how would y'all feel about setting up an ultralightweight readability test with a few folks (ideally with a mix of experience levels) who aren't familiar with our codebase? Just to get a smell test on the understandability.

nkkl:
I also suspect that there will be an upper limit to the number of customizations that a new team can easily grok, but I don't know what that limit is. It wouldn't be a terrible thing to set a boundary and try to stay within it.

mgwalker:
We took up the conversation on Slack so it'd be a little quicker and we reached a good agreement, I think.  Rather than customize the linting rules more and more, we are instead going to incorporate a second 3rd-party rules configuration that makes our existing rules work well with `prettier`.  That way, our customizations will be limited to just those that are necessary and unique for our project instead of customizations to make our two tools work together.

PR #209, opened 2018-02-15
Expenses form prototype

### LINKED ISSUE #151: Initial APD prototype - expenses

mgwalker:
Adds the expenses prototype forms.  This starts with a list of some common expenses (this is to get the submitter started, so they don't have to build a full list themselves), then drops into a full list that they can manually edit (this is for adding expenses that aren't in the common list).  From there, they can click to edit an expense (currently only the "Equipment and Supplies" start link works).

Will close #151.  Supersedes #173.

## Still needs:
- ~~needs to be inserted into the actual form workflow~~ let's do this in one PR, where the entire form is wired together at once - see #211
- ~~needs a basic a11y audit~~ going to do a more thorough a11y audit on the app as a whole after figuring out some general patterns that fit with our libraries and our needs


web/src/index.html
```diff
@@ -9,6 +9,6 @@
   <body>
     <noscript>You need to enable JavaScript to run this app.</noscript>
     <div id="app"></div>
-    <script src='app.js'></script>
+    <script src='/app.js'></script>
```
brendansudol: i don't understand why this is needed, but it also doesn't seem hurtful

mgwalker: Ah, it's because of the nested routes I had in the earlier version.  Not necessary for now, but when there were routes like `/expenses/list`, if you reloaded the page while at that route, the it would fail because it couldn't find `/expenses/app.js`.  So I made the path absolute.  😛

PR #210, opened 2018-02-16
Update CORS

brendansudol:
* sets origin to reflect the request origin
* now passes the `Access-Control-Allow-Credentials` header

on the front-end, we now have to explicitly set the `withCredentials` param to `true` for all axios requests. I created an axios instance that does this in `./util/api` (i wasn't sure where to put this)

### DIFF
--- a/api/endpoint-tests/auth/login.js
+++ b/api/endpoint-tests/auth/login.js
@@ -9,6 +9,18 @@ tap.test('login endpoint | /auth/login', async loginTest => {
     invalidTest.equal(response.statusCode, 400, 'gives a 400 status code');
   });
 
+  loginTest.test('proper response headers', async headerTest => {
+    const { response } = await request.post(url);
+    const { headers } = response;
+
+    headerTest.equal(headers.vary, 'Origin', 'vary origin');
+    headerTest.equal(
+      headers['access-control-allow-credentials'],
+      'true',
+      'allow credentials'
+    );
+  });
+
   const invalidCases = [
     {
       title: 'with a post body, but no username or password',
--- a/api/main.js
+++ b/api/main.js
@@ -10,7 +10,7 @@ const routes = require('./routes');
 const server = express();
 
 server.use(express.urlencoded({ extended: true }));
-server.use(cors());
+server.use(cors({ credentials: true, origin: true }));
 server.use(bodyParser.json());
 
 auth.setup(server);
--- a/web/src/util/api.js
+++ b/web/src/util/api.js
@@ -1,0 +1,5 @@
+import axiosClient from 'axios';
+
+const axios = axiosClient.create({ withCredentials: true });
+
+export default axios;
--- a/web/src/actions/auth.js
+++ b/web/src/actions/auth.js
@@ -1,5 +1,6 @@
-import axios from 'axios';
+import axiosClient from 'axios';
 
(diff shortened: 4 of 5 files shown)


api/endpoint-tests/auth/login.js
```diff
@@ -9,6 +9,18 @@ tap.test('login endpoint | /auth/login', async loginTest => {
...
+  loginTest.test('proper response headers', async headerTest => {
+    const { response } = await request.post(url);
+    const { headers } = response;
+
+    headerTest.equal(headers.vary, 'Origin', 'vary origin');
+    headerTest.equal(
+      headers['access-control-allow-credentials'],
+      'true',
+      'allow credentials'
+    );
```
mgwalker: Nice!  I'm curious if these two asserts should be part of all the login tests, but that's probably overkill.

api/main.js
```diff
@@ -10,7 +10,7 @@ const routes = require('./routes');
 const server = express();
 
 server.use(express.urlencoded({ extended: true }));
-server.use(cors());
+server.use(cors({ credentials: true, origin: true }));
```
mgwalker: :tada:

web/.eslintrc.json
```diff
@@ -4,6 +4,7 @@
   "rules": {
     "arrow-parens": ["error", "as-needed"],
     "comma-dangle": [1, "never"],
+    "no-confusing-arrow": 0,
```
mgwalker: Why would we disable this rule?

web/src/util/api.js
```diff
@@ -0,0 +1,5 @@
+import axiosClient from 'axios';
+
+const axios = axiosClient.create({ withCredentials: true });
+
+export default axios;
```
mgwalker: If we don't want to wrap up a single axios instance, we can also set `withCredentials` as a default option for all axios instances:

```
axios.defaults.withCredentials = true;
```

Then we'd have to find a home for *that*, but anyway...  😛

web/src/containers/Login.js
```diff
@@ -21,11 +21,12 @@ class Login extends Component {
...
 
   render() {
-    const { fetching, error, authenticated } = this.props;
+    const { authenticated, error, fetching, location } = this.props;
+    const { from } = location.state || { from: { pathname: '/' } };
     const { username, password } = this.state;
 
     if (authenticated) {
-      return <Redirect to="/" />;
+      return <Redirect to={from} />;
```
mgwalker: Nice!

web/src/util/api.js
brendansudol: yeah i saw that -- wasn't sure which one was better or where its home should be either!

mgwalker: I'm on board with leaving it here for now and we can figure out a better home when we're making more than one call.  😄

api/main.js
mgwalker: Now that I've thought about this a little more, and after doing some reading about what CORS is even supposed to be, I think the right solution from a security standpoint is to set the origin to a static value.  It'd probably have to come from an environment variable so we could vary it based on what was consuming it.  CORS still isn't super helpful, but 🤷‍♂️ 

Anyway, I'll open a backlog issue for that, this is perfectly fine for now.

Issue #18, opened 2017-12-13
Quotes from the Future exercise


nicoleslaw:
Imagine that we’ve shipped the MVP of the HITECH APD app. What are people saying about it?
Copy and paste the following template into the comment field below, and add your answers inline. (You can edit your comment after posting!)

```
## State users are saying...

## CMS analysts are saying...

## CMS leadership is saying...
```

jeromeleecms:
## State users are saying APD's, what's that? We just use this system and we get funding approval in less than two weeks.

## CMS analysts are saying the tools we have help us focus on helping states do their work and not on their administrative correctness. Why did it take so long for someone to create this?

## CMS leadership is saying, can we roll this out to every part of CMS? APDs complaints and turn around times are a thing of the past. The change happened overnight.

mgwalker:
## State users are saying...
* Gosh, this sure takes a lot less time to write!
* It's really great how I don't have to submit the same thing over and over.
* It's nice knowing exactly CMS wants.
  * (☝️  that seems less like an issue in HITECH than MMIS)

## CMS analysts are saying...
* Gosh, this sure takes a lot less time to review, and all the numbers add up!
* We can quickly see how much money we've already approved for a state.
* Having a consistent thread over time is really helpful.
* The states don't view us as a hurdle to clear anymore.

## CMS leadership is saying...
* Gosh, we can answer data calls really fast!
* It was a smart move to let Jerome and Nick lead this work, and we should totally extend it to E&E and MMIS.  (🤞)

nkkl:
## State users are saying...
- APDs get turned around so quickly now!
- We don't need full-time staff or vendors to write APDs anymore

## CMS analysts are saying...
- We have so much confidence in our state partners
- We get to spend more time helping states and less time enforcing regs and tracking numbers

## CMS leadership is saying...
- We're doing really well at customer service
- We know where every dollar is
- This should extend beyond HITECH

## State governors are saying...
- We get our Medicaid money so quickly now

mkhandekar:
## State users are saying...
* I have so much more time on my hands, I now get to spend it with my loved ones
* I don't have to keep doing the same things over and over
* In the end, we get to our money faster
* We get to see faster results for our state's residents!

## CMS analysts are saying...
* We have so much more time on my hands, I now get to spend it with my loved ones
* We am more in attuned to what states need and how to support them

## CMS leadership is saying...
* We're going to fund an extension of this work
[...]

PR #223, opened 2018-02-20
Role management POST endpoint

mgwalker:
Adds an API route for creating a new role.  The route is a `POST` to `/roles` and it expects a JSON object:

```json
{
  name: 'new role name',
  activities: [ 1, 2, 3 ]
}
```

...where the name must be unique and the values of the activities array must be numeric and correspond to IDs of activities known to the system. Will return a 400 if the body is invalid or 201 and the new role object if the creation is successful.


brendansudol:
these continual codeclimate false positives are getting quite annoying IMO

mgwalker:
Yeah.  It'd be cool if we could configure it, but unfortunately...  Are there are other code-quality services we could use instead, that's either configurable or maybe more simply doesn't have these checks that go beyond what our linter does?

api/routes/roles/post.js
```diff
@@ -0,0 +1,73 @@
...
+      `Activity IDs ${JSON.stringify(invalidIDs).replace(
+        /\[(.+)\]/,
+        '$1'
+      )} are invalid`
+    );
+  }
+
+  return {
+    name: role.name,
+    activities: role.activities.filter(id => !Number.isNaN(Number(id)))
```
brendansudol: is this filter necessary after the validation check on #18 checking for numbers only?

brendansudol:
i'm not aware of other/better ode-quality services, alas

api/routes/roles/post.js
mgwalker: 😂 Nope.  How silly.  I'll remove in a future PR - I need to run `prettier` on the whole API codebase.  I've finally setup Atom to run it automatically on every save so hopefully it won't be a big deal going forward, but still need to clean up the past...  😬

brendansudol: nice - the atom-prettier on save hook is pretty 🤘

PR #225, opened 2018-02-20
Activities API GET endpoint

mgwalker:
Adds an API endpoint to fetch a list of all activities.  Returns a list:

```js
[{
  id: 1,
  name: 'view-users'
}, {
  id: 2,
  name: 'view-roles'
}]
```

This should conclude the role management work for now.  Still need an endpoint to modify a user and add them to a role.


api/endpoint-tests/activities/get.js
```diff
@@ -0,0 +1,41 @@
...
+  });
+
+  getUsersTest.test('when authenticated', async validTest => {
+    const cookies = await login();
+    const { response, body } = await request.get(url, {
+      jar: cookies,
+      json: true
+    });
+
+    body.sort((activity1, activity2) => ((activity1.id < activity2.id) ? -1 : 1));
```
brendansudol: the sort here can be simplified to:

`body.sort((activity1, activity2) => (activity1.id - activity2.id))`

ref: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/sort

mgwalker: Oh, good catch!  This also deals with the "they're the same" case, which I was ignoring.  I hate that this sort is even necessary but tap's `same` test expects array elements to be in the same order.  I guess that makes sense a lot of times, but for our purposes, I don't think we care about the order. 🤷‍♂️

Issue #222, opened 2018-02-20
Don't allow users to change the activities associated with the role they are in


mgwalker:
Currently a user with the `edit-roles` permission can change which activities ***any*** role is associated with, meaning they could give their own role more permissions than it is supposed to have, or remove permissions.  Users should not be allowed to modify the permissions of roles they belong to.

PR #226, opened 2018-02-21
Role management DELETE endpoint

mgwalker:
Delete roles.  Sorry for the flood of these, @brendansudol.  They happen to all be pretty similar, which means I've been able to churn them out pretty fast.  There's really just one left (`PUT /users/:id`) which will give us all the user/role management API functionality we should need for a while.

### DIFF
--- a/api/endpoint-tests/roles/delete.js
+++ b/api/endpoint-tests/roles/delete.js
@@ -1,0 +1,66 @@
+const tap = require('tap'); // eslint-disable-line import/no-extraneous-dependencies
+const { db, request, getFullPath, login } = require('../utils');
+
+tap.test('roles endpoint | DELETE /roles/:roleID', async deleteRolesTests => {
+  await db().seed.run();
+
+  const url = getFullPath('/roles');
+
+  deleteRolesTests.test('when unauthenticated', async unauthenticatedTests => {
+    unauthenticatedTests.test('with invalid ID', async invalidTest => {
+      const { response, body } = await request.delete(`${url}/9001`);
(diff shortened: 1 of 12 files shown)


api/endpoint-tests/roles/delete.js
```diff
@@ -0,0 +1,72 @@
...
+    authenticatedTests.test('with an invalid ID', async invalidTest => {
+      const { response, body } = await request.delete(`${url}/9001`, {
+        jar: cookies
+      });
+      invalidTest.equal(response.statusCode, 404, 'gives a 404 status code');
+      invalidTest.notOk(body, 'does not send a body');
+    });
+
+    authenticatedTests.test(
+      'with an ID for a role that the user belongs to',
```
brendansudol: can u explain this test to me?

brendansudol: can u explain this test to me and why we want 401 here?

mgwalker: This is testing the case where a user attempts to delete the role that they're in.  They shouldn't be allowed to do that because they'd end up locking themselves out of the system, so it's a 401 (unauthorized).

This is related to #222 to go back and modify the PUT route to prevent users from modifying the roles they belong to, too.  I figured since I was writing this code at the time, I'd go ahead and fix it here.  But your question is a good one, and I think I should update the test name to be more descriptive.

brendansudol: ok, i see and that makes perfect sense -- thanks for explanation!

PR #246, opened 2018-02-23
Logging

mgwalker:
Logging [will be required for compliance](https://before-you-ship.18f.gov/infrastructure/logging/), so here's some initial logging pieces.  I've gone all-in on `silly` logging, which is stuff we shouldn't actually have to look at very often.  I've tried to be more conservative with the more significant logging levels.  For example, only unexpected errors get logged as errors; unauthenticated requests get logged as info; most everything else is verbose.

Our ISSO will be able to help us nail down what the log levels should be for different kinds of events.

I am using the [winston logging library](https://www.npmjs.com/package/winston).  It's configured to optionally log to the console and to a file.  These are controlled with the `LOG_CONSOLE` and `LOG_FILE` environment variables, respectively - set to a truthy value (see the [yn documentation](https://www.npmjs.com/package/yn)), each will be enabled.  The log level is controlled by the `LOG_LEVEL` environment variable.  By default, console logging is enabled, file logging is disabled, and the log level is `info`.

I added an override for the `log` function so we can capture request information:

screen shot 2018-02-23 at 10 55 34 am

The first thing is this request ID, which is created and attached to every incoming request to the API.  This way, as the request works its way through the code, we can keep a consistent thread in the logs.  In the screenshot above, a GET request arrived at `/roles` and that request was not authenticated.  We can see that both log messages have the same `requestID`.

The second thing I added was logging the user ID for authenticated requests:

screen shot 2018-02-23 at 10 56 29 am

This way we can track which user accounts are taking these actions.

  cc: @NAretakis @jeromeleecms.  Just a heads-up that this is here.  Feel free to check the little box if you're satisfied, or comment if you have feedback/questions/opinions for lunch.


api/db/user.js
```diff
@@ -7,7 +9,9 @@ module.exports = {
     },
 
     async activities() {
-      if (!this.relations.role) {
+      logger.silly('getting user activities');
+      if (!this.relations.role || !this.relations.role.activities) {
```
mgwalker: Fixed.  Needed to actually check that both of these exist, not just the higher-level one.

api/db/user.test.js
```diff
@@ -69,7 +69,8 @@ tap.test('user data model', async userModelTests => {
     activitesTests.test(
       'resolves a list of activites when the role relationship is already loaded',
       async alreadyLoadedTests => {
-        self.relations.role = true;
+        self.relations.role = {};
+        self.relations.role.activities = true;
```
mgwalker: Updated the test to support the more thorough check in the program code.

api/endpoint-tests/roles/get.js
```diff
@@ -23,6 +23,7 @@ tap.test('roles endpoint | GET /roles', async getUsersTest => {
       json: true
     });
 
+    body.sort((role1, role2) => role1.id - role2.id);
```
mgwalker: It feels silly to have to sort these things for tests, but 🤷‍♂️

api/logger.js
```diff
@@ -0,0 +1,58 @@
...
+        colorize: true,
+        label: name,
+        level: process.env.LOG_LEVEL,
+        prettyPrint: true,
+        silent: process.argv.includes('--silent'),
+        timestamp: true
+      };
+    } else {
+      transports.console = {
+        silent: true
```
mgwalker: If you don't define a console transport, it seems like you get a default one.  To prevent that, I attach a console transport that is silent if console logging should be disabled.

api/logger.js
```diff
@@ -0,0 +1,58 @@
...
+    let requestInfo = req;
+    if (req && typeof req !== 'string') {
+      requestInfo = {
+        requestID: req.id
+      };
+      if (req.user) {
+        requestInfo.userID = req.user.id;
+      }
+    }
+    winston.Logger.prototype.log.apply(logger, [level, ...args, requestInfo]);
```
mgwalker: I opted to reverse the log message and the request info because it looks cleaner in the output.  The result is a log line that looks about like this:

```
2018-02-23T16:59:49.680Z - silly: [auth middleware] got a can middleware request for [create-roles]
{ requestID: 'f57e02e0-18ba-11e8-9ad0-893dd6149dba' }
```

api/routes/activities/index.js
```diff
@@ -1,14 +1,20 @@
...
 module.exports = (app, ActivityModel = defaultActivityModel) => {
+  logger.silly('setting up GET /activities route');
   app.get('/activities', can('view-activities'), async (req, res) => {
+    logger.silly(req, 'handling GET /activities');
     try {
-      const activities = await ActivityModel.fetchAll({
+      const activities = (await ActivityModel.fetchAll({
         columns: ['id', 'name']
-      });
+      })).map(a => ({ id: a.get('id'), name: a.get('name') }));
```
mgwalker: Switched to using bookshelf's `get` methods in a few place instead of sending back the raw bookshelf object.  Sending back the bookshelf object is an act of trust that it will serialize the way we want it to.  This way we don't have to trust.  (And also, we can log the pure objects directly.)

api/routes/activities/index.test.js
```diff
@@ -63,7 +63,20 @@ tap.test('activities GET endpoint', async endpointTest => {
...
+        .returns(1)
+        .onSecondCall()
+        .returns(2);
+      get
+        .withArgs('name')
+        .onFirstCall()
+        .returns('hi')
+        .onSecondCall()
+        .returns('bye');
+      const activities = [{ get }, { get }];
```
mgwalker: Tests updated to reflect using the bookshelf `get` methods instead of returning the raw bookshelf objects.

api/routes/roles/put.test.js
```diff
@@ -68,7 +68,7 @@ tap.test('roles POST endpoint', async endpointTest => {
       'rejects invalid role objects...',
       async validationTest => {
         validationTest.beforeEach(async () => {
-          const fetch = sinon.stub.resolves(true);
+          const fetch = sinon.stub().resolves(true);
```
mgwalker: I don't know how this ***ever*** worked.

api/logger.js
brendansudol: nice!

brendansudol: #hefancy

api/routes/activities/index.js
brendansudol: 👍

PR #248, opened 2018-02-23
css - back to basics

brendansudol:
This PR removes `styled-components` and `rebass`, a CSS-in-JS solution / library for styling web apps. While it provided nice drop-in UI components, it was making life difficult for a number of reasons: 1) customization / debugging was difficult (and seemingly required use of private / non-documented API methods to work properly); 2) lack of compose-ability leading to a lot of nested component wrapping (i.e., `<Border><Relative><Absolute><Link>...`); 3) a worry that the learning curve for new people building on top of our app / updating things would be frustratingly high.

In its place, I added a more traditional, though still modern, approach to use PostCSS (and its family of plugins) to process / bundle / minify our CSS; it supports css variables, nesting, adding vendor prefixes, etc. It's like a more customizable, faster, javascript-based SASS (https://ashleynolan.co.uk/blog/postcss-a-review).

As a foundational coat of paint, we're leveraging Basscss (http://basscss.com/ace/ to be exact). This is an atomic/utility class library that helps alleviate a bunch of issues that come up in Crazy CSS land -- recommended reading: http://mrmrs.github.io/writing/2016/03/24/scalable-css/, https://brendansudol.com/decks/atomic-css/#/5) 

I think this switch will help our velocity in terms of UI iteration as well improve our chances of a successful and confident project pass-off.

- [ ] ~~Design has approved the experience~~
- [ ] ~~Product has approved the experience~~


brendansudol:
i'm sorry for how ginormous this PR is @mgwalker -- it pains me how many lines are changed

mgwalker:
A lot of those are in `package-lock.json`, so don't worry too much about it.  😄

web/src/components/FormStateContacts.js
```diff
@@ -1,48 +1,51 @@
...
         <SectionHeader>Contact #{idx + 1}:</SectionHeader>
-        <Flex wrap mx={-2}>
-          <Box px={2} w={[1, 1 / 3]}>
+        <div className="clearfix mxn1">
+          <div className="col col-12 sm-col-4 px1">
             <Field name={`${contact}.name`} component={Input} label="Name" />
-          </Box>
-          <Box px={2} w={[1, 1 / 3]}>
+          </div>
+          <div className="col col-12 sm-col-4 px1">
```
mgwalker: I'm clearly going to have to go read up on basscss.  I'm guessing this indicates a grid layout, with the typical 12 columns, where at the largest breakpoint the div uses all 12 columns but at the smallest breakpoint is uses 4?  (Are there fewer columns at the different breakpoints?

web/src/containers/ActivitiesList.js
```diff
@@ -30,33 +18,33 @@ const activities = [
...
+              <Link to={href}>{status}</Link>
+            )}
+          </div>
+          {name}
+        </div>
       ))}
-    </Box>
-    <ButtonOutline mr={2} onClick={() => goTo('/activities-start')}>
+    </div>
+    <button
```
mgwalker: For the sake of maintainability, should we consider creating a set of wrappers around these low-level components that we style the same way in various parts of the app?  We might end up with something like...

```js
import { BlueOutlineButton } from '../buttons'; // ignore the name, it's just to sketch the idea

...

<BlueOutlineButton onClick={}>Back</BlueOutlineButton>
```

I'm just thinking about the effort involved if we decide to change how those buttons look.  We'd have to change it in several places right now, which is both extra time and extra risk that we'll forget some.

web/src/styles/ButtonOutline.js
```diff
@@ -1,27 +0,0 @@
...
-
-      &:active {
-        background-color: ${hex};
-        box-shadow: inset 0 0 0 2px ${hex}, inset 0 0 8px ${darken(1 / 4)};
-      }
-    `;
-  }};
-`;
-
-export default ButtonOutlineNew;
```
mgwalker: Ha.  Just after I proposed something similar, here it is being removed.  😆

PR #251, opened 2018-02-27
state profile w/ real data

brendansudol:
Alright, this isn't close to perfect, but it's working and probably good enough for initial review. It fetches user (profile) data from api, populates form, and updates info upon form submission.

(FYI, the user reducer is especially raw.)

### DIFF
--- a/api/routes/users/put.js
+++ b/api/routes/users/put.js
@@ -6,7 +6,7 @@ const can = require('../../auth/middleware').can;
 
 const userIsNew = async (email, UserModel) => {
   const user = await UserModel.where({ email }).fetch();
-  return user ? false : true;
+  return !user;
 };
 
 // TODO: figure out better/cleaner solution.
--- a/api/routes/users/put.js
+++ b/api/routes/users/put.js
@@ -6,12 +6,12 @@ const can = require('../../auth/middleware').can;
 
 const userIsNew = async (email, UserModel) => {
   const user = await UserModel.where({ email }).fetch();
-
-  if (user) {
-    return false;
-  }
-  return true;
+  return user ? false : true;
 };
+
+// TODO: figure out better/cleaner solution.
+// At the very least, move this helper to a better location
+const toObj = entry => JSON.parse(JSON.stringify(entry));
 
 module.exports = (
   app,
@@ -25,7 +25,10 @@ module.exports = (
     logger.silly(req, `attempting to update user [${req.params.id}]`);
 
     try {
-      const user = await UserModel.where({ id: req.params.id }).fetch();
+      const user = await UserModel.where({ id: req.params.id }).fetch({
+        columns: ['id', 'email', 'name', 'position', 'phone', 'state']
+      });
+
       if (!user) {
         logger.verbose(req, `no user found for [${req.params.id}]`);
         return res.status(404).end();
@@ -73,7 +76,7 @@ module.exports = (
 
       await user.save();
       logger.silly(req, 'all done');
-      return res.status(204).end();
+      return res.send(toObj(user));
     } catch (e) {
       logger.error(req, e);
       return res.status(500).end();
--- a/api/routes/users/put.test.js
+++ b/api/routes/users/put.test.js
@@ -183,7 +183,7 @@ tap.test('users PUT endpoint', async endpointTest => {
           save.calledAfter(set),
           'the model is called after values are set'
         );
-        validTest.ok(res.status.calledWith(204), 'HTTP status set to 200');
+        validTest.ok(res.status.notCalled, 'HTTP status is not explicitly set');
       }
     );
 
@@ -221,7 +221,7 @@ tap.test('users PUT endpoint', async endpointTest => {
           save.calledAfter(set),
           'the model is called after values are set'
         );
-        validTest.ok(res.status.calledWith(204), 'HTTP status set to 200');
+        validTest.ok(res.status.notCalled, 'HTTP status is not explicitly set');
       }
     );
 
@@ -255,7 +255,7 @@ tap.test('users PUT endpoint', async endpointTest => {
           save.calledAfter(set),
           'the model is called after values are set'
         );
-        validTest.ok(res.status.calledWith(204), 'HTTP status set to 200');
+        validTest.ok(res.status.notCalled, 'HTTP status is not explicitly set');
       }
     );
   });
--- a/api/routes/users/put.test.js
+++ b/api/routes/users/put.test.js
@@ -184,6 +184,7 @@ tap.test('users PUT endpoint', async endpointTest => {
           'the model is called after values are set'
         );
         validTest.ok(res.status.notCalled, 'HTTP status is not explicitly set');
+        validTest.ok(res.send.calledWith({}), 'updated user data is sent');
       }
     );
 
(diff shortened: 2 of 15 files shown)


mgwalker:
Created #252 to compliment this.

web/src/actions/user.test.js
```diff
@@ -0,0 +1,119 @@
...
+      ];
+
+      return store.dispatch(actions.fetchUser()).then(() => {
+        expect(store.getActions()).toEqual(expectedActions);
+      });
+    });
+
+    it('creates GET_USER_FAILURE after unsuccessful user fetch', () => {
+      const store = mockStore({});
+      fetchMock.onGet().reply(403, 'foo');
```
mgwalker: Somewhere down the line, we might want to consider catching 401 and 403 statuses and treating them a little differently since the status codes in those cases tell us everything we need to know - not logged in, and don't have permission, respectively.

web/src/actions/user.test.js
```diff
@@ -0,0 +1,119 @@
...
+        { type: actions.UPDATE_USER_REQUEST },
+        { type: actions.UPDATE_USER_FAILURE, error: 'foo' }
+      ];
+
+      return store.dispatch(actions.updateUser()).then(() => {
+        expect(store.getActions()).toEqual(expectedActions);
+      });
+    });
+  });
+});
```
mgwalker: No test for `fetchUserDataIfNeeded`?  😄

I think it'd only need to check that the `GET_USER_REQUEST` action is dispatched or not, depending on the state, since that's the only unique logic.  For the correct functioning of everything else, we can rely on the `GET_USER_SUCCESS` and `GET_USER_FAILURE` tests up above.

web/src/reducers/user.js
```diff
@@ -0,0 +1,48 @@
...
+      return { ...state, fetching: true, error: '' };
+    case GET_USER_SUCCESS:
+      return {
+        ...state,
+        fetching: false,
+        loaded: true,
+        data: { ...action.data }
+      };
+    case GET_USER_FAILURE:
+      return { ...state, fetching: false, error: action.error };
```
mgwalker: Should this set `loading` to `false`?  I'm not feeling it strongly, but I'm thinking about a case where the user was loaded in the past and we attempt to load them again and it fails.  Is that a problem?  :shrug:

web/src/components/FormStateStart.js
```diff
@@ -23,42 +23,33 @@ const FormStateStart = ({ handleSubmit, pristine, reset, submitting }) => (
...
+      </button>
+    </div>
   </form>
 );
 
 FormStateStart.propTypes = {
   handleSubmit: PropTypes.func.isRequired,
-  pristine: PropTypes.bool.isRequired,
-  reset: PropTypes.func.isRequired,
   submitting: PropTypes.bool.isRequired
```
mgwalker: ❗️ Where does `submitting` come from?  Should it be wired up from the StateStart container?

api/routes/users/put.js
```diff
@@ -6,13 +6,13 @@ const can = require('../../auth/middleware').can;
...
-  if (user) {
-    return false;
-  }
-  return true;
+  return !user;
 };
 
+// TODO: figure out better/cleaner solution.
+// At the very least, move this helper to a better location
+const toObj = entry => JSON.parse(JSON.stringify(entry));
```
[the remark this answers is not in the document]
brendansudol: yes!

api/routes/users/put.test.js
```diff
@@ -183,7 +183,7 @@ tap.test('users PUT endpoint', async endpointTest => {
           save.calledAfter(set),
           'the model is called after values are set'
         );
-        validTest.ok(res.status.calledWith(204), 'HTTP status set to 200');
+        validTest.ok(res.status.notCalled, 'HTTP status is not explicitly set');
```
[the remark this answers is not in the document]
brendansudol: ah, nice, will do

web/src/actions/user.js
```diff
@@ -0,0 +1,58 @@
+import axios from '../util/api';
+
+const API_URL = process.env.API_URL || 'http://localhost:8000';
```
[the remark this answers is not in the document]
brendansudol: oh neat! much cleaner

web/src/actions/user.js
```diff
@@ -0,0 +1,58 @@
...
+
+export const requestUser = () => ({ type: GET_USER_REQUEST });
+export const receiveUser = data => ({ type: GET_USER_SUCCESS, data });
+export const failUser = error => ({ type: GET_USER_FAILURE, error });
+
+export const requestUserUpdate = () => ({ type: UPDATE_USER_REQUEST });
+export const receiveUserUpdate = data => ({ type: UPDATE_USER_SUCCESS, data });
+export const failUserUpdate = error => ({ type: UPDATE_USER_FAILURE, error });
+
+export const fetchUser = id => dispatch => {
```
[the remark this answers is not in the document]
brendansudol: agreed and agreed - we're def assuming a fair amount of comfortability with es6 stuff

web/src/actions/user.test.js
```diff
@@ -0,0 +1,119 @@
...
+  });
+
+  describe('fetchUser (async)', () => {
+    afterEach(() => {
+      fetchMock.reset();
+    });
+
+    it('creates GET_USER_SUCCESS after successful user fetch', () => {
+      const store = mockStore({});
+      fetchMock.onGet().reply(200, { foo: 'bar' });
```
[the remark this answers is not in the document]
brendansudol: 👍

web/src/components/FormStateStart.js
brendansudol: auto-generated by `reduxForm` -- https://redux-form.com/7.2.3/docs/api/props.md/#-code-props-code-

web/src/reducers/user.js
brendansudol: yes it should

web/src/actions/user.test.js
brendansudol: agreed, needs to be added

web/src/components/FormStateStart.js
mgwalker: Oh, I should have searched harder.  That's really handy!

brendansudol:
updated those tests, but i'm not sure how to stub it properly -- i'm gonna have to see how u do it

mgwalker:
I think what you did makes sense.  The alternative is to modify what the `fetch` stub resolves, so that it either has serializable properties or has a `toJSON()` method.  E.g.:

```js
fetch.resolves({
  prop1: 'value1',
  prop2: 'value2',
  save,
  ...
});
```

...or...

```js
fetch.resolves({
  toJSON: () => ({
    prop1: 'value1',
    prop2: 'value2'
  }),
  save,
  ...
});
```

And then you'd check that those props were returned:

```js
validTest.ok(res.send.calledWith({
  prop1: 'value1',
  prop2: 'value2'
}), 'updated user data is sent');
```

Using the `toJSON()` method might be preferable because it also checks that the fetched object is converted to a pure object before being sent to the user.  We could go a step further and make `toJSON` a stub and actually include a test that it was called.  😛  I'll create an example PR in a couple minutes and we can discuss there whether we think it's good or overkill.

PR #258, opened 2018-03-01
Change how the database is updated on deploys

mgwalker:
The previous database update process had these problems:

- on deployment, database update scripts ("migrations") were not run
- on deployment, "initial" data was not loaded ("seeding")
- loading initial data always deleted all existing data from the database (by design)

This pull request modifies how migrations and seeding are handled:

#### deployment scripts are updated to always run migrations and seeds

The deployment scripts now include a call to execute a cloud.gov "task" that will run the migration and seed activities.

#### there are no "automatic" seeds any more

Previously, seeds for "delete everything" and "create activities and roles" were run any time the database was seeded.  That was problematic in staging, which is why we weren't running seeds at all.  Now, the "delete everything" and "create activities and roles" seeds are not run by default, but each deployment environment "opts in" to them if appropriate.

The development and testing environments will delete all existing data on seeding, but the staging environment will not delete anything.

#### the "create roles and activities" seed is more defensive

Previously, this seed inserted all of the activities, roles, and mappings between them that it knew about.  The database will not allow two activities (or roles) to have the same names, so if the data was not deleted first, this seed would cause a database integrity error.

Now, the seed checks existing activities and roles and only inserts the ones that are not already in the database.  It also checks existing mappings and only adds new ones.

- ~~Tests have been updated (and all tests are passing)~~


api/endpoint-tests/activities/get.js
```diff
@@ -27,14 +27,14 @@ tap.test('activities endpoint | GET /activities', async getUsersTest => {
...
-        { id: 7, name: 'view-activities' },
-        { id: 8, name: 'edit-users' }
+        { id: 1001, name: 'view-roles' },
+        { id: 1002, name: 'create-roles' },
+        { id: 1003, name: 'edit-roles' },
+        { id: 1004, name: 'delete-roles' },
+        { id: 1005, name: 'view-users' },
+        { id: 1006, name: 'add-users' },
+        { id: 1007, name: 'view-activities' },
+        { id: 1008, name: 'edit-users' }
```
mgwalker: Set these IDs high so that they won't conflict with the database's autoincrement values in the tests that involve inserting new rows.  As long as we don't insert 1000 rows during our tests, these should be fine.  We also reset the database at the start of each set of tests, so that should give us lots and lots of wiggle room.

api/seeds/development.js
```diff
@@ -7,5 +9,7 @@ exports.seed = async knex => {
   }
 
   // Call specific seeds from here.
+  await truncate.seed(knex);
+  await roles.seed(knex);
```
mgwalker: The development seed is opting-in to the "delete everything" and "create roles and activities" seeds.

api/seeds/staging.js
```diff
@@ -0,0 +1,11 @@
+const roles = require('./shared/roles-and-activities');
+
+exports.seed = async knex => {
+  // Don't seed this data if we're not in a development environment.
+  if (process.env.NODE_ENV !== 'staging') {
+    return;
+  }
+
+  // Call specific seeds from here.
+  await roles.seed(knex);
```
mgwalker: For staging, we don't want to delete anything - user accounts that were created, form data that was submitted, etc.  Instead, just update the roles and activities to add anything new that we've created.

api/seeds/test/roles.js
```diff
@@ -1,57 +1,52 @@
 exports.seed = async knex => {
-  // For testing, throw out the standard stuff and put in our
-  // hard-coded stuff so we know what IDs to expect.
-  await knex('auth_role_activity_mapping').del();
-  await knex('auth_roles').del();
-  await knex('auth_activities').del();
```
mgwalker: Because role creation was automatic, the test seed previously had to delete them again so it could insert them with known IDs.  Now role creation is opt-in and the test seed doesn't opt-in, so it can just create the data it needs.

bin/deploy-ux-testing.sh
```diff
@@ -30,3 +30,6 @@ cd ..
 # Log into CF and push
 cf login -a $API -u $CF_USER_UX -p $CF_PASSWORD_UX -o $ORG -s $SPACE
 cf push -f manifest-ux-testing.yml
+
+# Migrate and seed the database
+cf run-task hitech-apd-api "npm run migrate && npm run seed"
```
mgwalker: We need to come up with a way to tell the UX environment to delete everything from its database on a deployment.  I'd like for it to be something we can do without adding deployed code - that is, I don't want code deployed that will wipe the database.  That seems like it'd be too ripe for abuse (or even an accidental click).

I'm thinking if we install `postgresql-client` as part of this script, we might be able to take advantage of `cf ssh` somehow by mapping the remote database into this script and then connecting and...  blah blah, but that feels less than ideal...

api/endpoint-tests/activities/get.js
brendansudol: nice

api/seeds/staging.js
brendansudol: niiice

bin/deploy-ux-testing.sh
brendansudol: how about a command but you have to type out something like DANGEROUSLY_DELETE_DATA for it to run?

mgwalker: What if, as part of the staging/production deploy script, we delete the file that deletes everything?  Then it just can't run.  😛  We could have another env var called something like `EMPTY_DATABASE` or something that told the staging seed file to delete stuff.  That way if someone accidentally sets that env var in staging/production, it'll just fail because the file won't be there.

PR #271, opened 2018-03-06
prettier tweak

brendansudol:
Sets `arrow-parens` option to `always` (https://prettier.io/docs/en/options.html#arrow-function-parentheses); I believe this style is preferred by @mgwalker (if not true, feel free to zap this PR!)

Also, I removed a file that we shouldn't have committed (`package-lock` in root)

- !The experience passes a basic manual accessibility audit (keyboard nav, screenreader, text scaling) OR an exemption is documented~


mgwalker:
I actually prefer not having the parens around single arguments, but had started using them because that's what our eslint rules require.  Now that we've adopted different rules...  😂

PR #275, opened 2018-03-07
better login fail handing

brendansudol:
* Now checks for `error.response` (network errors did not have a response so this was a throwing an error inside the catch block)

* A more generic and better styled error message if something failed while logging in. (We can abstract this into a `Message` or `Alert` component once we have a couple more examples and design guidance)

Once we have better custom error messages, we can sub those in easily into the UI via the `auth.error` store value.

Preview:
login-fail


brendansudol:
cc: @lauraponce

lauraponce:
@brendansudol is the plan to eventually use the USWDS styles for the form elements? Either way is fine, I was just wondering! 

I like where this is going, is there a way to soften the red notification and tie it to the fields? Maybe by outlining the form fields to show where there is an issue and displaying the message in red text without the fill? 

screen shot 2018-03-08 at 9 43 09 am

brendansudol:
yes, we can cherrypick any of the USWDS styles we like, as well as change / supplement with our own.

Yes, I definitely agree about having per-field notifications (vs. the one big, vague global one at the top)

That is work we should absolutely slot in shortly -- for now, I'm just hoping to slightly improve what's currently there.

brendansudol:
Softened the red, and made more in line with uswds:

screen shot 2018-03-08 at 10 06 13 am

brendansudol:
If you're comfortable with this direction / improvement @lauraponce, you'll have to explicitly "approve" this pull request before it gets added to the site. If you're not comfortable, that's fine too (and we can iterate)!

Issue #188, opened 2018-02-12
Create navigational anchor to give users a sense of where they are in the process


nkkl:
In usability testing, participants were distracted by not knowing where in the process they were. We should add some structure so that participants are more comfortable and have a sense of where they in the process and how what they're working on fits in with everything else in the submission.

### This task is done when...
 - [x] One or more potential designs have been sketched
 - [x] A provisional design has been chosen
 - [x] Dev has implemented the chosen design, so that it can be tested

lauraponce:
Moving this to in progress and am working on a wireframe for the nav

lauraponce:
Here is the first pass at a wireframe for the nav based on the USWDS. The actual content for the nav items is just placeholder until we work through that a bit more. @quarterback @brendansudol open to your thoughts on this!

<img width="584" alt="screen shot 2018-03-05 at 9 36 39 am" src="https://user-images.githubusercontent.com/21222704/36980595-c9982fd4-2058-11e8-8cac-ac51487e2d94.png">

nkkl:
@NAretakis @jeromeleecms I've seen the future and it's pretty swanky. ;)

quarterback:
@lauraponce I like that, Laura.

PR #282, opened 2018-03-07
nav/header ui iterate

brendansudol:
Iterates on nav design/layout per https://github.com/18F/cms-hitech-apd/issues/188

Adds icon library (font awesome) and using explicit importing (https://github.com/FortAwesome/react-fontawesome#usage) to keep our bundle size down.

Rejiggers location of `TopNav` so that sidebar span entire height (per mockup)

Hides sidebar (on mobile view) until we figure out how we want to handle that. (can still navigate via back and continue buttons)

Preview:
header


brendansudol:
For consistency with where the "log in" button/link is, I kept the "log out" link in the same place (vs. at the bottom of sidebar) -- how do you feel about that?

mgwalker:
Pinging @lauraponce!  😄

lauraponce:
I feel good about that, looks good to me! 👍

web/src/containers/TopNav.js
```diff
@@ -3,6 +3,13 @@ import React, { Component } from 'react';
 import { Link } from 'react-router-dom';
 import { connect } from 'react-redux';
 
+// to optimize bundle, explicitly importing only the icons used
+// TODO: once we're using icons more, move these to a central location
```
mgwalker: We should probably add the Uglify plugin to WebPack to do tree-shaking, too...

web/src/containers/TopNav.js
```diff
@@ -24,15 +31,25 @@ class TopNav extends Component {
...
-              <button type="button" className="btn" onClick={this.handleLogout}>
-                Log out
+              <button type="button" className="btn h5 regular">
+                <span className="mr-tiny">Notifications</span>
+                <FontAwesomeIcon icon={faBell} />
+              </button>
+              <button type="button" className="btn h5 regular">
+                <span className="mr-tiny">Help</span>
+                <FontAwesomeIcon icon={faHelp} />
+              </button>
```
mgwalker: I'm curious if the help button should really only be accessible if you're logged in.  Maybe a question for @lauraponce?  If it should always be available, is it okay to create an issue to fix it so we can go ahead and merge this?

web/src/styles/app/spacing.css
```diff
@@ -0,0 +1,8 @@
+/*
+  we need a spacing unit that's smaller than mb1, mt1, etc.
+  could call it "tiny" or "05"; using the former for now
+*/
+
+.mr-tiny {
```
mgwalker: Mr. Tiny, the smallest man on his block.

web/package.json
```diff
@@ -36,6 +36,10 @@
   },
   "homepage": "https://github.com/18F/cms-hitech-apd#readme",
   "dependencies": {
+    "@fortawesome/fontawesome": "^1.1.4",
```
mgwalker: I was a little worried about what FA does to a11y, but it turns out they're handling that internally, and this React wrapper doesn't interfere with it!  :tada:  Checked it locally to make sure it was rendering properly and everything looks gravy.

brendansudol: i thought they might, given its popularity

PR #284, opened 2018-03-07
Hook up state Medicaid office and contact info to API

### LINKED ISSUE #278: UI to API end-to-end: state profile page 2

mgwalker:
This PR hooks up the second page (titled "State set-up 2" in the sidebar right now) to the API, so we can capture things like Medicaid office address, Medicaid director name and contact info, and additional state points of contact.

It required some changes to the JSON schema of the `medicaid_office` field in the states data model, and that's reflected in changes to the database seeds and the API tests.  I factored out the "invalid data" that is used by both the unit tests and endpoint tests so that they actually share it now, and it only has to be updated in one place (:tada:).

I also updated our shared axios instance to set the base URL, and the state action dispatcher now just uses relative URLs.  The associated tests are updated to enforce that the axios mock only responds to the expected URLs, which should make the tests a little more robust.

Since this work doesn't have a design element preceding it yet, I'm going to nix the design review requirement, but @lauraponce and @quarterback feel free to weigh in of course!

Closes #278.


mgwalker:
Good news is that 4,900 lines of that massive changeset is Jest snapshots.  😬

web/src/actions/state.js
```diff
@@ -0,0 +1,71 @@
...
+export const fetchState = id => dispatch => {
+  dispatch(requestState());
+
+  const url = `/states${id ? `/${id}` : ''}`;
+
+  return axios
+    .get(url)
+    .then(req => dispatch(receiveState(req.data)))
+    .catch(error => {
+      const reason = error.response.data || 'N/A';
```
brendansudol: we can't rely on `error.response` being there; because of that, i did `const reason = error.response ? error.response.data : 'N/A';` in `auth.js`, though now that this has come up again, i wonder if we should add a utility method (`handleApiError`?) and use that in all these catches (also because i'm sure we're gonna want to change the `"N/A"` bit at some point

web/src/actions/state.js
```diff
@@ -0,0 +1,71 @@
...
+    });
+};
+
+export const updateState = (id, data) => dispatch => {
+  dispatch(requestStateUpdate());
+
+  const url = `/states/${id}`;
+
+  if (data.medicaid_office) {
+    const { medicaid_office: { director } } = data;
```
brendansudol: for these nested ones, i think `const { director } = data.medicaid_office` is a little easier to read

brendansudol: feel free to agree to disagree

web/src/actions/state.test.js
```diff
@@ -0,0 +1,200 @@
...
+
+      const expectedActions = [{ type: actions.GET_STATE_REQUEST }];
+
+      await store.dispatch(actions.fetchStateDataIfNeeded());
+      expect(store.getActions()).toEqual(
+        expect.arrayContaining(expectedActions)
+      );
+    });
+
+    it('does not fetch data if it is already loaded', async () => {
```
brendansudol: nice

web/src/actions/user.js
```diff
@@ -37,11 +37,9 @@ export const updateUser = (id, data) => dispatch => {
   return axios
     .put(`${API_URL}/users/${id}`, data)
     .then(req => {
-      console.log('update user success!', req);
```
brendansudol: thx

mgwalker: I was just getting tired of the console output when I was running the tests.  😆

web/src/components/FormStateContacts.test.js
```diff
@@ -0,0 +1,109 @@
...
+      meta.submitFailed = true;
+
+      const component = shallow(<Contacts fields={fields} meta={meta} />);
+      expect(component).toMatchSnapshot();
+    });
+
+    it('adds a contact when the add button is pressed', () => {
+      const component = shallow(<Contacts fields={fields} meta={meta} />);
+
+      component.find('button.btn.btn-primary.bg-blue').simulate('click');
```
brendansudol: not now, but we may want to think about some "hook" classes (i.e., `js-add-btn`) so that this will be less brittle to new / updated utility classes.

https://coderwall.com/p/qktuzw/decouple-javascript-classes-from-css-ones-by-using-prefix

one class, one responsibility!

web/src/components/FormStateContacts.test.js
```diff
@@ -0,0 +1,109 @@
...
+
+      expect(fields.push.calledWith({})).toBe(true);
+    });
+
+    it('removes a contact when the remove button is pressed', () => {
+      const component = shallow(<Contacts fields={fields} meta={meta} />);
+
+      component
+        .find('button[title="Remove Contact"]')
```
brendansudol: ah, this could work too, though buttons don't need titles for accessibility (i think)

web/src/containers/StateContacts.js
```diff
@@ -1,38 +1,73 @@
...
 import FormStateContacts from '../components/FormStateContacts';
 import PageNavButtons from '../components/PageNavButtons';
 import withSidebar from '../components/withSidebar';
 import FormLogger from '../util/formLogger';
 
 class StateContacts extends Component {
-  showResults = data => {
-    console.log(data);
+  componentDidMount() {
+    // TODO: don't hardcode user id
```
brendansudol: comment doesn't apply

web/src/containers/StateContacts.js
```diff
@@ -1,38 +1,73 @@
...
+    this.props.fetchStateDataIfNeeded();
+  }
+
+  updateForm = data => {
+    this.props.updateState(this.props.state.data.id, data);
   };
 
   render() {
-    const { goTo } = this.props;
+    const { goTo, state } = this.props;
```
brendansudol: man, the `state` field on the `props` argument is messing with my mind :)

PR #293, opened 2018-03-09
start APD endpoints

brendansudol:
Adds `/apds` GET request; updates relationship between APDs and states

### DIFF
--- a/api/routes/apds/get.js
+++ b/api/routes/apds/get.js
@@ -1,0 +1,41 @@
+const logger = require('../../logger')('users route get');
+const defaultApdModel = require('../../db').models.apd;
+const defaultUserModel = require('../../db').models.user;
+const loggedIn = require('../../auth/middleware').loggedIn;
+
+module.exports = (
+  app,
+  ApdModel = defaultApdModel,
+  UserModel = defaultUserModel
+) => {
+  logger.silly('setting up GET /apds route');
+
+  app.get('/apds', loggedIn, async (req, res) => {
+    logger.silly(req, 'handling GET /apds');
+
+    try {
+      // TODO: we should have a helper method that fetches
+      // and returns state/stateId given userId
+      const user = await UserModel.where({ id: req.user.id }).fetch({
+        withRelated: ['state']
+      });
+      const state = user.related('state');
+      const stateId = state.get('id');
+
+      if (!stateId) {
+        logger.verbose('user does not have an associated state');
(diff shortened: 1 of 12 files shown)


api/routes/apds/get.js
```diff
@@ -0,0 +1,41 @@
...
+  UserModel = defaultUserModel
+) => {
+  logger.silly('setting up GET /apds route');
+
+  app.get('/apds', loggedIn, async (req, res) => {
+    logger.silly(req, 'handling GET /apds');
+
+    try {
+      // TODO: we should have a helper method that fetches
+      // and returns state/stateId given userId
```
mgwalker: I suppose there's no reason we couldn't attach the user model directly to the request object.  Then if we add a helper method to the user model class, we could have something like `await req.user.model.getState()`?

api/routes/apds/get.js
```diff
@@ -0,0 +1,41 @@
...
+      const state = user.related('state');
+      const stateId = state.get('id');
+
+      if (!stateId) {
+        logger.verbose('user does not have an associated state');
+        return res.status(401).end();
+      }
+
+      const whereCondits = { state_id: stateId };
+      const apds = (await ApdModel.where(whereCondits).fetchAll()).toJSON();
```
mgwalker: Do we want to restrict this fetch to just specific columns?  I'm guessing there's nothing in our data model that users shouldn't see right now but 🤷‍♂️

api/routes/apds/get.js
brendansudol: yes, i like that!

api/routes/index.js
```diff
@@ -10,10 +11,11 @@ const openAPI = require('./openAPI');
 module.exports = (
   app,
   activitiesEndpoint = activities,
+  apdsEndpoint = apds,
   rolesEndpoint = roles,
   statesEndpoint = states,
   usersEndpoint = users,
-  formLoggerEndopint = formLogger,
+  formLoggerEndpoint = formLogger,
```
mgwalker: What have you got against endo-pints?

api/routes/apds/get.js
brendansudol: yeah, nothing worried me in this table (at this time), but perhaps we always want to be explicit

brendansudol:
good eye re: endpoint tests -- gonna batch those up at the end!

Issue #279, opened 2018-03-07
UI to API end-to-end: APD overview


brendansudol:

brendansudol:
deprecated in favor of newer, more targeted issues

PR #309, opened 2018-03-13
wire up apd overview front-end

brendansudol:
Allows user to add/edit/save their program vision and benefits.

### DIFF
--- a/api/routes/states/put.js
+++ b/api/routes/states/put.js
@@ -64,7 +64,12 @@ module.exports = (app, dataHelper = defaultDataHelper) => {
       logger.silly(req, newData);
 
       logger.silly(req, 'validating program benefits');
-      if (joi.string().validate(newData.program_benefits).error) {
+      if (
+        joi
+          .string()
+          .allow('')
+          .validate(newData.program_benefits).error
+      ) {
         logger.verbose(req, 'invalid program benefits');
         return res
           .status(400)
@@ -73,7 +78,12 @@ module.exports = (app, dataHelper = defaultDataHelper) => {
       }
 
       logger.silly(req, 'validating program vision');
-      if (joi.string().validate(newData.program_vision).error) {
+      if (
+        joi
+          .string()
+          .allow('')
+          .validate(newData.program_vision).error
+      ) {
         logger.verbose(req, 'invalid program vision');
         return res
           .status(400)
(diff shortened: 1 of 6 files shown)


api/routes/states/put.js
```diff
@@ -64,7 +64,12 @@ module.exports = (app, dataHelper = defaultDataHelper) => {
       logger.silly(req, newData);
 
       logger.silly(req, 'validating program benefits');
-      if (joi.string().validate(newData.program_benefits).error) {
+      if (
+        joi
+          .string()
+          .allow('')
```
brendansudol: empty string was failing validation, and we want to accept it if user wants to clear out their prior input.

https://github.com/hapijs/joi/issues/448

brendansudol:
This should close: https://github.com/18F/cms-hitech-apd/issues/279

api/routes/states/put.js
mgwalker: Interesting!

web/src/components/FormApdOverview.js
```diff
@@ -5,27 +5,38 @@ import { Field, reduxForm } from 'redux-form';
...
 FormApdOverview.propTypes = {
-  handleSubmit: PropTypes.func.isRequired
+  handleSubmit: PropTypes.func.isRequired,
+  submitting: PropTypes.bool.isRequired
 };
 
 const formConfig = {
   form: 'apdOverview',
-  destroyOnUnmount: false
+  destroyOnUnmount: false,
```
mgwalker: I wonder if we can get rid of the `destroyOnUnmount` for the forms that are wired up now.  We take the submitted data, send it to the API, then update the app state with the returned data.  That is then used to set the `initialValues`, so if the form is destroyed when we move on to the next section, it's no big deal - it gets rebuilt with the data the API sent back after saving previously.  Is that right?  I'm not sure if it's a big deal, but I assume there's a reason it defaults to destroying the forms. 🤷‍♂️

web/src/containers/ApdOverview.js
```diff
@@ -1,42 +1,79 @@
...
 ApdOverview.propTypes = {
-  goTo: PropTypes.func.isRequired
+  goTo: PropTypes.func.isRequired,
+  fetchStateDataIfNeeded: PropTypes.func.isRequired,
+  updateState: PropTypes.func.isRequired,
+  formData: PropTypes.object.isRequired,
+  stateInfo: PropTypes.object.isRequired
+};
+
+const mapStateToProps = ({ state }) => {
```
mgwalker: This is our state state.  👍

PR #314, opened 2018-03-14
Add API endpoints to create APD activities, goals, and objectives

mgwalker:
Creates all the API-side stuff to support creating and editing APD activities, goals, and objectives.

This PR got pretty big so I'm going to split it into a few.  This one adds the API changes, the next one will add the frontend changes, and then I will create more to add the handful of things missing:

* unit tests on the frontend
* endpoint tests on the backend
* OpenAPI documentation
* logging on the backend

## Endpoints

* POST `/apds/:apd_id/activities`
  Create a new activity on the APD specified by `apd_id`.  This just creates an activity with a name, no description.  This is based on the UI section for selecting and creating new activities - it begins with just names.  Descriptions are added in a separate form.  The activity name must be unique within the APD (that is, a single APD may not have multiple activities with the same name).

* PUT `/activities/:id`
  Updates the activity specified by `id`.  This lets you change the name or description for an existing activity.

* PUT `/activities/:id/goals`
  Add goals to the activity specified by `id`.  This will remove all goals and objectives associated with the activity and replace them with the new values.  I decided to do this because the UI is hooked up in such a way that it will send us all of the goals and objectives when it saves, and deleting and inserting new was easier than doing a diff and dancing around to figure out which ones had changed, etc.

## Models

* Added `toJSON` overrides for all the activity-related models (activity, goal, objective) to get rid of extraneous mapping IDs in most cases, but also to return objectives as just strings.  Since they cannot be updated individually, only replaced with goals, there was no reason to make them objects with IDs.  Similarly, no longer returns goal IDs since goals cannot be modified individually.

  - no updated endpoint tests, but this PR was already too big so I will add those in the next one 😬
  - will add OpenAPI documentation in an upcoming PR, because this thing is already enormous

### DIFF
--- a/api/db/activity.js
+++ b/api/db/activity.js
@@ -8,6 +8,15 @@ module.exports = () => ({
 
     goals() {
       return this.hasMany('apdActivityGoal');
+    },
+
+    toJSON() {
+      return {
+        id: this.get('id'),
+        name: this.get('name'),
+        description: this.get('description'),
+        goals: this.related('goals')
+      };
     }
   },
 
@@ -20,6 +29,13 @@ module.exports = () => ({
 
     objectives() {
       return this.hasMany('apdActivityGoalObjective');
+    },
+
+    toJSON() {
+      return {
+        description: this.get('description'),
+        objectives: this.related('objectives')
+      };
     }
   },
 
@@ -28,6 +44,10 @@ module.exports = () => ({
 
     goal() {
       return this.belongsTo('apdActivityGoal');
+    },
+
+    toJSON() {
+      return this.get('description');
     }
   }
 });
--- a/api/db/activity.js
+++ b/api/db/activity.js
@@ -3,7 +3,7 @@ module.exports = () => ({
     tableName: 'activities',
 
     apd() {
-      return this.hasOne('apd', 'id', 'apd_id');
+      return this.belongsTo('apd');
     },
 
     goals() {
--- a/api/routes/apds/activities/goals/put.js
+++ b/api/routes/apds/activities/goals/put.js
@@ -1,0 +1,109 @@
+const logger = require('../../../../logger')('apd activites route put');
+const {
+  apdActivity: defaultActivityModel,
+  apdActivityGoal: defaultGoalModel,
+  apdActivityGoalObjective: defaultObjectiveModel
+} = require('../../../../db').models;
+const { userCanEditAPD } = require('../../utils');
+const loggedIn = require('../../../../auth/middleware').loggedIn;
+
+module.exports = (
+  app,
+  ActivityModel = defaultActivityModel,
+  GoalModel = defaultGoalModel,
+  ObjectiveModel = defaultObjectiveModel
+) => {
+  logger.silly('setting up PUT /activities/:id/goals route');
+  app.put('/activities/:id/goals', loggedIn, async (req, res) => {
+    logger.silly(req, 'handling PUT /activities/:id/goals route');
+    logger.silly(
+      req,
+      `attempting to update goals on activity [${req.params.id}]`
+    );
+
+    try {
+      const activityID = +req.params.id;
+      const activity = await ActivityModel.where({ id: activityID }).fetch({
+        withRelated: ['apd', 'goals.objectives']
+      });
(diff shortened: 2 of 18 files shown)


api/db/activity.js
```diff
@@ -3,11 +3,20 @@ module.exports = () => ({
...
     apd() {
-      return this.hasOne('apd', 'id', 'apd_id');
+      return this.belongsTo('apd');
     },
 
     goals() {
       return this.hasMany('apdActivityGoal');
+    },
+
+    toJSON() {
```
brendansudol: nice, i like these new to_json methods

api/routes/apds/activities/goals/put.js
```diff
@@ -0,0 +1,98 @@
...
+            description: goal.description,
+            activity_id: activityID
+          });
+          await goalModel.save();
+          const goalID = goalModel.get('id');
+
+          if (Array.isArray(goal.objectives)) {
+            goal.objectives.forEach(async objective => {
+              if (typeof objective === 'string' && objective) {
+                const objectiveModel = ObjectiveModel.forge({
```
brendansudol: future optimization maybe: adding a helper method on `ObjectiveModel` to add objects; something like:

`addObjectives(objs, goal) { `await this.insert(objs.map(o => ({ description: o, activity_goal_id: goal }))`

api/routes/apds/activities/goals/put.test.js
```diff
@@ -0,0 +1,274 @@
...
+        GoalModel.forge.calledWith({ description: 'goal 1', activity_id: 1 }) &&
+          GoalModel.forge.calledWith({ description: 'goal 2', activity_id: 1 }),
+        'the two expected goals are created'
+      );
+      validTest.ok(
+        goal.save.calledBefore(ObjectiveModel.forge),
+        'goals are saved before objectives are created'
+      );
+      validTest.ok(
+        ObjectiveModel.forge.calledThrice,
```
brendansudol: great name

api/routes/apds/activities/post.js
```diff
@@ -0,0 +1,77 @@
...
+  logger.silly('setting up POST /apds/:apdID/activities route');
+  app.post('/apds/:apdID/activities', loggedIn, async (req, res) => {
+    logger.silly(req, 'handling POST /apds/:apdID/activities route');
+    logger.silly(
+      req,
+      `attempting to add an activity on apd [${req.params.apdID}]`
+    );
+
+    try {
+      const apd = await check404(req, res);
```
brendansudol: i like this `check404` function, but it's kinda weird (to me at least) that it returns either false or an apd instance (the latter of which surprises me given the function name), but i'm not sure of a better way without a duplicative lookup; i think maybe just a name tweak would help, is `check404AndGetApd` terrible?

mgwalker: I'm glad you caught that, I meant to go back and try to figure out a better name and/or return type.  The reason for `false` is just to be explicit that it's not returning an APD, but returning `null` would be fine too.  Then it could maybe just be `getApd()`?

api/routes/apds/activities/post.js
```diff
@@ -0,0 +1,77 @@
...
+
+      if (typeof req.body.name !== 'string' || req.body.name.length < 1) {
+        logger.verbose(req, 'Invalid activity name');
+        return res
+          .status(400)
+          .send({ error: 'add-activity-invalid-name' })
+          .end();
+      }
+
+      const existingNames = apd.related('activities').pluck(['name']);
```
brendansudol: nice

api/seeds/test/apds.js
```diff
@@ -0,0 +1,83 @@
...
+      activity_id: 1000
+    },
+    {
+      id: 1001,
+      description: 'Win a Nobel prize for physics',
+      activity_id: 1000
+    },
+    {
+      id: 1002,
+      description: 'Go on Ellen',
```
brendansudol: lolz

api/routes/apds/activities/post.js
brendansudol: `getApd` is great

PR #315, opened 2018-03-14
client side validation beginnings

brendansudol:
Adds a couple basic validation methods and basic input error styles.

Preview:
validate-ex


lauraponce:
@brendansudol at what point does the "please fill out this field" pop up? Is it when you tab out of it and it's blank or when they try to submit something and it's blank?

web/src/components/FormStateStart.js
```diff
@@ -6,13 +6,26 @@ import { Input } from './Inputs';
...
+      validate={required}
+    />
     <Field name="position" type="text" component={Input} label="Position" />
-    <Field name="email" type="email" component={Input} label="Email address" />
+    <Field
+      name="email"
+      type="email"
+      component={Input}
+      label="Email address"
+      validate={[required, email]}
```
mgwalker: Oh nice!  I like both how the actual validation is abstracted away, and that you can pass in an array of them.  :tada:

web/src/components/Inputs.js
```diff
@@ -13,15 +13,19 @@ const makeInput = InputInner => {
...
-    <div className="mb2">
-      <label htmlFor={name} className={hideLabel ? 'sr-only' : ''}>
-        {label}
-      </label>
-      <InputInner id={name} type={type} {...rest} />
-      {touched && error && <span>{error}</span>}
-    </div>
-  );
+  }) => {
+    const hasError = touched && error;
```
mgwalker: So, ignoring validation errors until the field has been focused and then blurred?  Makes sense, just making sure I understand.  😛

web/src/containers/StateStart.js
```diff
@@ -9,28 +9,19 @@ import PageNavButtons from '../components/PageNavButtons';
...
-// };
-
 class StateStart extends Component {
   componentDidMount() {
     // TODO: don't hardcode user id
     this.props.fetchUserDataIfNeeded(57);
   }
 
-  showResults = data => {
+  updateForm = data => {
```
mgwalker: heheh.  Thanks!

web/src/util/validation.js
```diff
@@ -0,0 +1,12 @@
+export const MESSAGES = {
+  required: 'Please fill out this field',
+  email: 'Please enter a valid email address'
+};
+
+export const required = val =>
+  val === '' || val === null ? MESSAGES.required : null;
+
+export const email = val =>
+  val && !/^[A-Z0-9._%+-]+@[A-Z0-9.-]+\.[A-Z]{2,4}$/i.test(val)
```
mgwalker: Email validation!  😬  What about non-ascii or emoji domains?!  (About 80% kidding, but email validation always makes me squirm because it's so freakin' complicated if you want to capture every possible valid case.)

brendansudol:
@lauraponce good question! right now, it's set to show up after you tab out of the input, though we could change to on submit click if you prefer

web/src/util/validation.js
brendansudol: i agree, we should use a library

web/src/components/Inputs.js
brendansudol: 👍

mgwalker:
Should this be changed to validate-on-submit, or should we go ahead and merge and re-evaluate later?

PR #347, opened 2018-03-22
Factor out some common code

mgwalker:
**NOTE**: I'd recommend looking at `/middleware/apd.js` and then one of the route handlers and its unit tests just to get a feel for whether this is a worthwhile change before spending a bunch of time reviewing it all.

This PR moves the existing authorization middleware into a `./api/middleware` directory, then adds four new pieces of middleware:

1. routes can optionally have the relevant APD data object attached to the request object at `req.meta.apd`
   - By default, this middleware expects the request to have a URL parameter called `id` and assumes that is the APD ID.  It will query against the `apd` data model
       - ` loadApd()`
   - To search against a _different_ data model, pass a first argument into the middleware.  The new target model must have a singular relationship to an APD called `apd` (i.e., in the Bookshelf model, it should have a `belongsTo('apd')` or `hasOne('apd')` relationship created).  In this case, the request parameter `id` is assumed to be the unique ID of the target model.
      - `loadApd(apdActivity)` - will search for an activity data object whose ID is equal to `req.params.id`, and then fetch its related APD
    - To use a different request parameter as the ID of the target model, pass a second argument into the middleware with the name of the request parameter to use.
      - `loadApd(apdActivity, 'activityID')` - will search for an activity data object whose ID is equal to `req.params.activityID`, and then fetch its related APD
   - `req.meta.apd` is not set if there's no relevant APD or if there are any database errors
      - let's discuss whether we think this is the correct behavior, or should it go ahead and return a 404/500?

2. routes can optionally require that the relevant APD data object exist, or the middleware will return a 404 and stop processing
   - Has the same configuration options as `loadApd()`
   - Calls `loadApd()` before doing any checks, so it's not necessary to use both middlewares
   - `requireApd()`

3. routes can optionally require that the current user has access to the relevant APD data object, or the middleware will return a 404 and stop processing
   - Has the same configuration options as `loadApd()`
   - Calls `requireApd()` before doing any checks, so it's not necessary to use both middlewares
   - Assumes that the user is already logged in, so probably ought to have a `loggedIn` or `can` middleware before this one
   - `userCanEditApd()`

4. routes can optionally have the relevant APD activity data object attached to the request object at `req.meta.activity`
    - Uses the `apdActivity` model, and the `id` request parameter
    - To use a different request parameter as the activity ID, pass a first argument into the middleware with the name of the request parameter
      - `loadActivity('activityID')`
    - returns a 404 and stops processing if the activity is not found
    - returns a 500 and stops processing if there are errors with the database
      - ☝️ note how this is different from the `loadApd` middleware; this is why we should discuss, and honestly I'm not sure why I ended up doing them two different ways...  😬

This PR ends up looking like a wash, but a bunch of the "lines added" are places where prettier expanded one line into multiple, I promise.  Before I added new unit tests, git was saying 1,000 lines deleted, 480 lines added.  I felt like that was pretty good.


api/auth/serialization.js
```diff
@@ -25,7 +25,8 @@ module.exports.deserializeUser = async (
         id: user.get('id'),
         role: user.get('auth_role'),
         state: user.get('state_id'),
-        activities: await user.activities()
+        activities: await user.activities(),
+        model: user
```
mgwalker: Attach the user DB model to the request, as we'd discussed previously!

api/main.js
```diff
@@ -14,6 +14,7 @@ const server = express();
 endpointCoverage.registerCoverageMiddleware(server);
 server.use((req, res, next) => {
   req.id = uuid();
+  req.meta = {};
```
mgwalker: Since the new middlewares add stuff to `req.meta`, I wanted to make sure it was defined first.  That way each downstream middleware doesn't have to check for it and create it as necessary.

api/middleware/apd.js
```diff
@@ -0,0 +1,108 @@
+const logger = require('../logger')('apd middleware');
+const {
+  apd: ApdModel,
+  apdActivity: defaultActivityModel
+} = require('../db').models;
+
+const cache = {};
+
+module.exports.loadApd = (model = ApdModel, idParam = 'id') => {
+  if (!cache[['loadApd', model, idParam]]) {
```
mgwalker: This cacheing isn't super necessary, but it helps make our tests more robust.  It seems like a pretty low cost, but I'm open to discussion about it.

api/middleware/auth.js
```diff
@@ -2,7 +2,7 @@ const logger = require('../logger')('auth middleware');
 
 const canCache = {};
 
-module.exports.loggedIn = (req, res, next) => {
+const loggedIn = (req, res, next) => {
```
mgwalker: Assigning these arrow functions to actual variables gives them names in the stack trace, which turns out to be terribly useful.

api/routes/apds/activities/expenses/put.js
```diff
@@ -4,103 +4,92 @@ const {
...
-    try {
-      const activityID = +req.params.id;
-      const activity = await ActivityModel.where({ id: activityID }).fetch({
-        withRelated: ['apd', 'expenses.entries']
-      });
+  app.put(
+    '/activities/:id/expenses',
+    loggedIn,
+    loadActivity(),
+    userCanEditAPD(ActivityModel),
```
mgwalker: Uses the activity model to find the activity with the ID given by `req.params.id`, then uses that object's relationships to get the APD.

api/routes/apds/activities/expenses/put.js
```diff
@@ -4,103 +4,92 @@ const {
...
-      }
-
-      const apdID = activity.related('apd').get('id');
-      if (!await userCanEditAPD(req.user.id, apdID)) {
-        logger.verbose(
-          req,
-          'user (state) not associated with the apd this activity belongs to'
-        );
-        return res.status(404).end();
-      }
```
mgwalker: That new middleware means these two checks don't need to be anywhere else anymore.  I think we had them in 3-4 places.

api/routes/apds/activities/goals/put.test.js
```diff
@@ -74,77 +78,39 @@ tap.test('apd activity goal PUT endpoint', async endpointTest => {
...
-        activityObj.get.withArgs('id').returns('apd-id');
-        userCanEditAPD.resolves(false);
-
-        await handler(req, res);
-
-        notFoundTest.ok(res.status.calledWith(404), 'HTTP status set to 404');
-        notFoundTest.ok(res.send.notCalled, 'no body is sent');
-        notFoundTest.ok(res.end.calledOnce, 'response is terminated');
-      }
-    );
```
mgwalker: Likewise with the tests, these checks can just go away.

api/middleware/apd.js
```diff
@@ -0,0 +1,108 @@
...
+          }).fetch({
+            withRelated: ['activities']
+          });
+          next();
+        } else {
+          const obj = await model
+            .where({ id: req.params[idParam] })
+            .fetch({ withRelated: ['apd.activities'] });
+          if (obj) {
+            req.meta.apd = obj.related('apd');
```
brendansudol: feels like we should add a log statement here if there's no `obj` instance (or throw an exception) vs. just the silent `next()` with a null `req.meta.apd`

api/middleware/apd.js
```diff
@@ -0,0 +1,108 @@
...
+          }
+          next();
+        }
+      } catch (e) {
+        logger.error(req, e);
+        next();
+      }
+    };
+    cache[['loadApd', model, idParam]] = loadApd;
```
brendansudol: consider pulling this (`['loadApd', model, idParam]`) out into a variable (`lookup` or `key`?) since we use a few times

api/middleware/apd.test.js
```diff
@@ -0,0 +1,206 @@
...
+
+  done();
+});
+
+tap.test('APD-related middleware', async middlewareTests => {
+  middlewareTests.test('load apd', async loadApdTests => {
+    loadApdTests.skip('from an APD model', async () => {
+      // There's an in-built dependency on the real APD model,
+      // so we need to figure out how to either mock that in
+      // such a way that we fool the code, OR we need to find
```
brendansudol: would it be easier if we checked and mocked against an instance method on the APD model, i.e., `if model.foo == ApdModel.foo` ?

api/middleware/auth.js
brendansudol: yes, i'm a big fan of this small change

api/middleware/auth.js
```diff
@@ -13,11 +13,13 @@ module.exports.loggedIn = (req, res, next) => {
   }
 };
 
+module.exports.loggedIn = loggedIn;
```
brendansudol: how do you feel about this general pattern: if we assign an exported function to a variable name in a file, then we assign all exported functions to variables (in that file) and then the very last line is the collection of all exports:  `module.exports = { loggedIn, can }`

api/routes/apds/activities/expenses/put.js
brendansudol: sweeeet

brendansudol:
I think returning a 404 / 500 (vs. just null values) for these middleware functions makes sense (like you do in `loadActivity`). I can't think of a case where the latter would be preferred for us -- what do you think?

api/middleware/apd.js
mgwalker: Hmm...  Definitely should at least be a log.  Not sure about the throw - that'd cause a 500 error with the body of the exception.  The reason for this `loadApd` middleware being separate is, honestly, premature optimization.  My original plan was to send a 404 if the APD didn't exist (like what `requireApd` does), but then I thought, what if I want to load the APD if it exists, but it's okay if it doesn't?

So it's reasonable to me if you think we should collapse `loadApd` and `requireApd` into the same middleware.  Right now at least, `loadApd` is ***only*** used by `requireApd`, so it shouldn't have any outward-facing implications.  We could always factor it back out later if we needed to, but I'm not sure why we would...  😛

api/middleware/apd.js
mgwalker: 👍

api/middleware/auth.js
mgwalker: I am on board with that.

api/middleware/apd.test.js
mgwalker: I poked around on the bookshelf object to try to find something that would uniquely identify the model (e.g., wouldn't it be cool if `model.name` or `model.tableName` existed?), but I couldn't find anything that would be unique to the APD model.

But we could inject the "control" as well, I suppose:

```js
module.exports.loadApd = (model = ApdModel, idParam = 'id', ApdModel = ApdModel) => {
...
}
```

It seems hacky but a lot of the model injection we're doing elsewhere is ultimately just to enable testing so maybe not?  🤷‍♂️

brendansudol: yeah, `model.tableName` would have been perfect, alas... hmm yeah it does seem a little hacky but i'm fine with adding that param for now

api/middleware/apd.js
mgwalker: Ditched `requireApd`, made `loadApd` return the 404/500 instead.  👍/👎?

brendansudol: 👍

PR #361, opened 2018-03-29
add cost allocation table

brendansudol:
Adds cost allocation model

### DIFF
--- a/api/migrations/20180329105503_add-activity-cost-allocate.js
+++ b/api/migrations/20180329105503_add-activity-cost-allocate.js
@@ -1,0 +1,14 @@
+exports.up = async knex => {
+  await knex.schema.createTable('activity_cost_allocations', table => {
+    table.increments('id');
+    table.string('name', 128);
+    table.integer('share');
+
+    table.integer('activity_id');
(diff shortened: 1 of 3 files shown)


api/migrations/20180329105503_add-activity-cost-allocate.js
```diff
@@ -0,0 +1,14 @@
+exports.up = async knex => {
+  await knex.schema.createTable('activity_cost_allocations', table => {
+    table.increments('id');
+    table.string('name', 128);
+    table.integer('share');
```
mgwalker: Do we need a flag to indicate whether this comes "off the top," or if it's applied to the state's share?  E.g., if the state gets a grant for $100 on a $1000 project...

* if the grant is applied "off the top," then the CMS share is $810 and the state share is $90
* if the grant is applied to the state share, then the CMS share is $900 and the state share is $100, completely covered by the grant

I'm okay with not solving that right this second, just flagging that I think this might be important per our conversation with @NAretakis and @jeromeleecms earlier this week.

(Maybe a better way of thinking about it than "off the top" is percent-of-total vs. percent-of-state-share?)

Issue #367, opened 2018-03-29
It's possible to create multiple activities with the same name when submitting multiple activities names at once


mgwalker:
On the POST `/apds/:id/activities` endpoint, you can send either a single activity object or an array of activity objects.  If you send a single object, the endpoint enforces uniqueness on activity names.  If you send an array of objects, however, it's the wild west.

mgwalker:
Closed by #382

PR #379, opened 2018-03-30
Fix middleware cacheing

mgwalker:
Fixes middleware cacheing by converting Bookshelf.js model objects into numbers.  The cache is being keyed on an array of the form `[middlewareName, model, other, stuff]` - but it is stored as a plain object with that key as the object property.  However, object properties must be strings, so internally JS is stringifying the key.  No big deal, except ***all Bookshelf.js model objects stringify to null***.

Thus, a cache keyed on `['userCanEditAPD`, APDModel]` is exactly the same as one keyed on `['userCanEditAPD', ActivityModel]`.  Blam, everything breaks - we check whether users can edit APDs based on whichever model happened to be passed in the first time the middleware was cached, and thereafter every access to the cache is a hit and returns the first instance.  (Boo.)

To get around that, I create an array of models as the middlewares are created.  If the model exists in the array, I return its index.  If it doesn't, I add it to the array and then return its index.  Numbers stringify really well, so that fixes the cache.

While I was fiddling with it, I pulled cacheing out into its own little module and modified the `can` middleware to use it.  No sense doing cacheing independently in several places.

## Merging

This one still has a broken endpoint test, related to #367.  Since we're not merging into master, though, it should be fine to go ahead and merge this one.  A fix for #367 is incoming in a separate PR.

### DIFF
--- a/api/middleware/cache.js
+++ b/api/middleware/cache.js
@@ -1,0 +1,18 @@
+const middlewareCache = {};
+const cache = (key, getMiddleware) => {
+  if (!middlewareCache[key]) {
+    middlewareCache[key] = getMiddleware();
+  }
+  return middlewareCache[key];
+};
+
+const knownModels = [];
+const modelIndex = model => {
+  if (knownModels.includes(model)) {
+    return knownModels.indexOf(model);
+  }
+  knownModels.push(model);
+  return knownModels.length - 1;
+};
+
+module.exports = { cache, modelIndex };
--- a/api/middleware/apd.js
+++ b/api/middleware/apd.js
@@ -1,18 +1,17 @@
 const logger = require('../logger')('apd middleware');
+const { cache, modelIndex } = require('./cache');
+
 const {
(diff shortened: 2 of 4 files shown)


api/middleware/cache.js
```diff
@@ -0,0 +1,18 @@
+const middlewareCache = {};
+const cache = (key, getMiddleware) => {
+  if (!middlewareCache[key]) {
+    middlewareCache[key] = getMiddleware();
+  }
+  return middlewareCache[key];
+};
+
+const knownModels = [];
+const modelIndex = model => {
```
brendansudol: take or leave this suggestion, but i would tweak this to have only one return statement:

```
const modelIndex = model => {
  if (!knownModels.includes(model)) knownModels.push(model);
  return knownModels.indexOf(model);
};
```

mgwalker: I'm 100% in favor of this recommendation.

PR #387, opened 2018-04-03
Return consistent activity data across endpoints

mgwalker:
Across our different endpoints, we were setting the `withRelated` property individually in each one when we queried the database for activities.  The `withRelated` property controls which child sections get included when the APD is returned, such as expenses, approaches, etc.  Over the course of development, they became inconsistent as we added sections - that is, some of the older code had not been updated.

Rather than update the code everywhere to use the same `withRelated`, I figured there might be a better approach.  I modified the model objects to allow them to define static properties (that is, properties that are available on the model itself, rather than on instances of the model), and then defined a `withRelated` property there.  Now, the endpoints just reference that, and if we need to change what related sections get returned, we can do it in one place and have it updated immediately everywhere.

While I was adding static stuff to the models, I also added a `modelName` property, which let me simplify the APD middleware behavior a little bit.  It also let me fix a bug that would almost definitely never show up in practice, but it could have caused issues in unit testing.  (That's a weird kind of bug.)


api/db/activity.js
```diff
@@ -24,6 +24,17 @@ module.exports = () => ({
...
+    static: {
+      withRelated: [
+        'approaches',
+        'goals',
+        'goals.objectives',
+        'expenses',
+        'expenses.entries',
+        'schedule'
+      ]
+    },
```
mgwalker: If you put a `static` property on a model object, all of its properties will be attached to the model constructor.  They all get `modelName` for free (see below), but I figured this way we could easily add more stuff later if we needed/wanted.

api/db/index.js
```diff
@@ -28,7 +28,14 @@ const setup = (
...
       logger.silly(`loading model [${modelName}]`);
-      const model = orm.Model.extend(modelObjects[modelName]);
+
+      const modelObj = modelObjects[modelName];
+      let staticProps = { modelName };
+      if (modelObj.static) {
+        staticProps = { modelName, ...modelObj.static };
+      }
+
+      const model = orm.Model.extend(modelObjects[modelName], staticProps);
```
mgwalker: The second argument is for class properties.  Oddly, it will ignore a `name` property. 🤷‍♂️

api/endpoint-tests/activities/goals/put.js
```diff
@@ -98,19 +98,23 @@ tap.test(
               200,
               'gives a 200 status code'
             );
+
+            body.goals.forEach(goal => goal.objectives.sort());
```
mgwalker: It's really obnoxious having to sort all these arrays to make the tests pass, but maybe these are actual bugs - see the question in Slack about whether these should be returned in a consistent order!

api/middleware/apd.js
```diff
@@ -6,25 +6,27 @@ const {
...
 
-module.exports.loadApd = (
-  model = defaultApdModel,
-  idParam = 'id',
-  ApdModel = defaultApdModel
-) =>
-  cache(['loadApd', modelIndex(model), idParam, ApdModel], () => {
+const apdRelations = (defaultApdModel ? defaultApdModel.withRelated : []).map(
+  relation => `apd.${relation}`
+);
```
mgwalker: I'm open to suggestions on better ways of doing this.  The gist of it is when we're loading an APD from a model *other than* the APD model, we want to fetch all of the APD's relationships as well so that we end up with a fully-formed APD object.

brendansudol: when would this ternary evaluate to `false`?

api/db/activity.js
brendansudol: 👌

api/middleware/apd.js
mgwalker: When the `db.setup()` method is not called first, so the `models` property is an empty object resulting in `const { apd: defaultApdModel } = require('../db').models;` being undefined.

It happens when running the unit tests.  It _shouldn't_ happen otherwise, or else the whole system will crap out, but 🤷‍♂️.  We don't want to run `db.setup()` in the unit tests because then it wants to connect to a real database.  That introduces the risk of forgetting to mock something and then changing the database instead of the mocks.  Shouldn't be a big risk since unit tests are hopefully never running against the production database, but a risk nonetheless.

I'm 100% open to alternatives, I'm not a huge fan of having conditional code that exists solely because of unit tests...

PR #389, opened 2018-04-03
more PoC page refinements

brendansudol:
* help box component
* more activity sections
* wsiwig demo (for activity description)

previews:
screen shot 2018-04-03 at 1 10 55 pm

screen shot 2018-04-03 at 1 01 10 pm


web/package.json
```diff
@@ -49,13 +49,15 @@
     "d3-ease": "^1.0.3",
     "d3-selection": "^1.3.0",
     "d3-transition": "^1.1.1",
+    "draft-js": "^0.10.5",
```
mgwalker: One small concern here is that by using this library, we're basically committing ourselves to it indefinitely since nothing else will be able to (directly) render its style-annotated text format.  A couple of things we could do are also store the plaintext alongside the annotated text, or do a conversion to markdown and store that alongside the annotated text.  Either way, I'd just want to make sure we have a way of getting the text back out of the system in a way that things other than our current frontend can use.

NAretakis: This is interesting. So, if we store the plan text, that gives us flexibility in the future to have other front end stuff (or other things) access/use that text, is that right?

mgwalker: Yep, but you'd lose formatting.  And if we store it as markdown, then the formatting would be preserved and there are any number of tools that can show it formatted.

The good news is that if we only store the proprietary formatted text now, we can pretty easily convert it to plaintext and/or markdown in the future.  My only concern would be to do it sooner rather than later (sometime in the next 6-12 months, for example) just in case the proprietary thing goes kaput, which seems unlikely.

PR #408, opened 2018-04-10
add more sections

brendansudol:
* state personnel, expenses, standards & conditions, etc.

Preview:
preview

### DIFF
--- a/web/src/components/FormActivityApproach.js
+++ b/web/src/components/FormActivityApproach.js
@@ -11,14 +11,16 @@ const Approaches = ({ fields, meta: { error, submitFailed } }) => (
     {fields.map((approach, idx) => (
       <div key={approach} className="mb3 sm-col-8">
         <div className="relative">
-          <button
-            type="button"
-            className="absolute right-0"
-            title="Remove Goal"
-            onClick={() => fields.remove(idx)}
-          >
-            Remove
-          </button>
+          {idx !== 0 && (
+            <button
+              type="button"
+              className="absolute right-0"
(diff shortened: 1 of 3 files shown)


web/src/components/FormActivityApproach.js
```diff
@@ -11,14 +11,16 @@ const Approaches = ({ fields, meta: { error, submitFailed } }) => (
...
         <div className="relative">
-          <button
-            type="button"
-            className="absolute right-0"
-            title="Remove Goal"
-            onClick={() => fields.remove(idx)}
-          >
-            Remove
-          </button>
+          {idx !== 0 && (
```
mgwalker: Wondering if this should be something like `fields.length !== 0`?

brendansudol: the intention here is that you should not be allowed to delete the approach input if it's the only one on the page, but if you add multiple approaches, you can delete those.

brendansudol: if we use `fields.length`, then i think it will either allows have a delete button or never have one (for each input), but i think we want something slightly different (all but the first having the delete button)

mgwalker: What if the first one is the only one they want to delete?  That was the situation I was thinking of.  Like, I have five fields, and the 2nd-5th ones are right, but I want to get rid of that first one.

brendansudol: ahh, good point, will tweak 👍 🙌

PR #410, opened 2018-04-11
New base data model + update-everything endpoint + associated updates

mgwalker:
This PR modifies the `/apd/{id}/` PUT endpoint to accept a complete APD object, and updates the APD objects along with all of its related data.  To do that, I created a new base data model that all of our existing models derive from.  The base model adds some helper functionality to our models:

### On model instances
* overrides the default `save()` method to call `validate()` first
* overrides the default `destroy()` method to also destroy related data (basically implements a cascading delete; we may want to think on this one - should `cascade` be a flag so we can let the database prevent us from accidental deletes?)
* adds a `synchronize()` method
  * this takes a big ol' JSON blob and uses it to update the model instance as well as its children

### On model classes
* adds a `synchronize()` method
  * this takes an array (or a single object that it then turns into an array), an object indicating how to limit the method's impact (e.g., for all goals related to a specific activity, you might pass `{activity_id: 7}`), and an optional transaction object that database actions should be executed against

### Changes to model objects
To support dynamically walking the model objects and their relationships, model definitions get a few more pieces of data as well.  These properties are added to the `static` property:
* `updateableFields` is an array of the field names that should be taken from the JSON object and inserted into the body; this is basically used to filter so we don't try to insert columns that user's shouldn't be able to edit or that don't exist
* `owns` is an object whose keys are JSON fields representing models that are "owned" by this one, and whose values are the names of those models
* `foreignKey` is a string, and it is the name of the field in the child tables that refers back to this model


api/db/activity/activity.js
```diff
@@ -0,0 +1,77 @@
...
+      foreignKey: 'activity_id',
+      withRelated: [
+        'approaches',
+        'goals',
+        'goals.objectives',
+        'expenses',
+        'expenses.entries',
+        'schedule'
+      ]
+    },
```
mgwalker: I want to revisit this pattern in the future because there's "secret knowledge" going on.  That is, it's extending bookshelf in a way that's not documented and could confuse a dev new to the project.

I definitely prefer having this little bit of metadata attached to the model over having gobs of very similar code on every model, though.

api/db/apd.js
```diff
@@ -11,8 +11,11 @@ module.exports = () => ({
     },
 
     static: {
+      updateableFields: ['status', 'period'],
+      foreignKey: 'apd_id',
+      owns: { activities: 'apdActivity' },
       withRelated: [
-        'activities',
+        { activities: query => query.orderBy('id') },
```
mgwalker: This is how we can specify how to sort these related models.  I only did it on activities here, but we might want to do it on all of them later.  More on that in a minute...

api/db/authorization.js
```diff
@@ -34,19 +34,19 @@ module.exports = () => ({
       return this.related('activities').pluck('name');
     },
 
-    async validate(activities, model = this, activityModel) {
+    async validate({ transacting, activities = [] } = {}) {
```
mgwalker: The base model passes a single argument to `validate` with a possible `transacting` property.  If the `validate` method needs additional arguments, we can make them properties of the "one" object.

In most cases, extra arguments are unnecessary because you validate the model's "new" attributes, but this `roles` model is quirky because there's no way to get the new activities from the role model itself until the role is saved, but we don't want to save it until it's validated.  🤷‍♂️

api/db/base.js
```diff
@@ -0,0 +1,191 @@
...
+    super(message);
+
+    this.error = { error: message };
+  }
+
+  // eslint-disable-next-line class-methods-use-this
+  get statusCode() {
+    return 400;
+  }
+}
```
mgwalker: This might be a premature optimization, but I needed a way to signal out that this was a validation error rather than some other exception.  I tried a couple of things, then I thought hey, what if we had a set of custom errors and they could set a statusCode?  I'm not sold on this approach, though...

brendansudol: 👍

api/db/base.js
```diff
@@ -0,0 +1,253 @@
...
+      if (data[fieldName]) {
+        logger.silly(
+          `${modelName} | synchornize ${fieldName} with ${
+            childModels[fieldName]
+          } model`
+        );
+        const Model = allModels[childModels[fieldName]];
+        return Model.synchronize(data[fieldName], foreignKey, transacting);
+      }
+      return true;
```
brendansudol: for readability (and to allow the main chunk of code to be less indented), how do u feel about reversing this conditional return? i.e.,

```
if (!data[fieldName]) return true

const Model = allModels[childModels[fieldName]];
return Model.synchronize(data[fieldName], foreignKey, transacting);
...
```

api/db/base.js
```diff
@@ -0,0 +1,254 @@
...
+          transacting,
+          this.modelName
+        );
+      }
+      return false;
+    });
+
+    let inserting = Promise.resolve();
+    dataArray.filter(d => !d.id).map(async raw => {
+      // Chain these promises so they execute in the same
```
brendansudol: neat!

api/routes/apds/put.js
```diff
@@ -10,18 +8,23 @@ module.exports = app => {
     logger.silly(req, `attempting to update apd [${req.params.id}]`);
 
     try {
-      logger.silly(req, 'updating apd fields');
-      logger.silly(req, 'filter body data to editable fields');
-      const newData = pick(req.body, ['status', 'period']);
-      const apd = req.meta.apd;
-      apd.set(newData);
+      await req.meta.apd.synchronize(req.body);
```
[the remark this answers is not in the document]
brendansudol: 😍

api/routes/apds/put.js
```diff
@@ -10,18 +8,23 @@ module.exports = app => {
...
       logger.silly(req, 'all done');
-      return res.send(apd.toJSON());
+      const apd = await req.meta.apd.fetch({
+        withRelated: req.meta.apd.static.withRelated
+      });
+      res.send(apd.toJSON());
     } catch (e) {
-      logger.error(req, e);
-      return res.status(500).end();
+      if (e.statusCode) {
```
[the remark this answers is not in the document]
brendansudol: what's your reservation - that someone would know that there is now a `statusCode` property on the error?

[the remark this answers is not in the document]
brendansudol: ahh, gotcha

PR #411, opened 2018-04-12
tooltips, year column action buttons, etc

brendansudol:
Adds tooltip component, and creates a PoC of a "year actions" button (right now, the only action is greying out the columns if that year is not applicable for an activity, but we will have additional actions later, like copying values from last year over to this year). 


brendansudol:
FYI, this is the final POC editing PR before I dramatically clean up this one crazy big file (now that components and structure are more solidified)

PR #415, opened 2018-04-17
Update activities, approaches, expenses, goals, and schedule endpoints to use data model "synchronize" method

mgwalker:
Rather than each of these 5 endpoints having *very similar* code to do essentially the same thing, now that we have a generic data model method for synchronizing the data to a JSON object, let's use that instead.  Furthermore, let's pull that code into reusable middleware and only customize the pieces we actually need to.

Eliminates a lot of functional code and a lot of tests as well!


api/db/base.js
```diff
@@ -149,101 +158,114 @@ const classExtension = (orm, models) => ({
...
+      await Promise.all(saving);
+      await inserting;
+      await Promise.all(synchronizing);
+    };
+
+    if (transaction) {
+      await exec(transaction);
+    } else {
+      await orm.transaction(transacting => exec(transacting));
+    }
```
mgwalker: If you call `synchronize` on the model class outside of a transaction, this code will wrap it up in a transaction.  A whole lot of green lines just to represent indentation changes.  😛

api/endpoint-tests/activities/approaches/put.js
```diff
@@ -56,11 +56,22 @@ tap.test(
         );
 
         authenticated.test(
-          'with approaches that are not an array',
+          'with approaches that are invalid',
           async invalidTest => {
             const { response, body } = await request.put(url(4100), {
               jar: cookies,
-              json: { hello: 'world' }
```
mgwalker: Because of the way the `synchronize` method works, a non-array will actually work - `synchronize` will wrap it in an array if necessary.

api/endpoint-tests/apds/activities/post.js
```diff
@@ -129,6 +129,7 @@ tap.test(
                     name: 'My Second Activity'
                   },
                   {
+                    id: Number,
```
mgwalker: It turns out that tap's object pattern matcher understands the `Number` constructor.  :tada:

api/middleware/synchronize.js
```diff
@@ -0,0 +1,66 @@
+const logger = require('../logger')('sync middleware');
+const { cache, modelIndex } = require('./cache');
+
+const synchronizeAll = getSpecifics =>
+  cache(
+    ['sync-all', modelIndex(getSpecifics)],
```
mgwalker: `modelIndex` might not be the best name for this function anymore - it just turns a non-primitive into a number for cacheing purposes.  🤷‍♀️

api/middleware/synchronize.js
```diff
@@ -0,0 +1,66 @@
...
+    () => async (req, res, next) => {
+      let action;
+
+      try {
+        const { modelClass, foreignKey, action: actionToReport } = getSpecifics(
```
mgwalker: Naming things is hard.  `getSpecifics` seems dumb.  Maybe `getDetails`?  I dunno...

api/middleware/synchronize.js
```diff
@@ -0,0 +1,66 @@
...
+          logger.error(req, e);
+          res.status(500).end();
+        }
+      }
+    }
+  );
+
+const synchronizeSpecific = getModelToSync =>
+  cache(
+    ['sync-one', modelIndex(getModelToSync)],
```
mgwalker: `getModelToSync` should probably be something else too...  maybe also `getDetails`?

api/package.json
```diff
@@ -13,7 +13,8 @@
     "test":
       "NODE_ENV=test tap -J --cov --coverage-report=lcov --no-browser --reporter=spec --test-arg=--silent '{,!(node_modules)/**/}*.test.js'",
     "test-endpoints": "tap --reporter=spec 'endpoint-tests/*/**/*.js'",
-    "test-specific": "NODE_ENV=test tap --reporter=spec",
+    "test-specific":
+      "NODE_ENV=test tap --cov --coverage-report=lcov --no-browser --reporter=spec",
```
mgwalker: Collect coverage even on specific tests.  This is useful (to me, anyway) when I'm writing tests, so I can see which lines I've covered and make sure that there are good assertions covering what the code is actually doing.

api/routes/apds/activities/goals/openAPI.js
```diff
@@ -28,9 +28,14 @@ const openAPI = {
...
-                description:
-                  'The description of the objective associated with this goal'
+                type: 'object',
+                properties: {
+                  description: {
+                    type: 'string',
+                    description:
+                      'The description of the objective associated with this goal'
+                  }
+                }
```
mgwalker: `synchronize` is generic so it doesn't understand how to turn a single string into a model.  So, don't accept single strings anymore and instead take an object that fully represents the model.  Nobody is using this part of the API so there's nothing to break.  😛

api/routes/apds/activities/put.js
```diff
@@ -1,58 +1,36 @@
...
+      `/activities/:id${putter ? `/${putter}` : ''}`,
+      loggedIn,
+      loadActivity(),
+      userCanEditAPD(ActivityModel),
+      synchronizeSpecific(syncResponder)
+    );
+  });
 };
+
+module.exports.syncResponders = {};
```
mgwalker: This module now creates the `put` endpoints for `/activities/:id`, `/activities/:id/approaches`, `/activities/:id/expenses`, `/activities/:id/goals`, and `/activities/:id/schedule`, since they are almost exactly identical.  Just a little bit of conditional logic in this loop and you've got it.

mgwalker:
@brendansudol Definitely recommend reviewing this one in split view, if you don't normally.  So much deletion that the inline view is really muddy.

api/routes/apds/activities/goals/openAPI.js
brendansudol: 👍

api/routes/apds/activities/post.js
```diff
@@ -1,92 +1,25 @@
...
-          apd_id: apd.get('id')
-        }).fetchAll();
-
-        return res.send(allActivities.toJSON());
-      } catch (e) {
-        logger.error(req, e);
-        return res.status(500).end();
-      }
-    }
+    synchronizeAll(syncResponder)
```
brendansudol: so nice

PR #425, opened 2018-04-18
Squash DB migrations

### LINKED ISSUE #421: Squash database migrations
There's no reason to keep every incremental database migration forever - we only need to keep them between "releases."  So instead of having the model built up over several migrations, let's squash them into one.

mgwalker:
Since we haven't released to production yet, we can safely squash the database migrations down to a single file.  We should do this periodically.  Just before going to production, we should squash once more.  There should ideally just be one migration per release.

Closes #421.


api/endpoint-tests/users/get.js
```diff
@@ -69,7 +69,7 @@ tap.test('users endpoint | GET /users/:userID', async getUserTest => {
             name: null,
             position: null,
             phone: null,
-            state: null
```
mgwalker: For some reason, the User model had both `state` and `state_id` fields.  While I was in here, I got rid of the `state` field since it wasn't being used in the relationships anywhere, and I updated the code throughout to use the `state_id` field instead.

mgwalker:
I think there's a bug in here related to the users table and the `state` vs. `state_id` fields.  Going to update the tests to catch it, then fix it, then push.

mgwalker:
Never mind, apparently I fixed that yesterday in 442b25e.  😬

api/routes/users/get.js
```diff
@@ -5,11 +5,11 @@ const can = require('../../middleware').can;
 const allUsersHandler = async (req, res, UserModel) => {
   logger.silly(req, 'handling GET /users route');
   try {
-    const users = await UserModel.fetchAll({ columns: ['id', 'email'] });
-    logger.silly(req, 'sending users', JSON.parse(JSON.stringify(users)));
-    res.send(
-      users.map(user => ({ email: user.get('email'), id: user.get('id') }))
```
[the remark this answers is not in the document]
brendansudol: good catch

api/routes/users/put.js
```diff
@@ -2,10 +2,6 @@ const logger = require('../../logger')('users route post');
 const defaultUserModel = require('../../db').models.user;
 const can = require('../../middleware').can;
 
-// TODO: figure out better/cleaner solution.
```
brendansudol: TODO is DONE!

PR #426, opened 2018-04-19
start reduxing activity state

brendansudol:
Overview: Ok, this is the start of wiring up our new post-pivot activity stuff to redux. I'm leaving the former `poc` page as is (it's now at the root page) so we have it as a reference as we're doing this (and people can still look and click on stuff). and the new stuff is at `/activities`. Eventually, this will look very similar / better than what's at the `poc` page, AND it will sync with our client data store, which we can then easily push to the backend.

More details:

In terms of the structure of the `activities` state, I'm following this normalized structure (https://redux.js.org/recipes/structuring-reducers/normalizing-state-shape) which will help with performance.

I don't yet love the structure / API for the UPDATE_ACTIVITY action right now, but it's a start and it works.

For the rich editor fields, we convert the `EditorState` object to html when syncing with our state (and from html to that `EditorState` on initialization); we can abstract/consolidate these conversions as we build out additional components that need this.

When reviewing, please go to `/activities` page and play around with the fields and inspect the redux logger console statements, too.

The next steps are adding more components within the `ActivityDetailAll` (similar to what I've done with `ActivityDetailDescription`).

I also improved input focus states and made them consistent across our various input types (input, textarea, editor)


web/src/containers/ActivityDetailAll.js
```diff
@@ -0,0 +1,35 @@
...
+  </Collapsible>
+);
+
+ActivityDetailAll.propTypes = {
+  aId: PropTypes.number.isRequired,
+  title: PropTypes.string.isRequired
+};
+
+const mapStateToProps = ({ activities: { byId } }, { aId, num }) => {
+  const activity = byId[aId];
```
mgwalker: I'm curious why you get the activity from redux store by ID rather than taking it as a param?

brendansudol: yeah, i wasn't really sold on this either; i read that it was better performance to pass down the id and look up the data object via connect, let me track down the article

brendansudol: https://github.com/reactjs/redux/issues/1751

web/src/containers/ActivityDetailDescription.js
```diff
@@ -0,0 +1,122 @@
...
+        <div className="mb3">
+          <div className="mb1 bold">
+            Please describe the activity in detail
+            <Icon icon={faHelp} className="ml-tiny teal" size="sm" />
+          </div>
+          <Editor
+            toolbar={EDITOR_CONFIG}
+            editorState={descLong}
+            onEditorStateChange={this.onEditorChange('descLong')}
+            onBlur={this.syncEditorState('descLong')}
```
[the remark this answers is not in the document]
brendansudol: exactly!

web/src/containers/ActivityDetailAll.js
mgwalker: Fascinating.  Really gets into the weeds of how React handles re-rendering as state changes.

web/src/reducers/activities.js
```diff
@@ -0,0 +1,49 @@
...
+
+const initialState = {
+  byId: { 1: newActivity(1) },
+  allIds: [1]
+};
+
+const reducer = (state = initialState, action) => {
+  switch (action.type) {
+    case ADD_ACTIVITY: {
+      const id = Math.max(...state.allIds, 0) + 1;
```
mgwalker: Writing this out for my own benefit...  Creating an ID here is safe because when we sync to the API, we can refresh the redux state to match the "real" ID and that will cascade through the app and everything will be okay.  🤔

I'm not sure that's true, but let's fix it when it happens rather than spend any time theorizing about what might happen in the future since it can't possibly happen right now.

web/src/reducers/activities.js
```diff
@@ -0,0 +1,49 @@
...
+const newActivity = id => ({
+  id,
+  name: '',
+  types: ['HIT'],
+  descShort: '',
+  descLong: '',
+  altApproach: ''
+});
+
+const updateEntry = (state, action) => {
```
mgwalker: Do you think using [updeep](https://www.npmjs.com/package/updeep) here might be good?  The code would look something like...

```
return updeep({ byId: { [id]: { [name]: value }}}, state);
```

web/src/reducers/activities.js
brendansudol: yep, that was my thought process as well

web/src/reducers/activities.js
brendansudol: whoa - never heard of it, but i'm intrigued!!

brendansudol: updeep is awesome, gonna switch to this going forward -- great tip!

PR #430, opened 2018-04-19
Sync system roles to documentation

mgwalker:
- updates the documentation to make a distinction between a federal user's `submit-response` and a state user's `submit-response` activities
- makes the list of activities and roles in the system match the documentation
  - updates shared role/activities database seeds so initial database will have the right data
  - updates endpoint tests for roles and activities accordingly
- updates API routes that relied deleted activities:
  - `GET /auth/activities` now wants the `view-roles` activity
  - `DELETE /auth/activities/:id` now wants the `edit-roles` activity
  - `GET /states` and `GET /states/:id` no longer want an activity; instead, any logged in user can see state info (state name, Medicaid director, high-level stuff like that)
  - `PUT /states/:id` is removed because there is no activity that seems to correspond to updating arbitrary states
    - `PUT /states` still exists for updating a user's own state
    - :question: Should there be an activity for this?  Should all state users be able to update their state's info?  Only the coordinator?
    - :question: Should there be an activity that allows a federal user or admin to update a state's info?
  - `PUT /users` and `PUT /users/:id` are removed because there is no activity that seems to correspond to updating arbitrary users
    - users can currently not edit their own information at all; we can follow a similar approach like with states where `PUT /users` allows a user to edit their own information
    - :question: Should there be an activity that allows a federal user or admin to update a user's info?

Closes #362.

- [ ] Product has approved the experience (paging @NAretakis)


api/endpoint-tests/states/get.js
```diff
@@ -75,7 +74,7 @@ tap.test('states/:id endpoint | GET /states/:id', async getStateTest => {
   await db().seed.run();
 
   unauthenticatedTest('get', url, getStateTest);
-  unauthorizedTest('get', url, getStateTest);
+  // unauthorizedTest('get', url, getStateTest);
```
brendansudol: should we delete this?

mgwalker: Yep.  Thought I'd gotten those.  Nice catch.

mgwalker:
Yeah, the commented out code is mostly around the `PUT /states/:id` endpoint.  I'm not sure if it ***should*** be removed, but it is for now.   I'll go ahead and remove the commented code and we can use git history to fish it out if we realize we need it later.

PR #431, opened 2018-04-19
Fix a couple of bugs with the database migration tooling

mgwalker:
- the new squashed migration couldn't rollback properly because it tried to delete a table that did not exist ('apd' instead of 'apds')
- the `delete-everything` shared seed was missing a couple of tables that reference the activities table, so if anything happened to be in those tables, the seed would fail with a key constraint violation
- the `roles-and-activities` shared seed (the one that's responsible for setting up roles and activities in staging, production, etc.) tried to do some fancy asynchronous logic for setting up the mapping between roles and activities, but that was breaking.  Instead, just build the mappings up and do one big insert.  This works consistently.  The old way was super silly.

### DIFF
--- a/api/seeds/shared/roles-and-activities.js
+++ b/api/seeds/shared/roles-and-activities.js
@@ -77,22 +77,18 @@ const setupMappings = async (
   const match = (roleID, activityID) => role =>
     role.role_id === roleID && role.activity_id === activityID;
 
-  await Promise.all(
-    Object.keys(mappings).map(async role => {
-      const roleID = roleIDs[role];
-      for (let i = 0; i < mappings[role].length; i += 1) {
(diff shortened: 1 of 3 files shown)


api/seeds/shared/roles-and-activities.js
```diff
@@ -77,22 +77,18 @@ const setupMappings = async (
...
+      const activityID = activityIDs[activity];
+      if (!alreadyExisting.some(match(roleID, activityID))) {
+        inserts.push({ role_id: roleID, activity_id: activityID });
       }
-    })
-  );
+    });
+  });
+
+  await table.insert(inserts);
```
brendansudol: +1 for simplicity

PR #442, opened 2018-04-20
reduxing goals

brendansudol:
* changed `updateActivity` action / reducer function to use awesome `updeep` library
* reduxed new `ActivityDetailGoals` component
* created helper components for `Input` and `Textarea` (that create input and associated labels); right now, these are in `components/Inputs2` -- once we nuke `redux-form`, we can replace the original `components/Inputs` (which have some redux-form specific structure and props)

### DIFF
--- a/web/src/components/Icons.js
+++ b/web/src/components/Icons.js
@@ -5,6 +5,7 @@ import faHelp from '@fortawesome/fontawesome-free-regular/faQuestionCircle';
 import faBell from '@fortawesome/fontawesome-free-solid/faBell';
 import faChevronDown from '@fortawesome/fontawesome-free-solid/faChevronDown';
 import faChevronUp from '@fortawesome/fontawesome-free-solid/faChevronUp';
+import faCode from '@fortawesome/fontawesome-free-solid/faCode';
 import faCog from '@fortawesome/fontawesome-free-solid/faCog';
 import faHelpSolid from '@fortawesome/fontawesome-free-solid/faQuestionCircle';
 import faSignOut from '@fortawesome/fontawesome-free-solid/faSignOutAlt';
@@ -15,6 +16,7 @@ export {
   faBell,
   faChevronDown,
   faChevronUp,
+  faCode,
   faCog,
   faSignOut
 };
--- a/web/src/components/Inputs2.js
+++ b/web/src/components/Inputs2.js
@@ -1,0 +1,55 @@
+import PropTypes from 'prop-types';
+import React from 'react';
+
+const Input = props => <input className="m0 input" {...props} />;
+const Textarea = props => (
+  <textarea className="m0 textarea" spellCheck="true" {...props} />
+);
+
+// a HOC for Text / Textarea input that includes
+// div wrapper and accompanying label
+const makeInput = InputInner => {
+  const InputHolder = ({
+    name,
+    label,
+    type,
+    hideLabel,
+    className,
+    ...rest
+  }) => (
+    <div className={className || 'mb2'}>
+      <label
+        htmlFor={name}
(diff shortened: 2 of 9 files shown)


web/src/components/Icons.js
```diff
@@ -15,6 +16,7 @@ export {
   faBell,
   faChevronDown,
   faChevronUp,
+  faCode,
```
brendansudol: this is needed for `DataInspector` component (forgot to add this to last PR) 🤦‍♂

web/src/components/Inputs2.js
```diff
@@ -0,0 +1,55 @@
...
+};
+
+makeInput.propTypes = {
+  Component: PropTypes.element
+};
+
+const InputHolder = makeInput(Input);
+const TextareaHolder = makeInput(Textarea);
+
+export { InputHolder as Input, TextareaHolder as Textarea };
```
mgwalker: These are the same wrappers as in `Inputs.js` but without the redux-form entanglement, is that right?

web/src/containers/ActivityDetailGoals.js
```diff
@@ -1,16 +1,76 @@
...
+          List the goals you’re hoping to accomplish as part of this activity:
+        </div>
+        {activity.goals.map((d, i) => (
+          <div key={i} className="mb3">
+            <Textarea
+              name={`goal-${i}`}
+              label={`Goal #${i + 1}`}
+              rows="5"
+              value={d.desc}
+              onChange={this.handleChange(i, 'desc')}
```
mgwalker: Should these keys map to the keys in the object going back and forth with the API?

web/src/reducers/activities.js
```diff
@@ -1,27 +1,23 @@
-import { ADD_ACTIVITY, UPDATE_ACTIVITY } from '../actions/activities';
+import u from 'updeep';
+
+import {
+  ADD_ACTIVITY,
+  ADD_ACTIVITY_GOAL,
+  UPDATE_ACTIVITY
+} from '../actions/activities';
+
+const newGoal = () => ({ desc: '', obj: '' });
```
mgwalker: In this code, there's one "objective" per goal.  In the current data model, there are multiple objectives per goal.  Was this a change from product?  If so, we should also make an issue to update the data model too.  If not, we can hash out the redux specifics when we go to hook everything up.

web/src/reducers/activities.js
```diff
@@ -39,8 +35,27 @@ const reducer = (state = initialState, action) => {
...
+          }
+        },
+        state
+      );
     case UPDATE_ACTIVITY:
-      return updateEntry(state, action);
+      return u(
+        {
+          byId: {
+            [action.id]: { ...action.updates }
```
mgwalker: ***Nice***

web/src/reducers/activities.js
brendansudol: this was a change driven by a conversation with nick a couple weeks ago

web/src/containers/ActivityDetailGoals.js
brendansudol: possibly, though using the array index may be more efficient - O(1) lookup vs. O(n) using `find`, though if we structure the goals as an object vs an array of objects, then it would be the same...

web/src/components/Inputs2.js
brendansudol: yep, exactly, and fully intend to completely replace once redux-form is gone

web/src/containers/ActivityDetailGoals.js
mgwalker: Naw, I just meant `description` instead of `desc`.  😜

PR #448, opened 2018-04-23
"reduxify" expenses

### LINKED ISSUE #438: Reduxify activities: expenses
Hook up the activities expenses section of the site to the application state so that it can easily be synced with the API.

mgwalker:
Adds reduxified expenses form to the app

Closes #438

- the `/activities` page now includes an "Expenses" subsection

### DIFF
--- a/web/src/containers/ActivityDetailExpenses.js
+++ b/web/src/containers/ActivityDetailExpenses.js
@@ -8,7 +8,7 @@ import {
   updateActivity as updateActivityAction
 } from '../actions/activities';
 import Collapsible from '../components/Collapsible';
-import { Textarea } from '../components/Inputs2';
+import { Input, Textarea } from '../components/Inputs2';
 
 class ActivityDetailExpenses extends Component {
   handleChange = (index, key) => e => {
@@ -85,9 +85,10 @@ class ActivityDetailExpenses extends Component {
                   </td>
                   {years.map(year => (
                     <td key={year}>
-                      <input
+                      <Input
+                        name={`expense-${i}-${year}-cost`}
+                        label={`Cost for ${year}`}
                         type="number"
-                        className="m0 input"
                         value={expense.years[year]}
                         onChange={this.handleYearChange(i, year)}
                       />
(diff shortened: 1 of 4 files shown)


web/src/containers/ActivityDetailExpenses.js
```diff
@@ -0,0 +1,155 @@
...
+                      label="Describe the expense"
+                      rows="3"
+                      spellCheck="true"
+                      value={expense.desc}
+                      onChange={this.handleChange(i, 'desc')}
+                    />
+                  </td>
+                  {years.map(year => (
+                    <td key={year}>
+                      <input
```
mgwalker: I realize this needs to be switched to capital-eye `<Input>`...  Fix incoming

web/src/reducers/activities.js
```diff
@@ -64,6 +77,29 @@ const reducer = (state = initialState, action) => {
         },
         state
       );
+    case ADD_ACTIVITY_EXPENSE:
+      return u(
+        {
+          byId: {
+            [action.id]: {
+              expenses: expenses => [...expenses, newExpense(expenses.length)]
```
brendansudol: i don't think this will work -- let's say there are 3 expenses (0, 1, 2); if you delete the first one (0) and then add a new one, it will get an idx of 2 which will be a repeat.

3 options i think: do it similarly to `milestones` (no explicit id/idx field, just uses the index of the array element); create a `generateUniqueId` function and use that for new expenses; loop through existing expenses and get max idx and then add 1 to that.

web/src/containers/ActivityDetailExpenses.js
```diff
@@ -0,0 +1,156 @@
...
+  activity: PropTypes.object.isRequired,
+  years: PropTypes.array.isRequired,
+  addExpense: PropTypes.func.isRequired,
+  removeExpense: PropTypes.func.isRequired,
+  updateActivity: PropTypes.func.isRequired
+};
+
+const mapStateToProps = ({ activities: { byId } }, { aId }) => {
+  const { expenses } = byId[aId];
+  const expenseYears = expenses
```
brendansudol: this is fine for now; i think soon we want years to be an array field keyed to the *apd* that applies to all of their children activities, but this will be easy to sub in when we get there

PR #450, opened 2018-04-24
Move some text into resource files

mgwalker:
For the reduxified "activities" section components that are currently in master, moves most (hopefully *all*) of the user-oriented text into resource files.  To make it easier on @quarterback, I also added a way to see what keys the components are looking for to get their text.

- hopefully no immediately visible changes
- in a non-production environment, add `#i18n` to the address to turn off text resolution and see the keys that the components are looking for:

screen shot 2018-04-24 at 9 48 54 am

- [x] Design has approved the experience (ping @quarterback)
- [ ] Product has approved the experience (ping @jeromeleecms or @NAretakis)

### DIFF
--- a/web/src/containers/ActivityDetailAll.js
+++ b/web/src/containers/ActivityDetailAll.js
@@ -7,9 +7,10 @@ import ActivityDetailGoals from './ActivityDetailGoals';
 import ActivityDetailSchedule from './ActivityDetailSchedule';
 import ActivityDetailStandardsAndConditions from './ActivityDetailStandardsAndConditions';
 import Collapsible from '../components/Collapsible';
+import { t } from '../i18n';
 
 const activityTitle = (a, i) => {
-  let title = `Activity ${i}`;
+  let title = `${t('activities.namePrefix')} ${i}`;
   if (a.name) title += `: ${a.name}`;
   if (a.types.length) title += ` (${a.types.join(', ')})`;
   return title;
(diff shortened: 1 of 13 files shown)


web/src/containers/ActivityDetailAll.js
```diff
@@ -31,7 +32,7 @@ ActivityDetailAll.propTypes = {
 
 const mapStateToProps = ({ activities: { byId } }, { aId, num }) => {
   const activity = byId[aId];
-  const title = `Program Activities › ${activityTitle(activity, num)}`;
+  const title = `${t('activities.header')} › ${activityTitle(activity, num)}`;
```
mgwalker: Not wild about this one...  I think something like:

```js
t('activities.header', { number: num, name: a.name })
```

would be preferable, but I wasn't sure if I18n supported conditional parameters.  That is, unclear if it could differentiate between the cases where the activity did and didn't have a name.  I suppose we could have `activities.header` and `activities.headerAnonymous`?

I dunno, but I'm not wild about splitting up the text this way, and it does leave some user-visible text in the code (the `>`).

web/src/i18n/locales/en.json
```diff
@@ -1,3 +1,117 @@
...
+      },
+      "minimizeCost": {
+        "title":
+          "Strategies to Minimize Cost and Difficulty on Alternative Hardware or Operating System",
+        "description":
+          "Sed ut perspiciatis unde omnis iste natus error sit voluptatem accusantium doloremque laudantium, totam rem aperiam, eaque ipsa quae ab illo inventore veritatis et quasi architecto...",
+        "prompt": "Describe how you'll meet the Cost Minimization condition"
+      }
+    }
+  }
```
mgwalker: Would love thoughts on this nesting structure - should we have that, or should it be flat?  I.e., do we want `activities.standardsAndConditions.minimizeCost` or do we want (e.g.) `act-standards-minimize`?  The former is more complicated but the latter might be harder for humans to interpret.

quarterback: I think you're right on this re: interpretation.

quarterback:
I'm good with this setup. @mgwalker

PR #451, opened 2018-04-24
delete activity (with confirmation)

brendansudol:
### This pull request...
- adds ability to remove/delete an activity
- adds a confirmation modal (forked from https://github.com/GA-MO/react-confirm-alert) so that user doesn't hastily delete an activity and lose all the associated info unless they're sure.

Preview:
https://cl.ly/r7F9

screen shot 2018-04-24 at 11 27 28 am

Adding @lauraponce as well to get feedback on the look and feel of this

### DIFF
--- a/web/src/containers/Activities.js
+++ b/web/src/containers/Activities.js
@@ -1,5 +1,5 @@
 import PropTypes from 'prop-types';
-import React, { Fragment } from 'react';
+import React from 'react';
 import { connect } from 'react-redux';
 
 import ActivityDetailAll from './ActivityDetailAll';
@@ -29,9 +29,7 @@ const Activities = ({ activityIds, addActivity }) => (
         </button>
       </Collapsible>
       {activityIds.map((aId, idx) => (
-        <Fragment key={aId}>
-          <ActivityDetailAll aId={aId} num={idx + 1} />
-        </Fragment>
+        <ActivityDetailAll key={aId} aId={aId} num={idx + 1} />
       ))}
     </Section>
   </Container>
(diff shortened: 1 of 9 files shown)


web/src/containers/Activities.js
```diff
@@ -29,9 +29,7 @@ const Activities = ({ activityIds, addActivity }) => (
         </button>
       </Collapsible>
       {activityIds.map((aId, idx) => (
-        <Fragment key={aId}>
-          <ActivityDetailAll aId={aId} num={idx + 1} />
-        </Fragment>
```
mgwalker: Oops.  I think I did this.  😬  Thanks for fixing.

web/src/containers/DeleteActivity.js
```diff
@@ -0,0 +1,70 @@
...
+
+DeleteConfirm.propTypes = {
+  onClose: PropTypes.func.isRequired,
+  onConfirm: PropTypes.func.isRequired
+};
+
+class DeleteActivity extends Component {
+  handleClick = () => {
+    confirmAlert({
+      customUI: ({ onClose }) => (
```
mgwalker: One thing that bugs me a little bit about React sometimes is the amount of redirection that happens.  It can make it hard to track where things are actually happening.  This is an example of that.  I don't know if there's a much better way...  Maybe move the `confirmAlert(...` inside the `DeleteConfirm` component?  Then the `DeleteActivity` component wouldn't need to know about `confirmAlert`.  But I don't know that's really useful since they're both in this same module...  🤔

I mean, this is fine, it's just one of those cases where I look at React and realize it's not, in fact, ***perfect***.  😆

brendansudol: i agree with you -- this is certainly kinda gross / hard to follow at the least; will think about how to clean up / refine

PR #456, opened 2018-04-25
index vs id tweaks

brendansudol:

### This pull request...
- changes expense id field to be `id`, to be more explicitly different than the index of an expense entry within the expenses array (which we also use to update a field within an entry).
- fixes a bug around expense deletion (we were using index instead of id)
- creates a `Select` component
- small style tweaks

### DIFF
--- a/web/src/actions/activities.js
+++ b/web/src/actions/activities.js
@@ -13,10 +13,10 @@ export const addActivityGoal = id => ({ type: ADD_ACTIVITY_GOAL, id });
 
 export const addActivityExpense = id => ({ type: ADD_ACTIVITY_EXPENSE, id });
 
-export const removeActivityExpense = (id, expenseIdx) => ({
+export const removeActivityExpense = (id, expenseId) => ({
   type: REMOVE_ACTIVITY_EXPENSE,
   id,
-  expenseIdx
+  expenseId
 });
 
 export const addActivityMilestone = id => ({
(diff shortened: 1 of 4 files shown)


web/src/actions/activities.js
```diff
@@ -13,10 +13,10 @@ export const addActivityGoal = id => ({ type: ADD_ACTIVITY_GOAL, id });
 
 export const addActivityExpense = id => ({ type: ADD_ACTIVITY_EXPENSE, id });
 
-export const removeActivityExpense = (id, expenseIdx) => ({
+export const removeActivityExpense = (id, expenseId) => ({
   type: REMOVE_ACTIVITY_EXPENSE,
   id,
-  expenseIdx
+  expenseId
```
mgwalker: I'm 💯 for this.  I goofed.  😛

web/src/components/Select.js
```diff
@@ -0,0 +1,31 @@
...
+
+const Select = ({ name, options, label, hideLabel, wrapperClass, ...rest }) => (
+  <div className={wrapperClass || 'mb2'}>
+    <label
+      htmlFor={name}
+      className={hideLabel ? 'sr-only' : 'block mb-tiny bold'}
+    >
+      {label}
+    </label>
+    <select className="m0 select" {...rest}>
```
mgwalker: Should this have an `id=${name}` so the label has something to relate to?

brendansudol: good catch, yes it does, fixing and merging, thx!

PR #457, opened 2018-04-25
reduxify cost allocation

brendansudol:
### This pull request...
- adds state management for cost allocation and other funding section
- consolidates editor data conversion helpers
- improves naming of optional css class applied to wrapping div of inputs  

### DIFF
--- a/web/src/components/Inputs2.js
+++ b/web/src/components/Inputs2.js
@@ -14,10 +14,10 @@ const makeInput = InputInner => {
     label,
     type,
     hideLabel,
-    className,
+    wrapperClass,
     ...rest
   }) => (
-    <div className={className || 'mb2'}>
+    <div className={wrapperClass || 'mb2'}>
       <label
         htmlFor={name}
         className={hideLabel ? 'sr-only' : 'block mb-tiny bold'}
@@ -33,13 +33,13 @@ const makeInput = InputInner => {
     label: PropTypes.string.isRequired,
     type: PropTypes.string,
     hideLabel: PropTypes.bool,
-    className: PropTypes.string
+    wrapperClass: PropTypes.string
   };
 
   InputHolder.defaultProps = {
     type: 'text',
     hideLabel: false,
-    className: ''
+    wrapperClass: ''
   };
 
   return InputHolder;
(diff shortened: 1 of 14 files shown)


web/src/containers/ActivityDetailCostAllocate.js
```diff
@@ -0,0 +1,89 @@
...
+
+    const { costAllocateDesc, otherFundingDesc } = props.activity;
+
+    this.state = {
+      costAllocateDesc: htmlToEditor(costAllocateDesc),
+      otherFundingDesc: htmlToEditor(otherFundingDesc)
+    };
+  }
+
+  onEditorChange = name => editorState => {
```
brendansudol: we can/should create a HOC when there's an editor component so that these methods/component state can be abstracted away

web/src/containers/ActivityDetailStandardsAndConditions.js
```diff
@@ -23,36 +23,21 @@ class ActivityDetailStandardsAndConditions extends Component {
     return (
       <Collapsible title={t('activities.standardsAndConditions.title')}>
         <p>{t('activities.standardsAndConditions.subheader')}</p>
-        {STANDARDS.map(std => {
-          const inputId = `a-${activity.id}-standards-${std.id}`;
```
brendansudol: don't need this anymore (since we're leveraging the `Textarea` component which uses the name prop and automatically applies that to the corresponding label `htmlFor`)

web/src/components/Inputs2.js
```diff
@@ -14,10 +14,10 @@ const makeInput = InputInner => {
     label,
     type,
     hideLabel,
-    className,
+    wrapperClass,
```
mgwalker: This allows setting a class on the native `<input>`?  That's pretty slick.

mgwalker: Also, destructuring the props ***and*** having a `...rest`...  I feel like you've done that before but it just now clicked this is happening.  Very cool.

web/src/containers/ActivityDetailCostAllocate.js
mgwalker: `<RichTextarea>`  😛

Issue #454, opened 2018-04-25
Switch CI/CD build to use more deterministic dependency install


mgwalker:
Currently we use `npm install` to install dependencies in our CI/CD builds (and after #452, our deployed libraries as well).  This does some fancy dance moves to determine which versions to install, and we can end up with different versions every time we CI runs.

We can switch to `npm ci` (after #453), which installs exclusively from the package lockfile.  It's a faster install because there's less version searching and matching to do, but also we can be sure that what's used in CI matches whatever we last installed locally.

As a nice bonus, `npm ci` never modifies the package files where as `npm install` can modify the lockfile even if nothing actually changed.  The result is that `npm install` can mess up the CircleCI cache by changing the hash of the lockfile.  `npm ci` won't do that.  :tada:

NAretakis:
@mgwalker is this a thing we should still do? Maybe part of the refinement milestone? Or is this dependent on infrastructure?

mgwalker:
Oh!  This is done.

PR #489, opened 2018-04-30
Update build to npm@6

### LINKED ISSUE #453: Switch to npm@6 (dev/build tool)
The latest version of npm is faster (again) and comes with new features to enable more deterministic installs.

mgwalker:
Updates dev and CI environment to use npm version 6.x.  This allows us to take advantage of the `npm ci` command immediately, which produces a deterministic dependency install.  In the future, we will also be able to benefit from npm's soon-to-be built-in security checking, as an additional check alongside the static dependency analysis we're currently getting from [Gemnasium](https://gemnasium.com/github.com/18F/cms-hitech-apd).

Closes #453 #454

- no visible changes

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -1,23 +1,35 @@
 version: 2
 jobs:
-  prepare web dependencies cache:
+  prepare dependencies cache:
     docker:
       - image: node:8
     steps:
       - checkout
       - restore_cache:
-          key: v1-cms-hitech-apd-{{ checksum "web/package-lock.json" }}
+          key: cms-hitech-apd-npm-{{ checksum "web/package-lock.json" }}-{{ checksum "api/package-lock.json" }}
       - run:
           name: update npm
           command: npm install -g npm@6
       - run:
-          name: install dependencies
+          name: install web dependencies
           working_directory: ~/project/web
           command: npm ci
+      - run:
+          name: install api dependencies
+          working_directory: ~/project/api
+          command: npm ci
+      - save_cache:
+          key: cms-hitech-apd-npm-{{ checksum "web/package-lock.json" }}-{{ checksum "api/package-lock.json" }}
+          paths:
+            - /root/.npm
       - save_cache:
           key: v1-cms-hitech-apd-{{ checksum "web/package-lock.json" }}
           paths:
             - web/node_modules
+      - save_cache:
+          key: v1-cms-hitech-api-{{ checksum "api/package-lock.json" }}
+          paths:
+            - api/node_modules
 
   web lint:
     docker:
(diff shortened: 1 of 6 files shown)


.circleci/config.yml
```diff
@@ -1,20 +1,35 @@
...
+          paths:
+            - /root/.npm
       - save_cache:
           key: v1-cms-hitech-apd-{{ checksum "web/package-lock.json" }}
           paths:
             - web/node_modules
+      - save_cache:
+          key: v1-cms-hitech-api-{{ checksum "api/package-lock.json" }}
+          paths:
+            - api/node_modules
```
mgwalker: Instead of separate tasks for installing web and api dependencies, now we have just one dependencies task.  It slows down the build slightly, but we have less duplication in the config file now.  🤷‍♂️

.circleci/config.yml
```diff
@@ -1,20 +1,35 @@
...
-          command: npm install
+          command: npm ci
+      - run:
+          name: install api dependencies
+          working_directory: ~/project/api
+          command: npm ci
+      - save_cache:
+          key: cms-hitech-apd-npm-{{ checksum "web/package-lock.json" }}-{{ checksum "api/package-lock.json" }}
+          paths:
+            - /root/.npm
```
mgwalker: This saves off the npm cache.  `npm ci` deletes `node_modules` if it already exists and populates it fresh from `package-lock.json`.  We can still cache off `node_modules` within a build run (so downstream steps can use it), but it's useless between runs.  Instead, we can save the npm cache to speed up future runs of `npm ci`.

.circleci/config.yml
```diff
@@ -135,9 +134,10 @@ jobs:
     steps:
       - checkout
       - restore_cache:
-          key: v1-cms-hitech-apd-{{ checksum "web/package-lock.json" }}
-      - restore_cache:
-          key: v1-cms-hitech-api-{{ checksum "api/package-lock.json" }}
+          key: cms-hitech-apd-npm-{{ checksum "web/package-lock.json" }}-{{ checksum "api/package-lock.json" }}
```
mgwalker: The deploy steps all run `npm ci`, so restore the npm cache before running them.

bin/deploy-ux-testing.sh
```diff
@@ -29,6 +29,7 @@ rm -rf seeds/test
 rm seeds/shared/delete-everything.js
 rm seeds/development.js
 rm seeds/test.js
+npm ci
```
mgwalker: Here's a place where the deploy behavior was dependent on the external environment - if `node_modules` was already populated, we'd get a vendored dependency deploy; otherwise, we wouldn't.  Not a huge fan of that kind of "accidental" behavior.  This way it's more explicit - we get a vendored deploy.

brendansudol: 👍

PR #501, opened 2018-05-03
[DEPRECATED] Update dependencies

mgwalker:
Update dependencies.  This should get us the latest performance and security enhancements.  See [Gemnasium](https://gemnasium.com/github.com/18F/cms-hitech-apd) for the current list of outdated dependencies.

### Not updated
- **webpack**
  There were big changes from 3.x to 4.x.  We'll need to rejigger our build scripts a little bit, though probably not much.  I will look into that in the next day or two and may end up rolling it into this PR if it's a small enough change.
- **webpack-dev-server**
  The dev-server is in lockstep with webpack, so as long as don't update webpack, we should also not update the dev-server.

- **bookshelf.js**
  There was a change somewhere along the line that broke how our `sync` method works _in one specific case_.  When trying to update an activity, it is being hyper-specific, keying in on every column in the table instead of just using the primary key.  This would be fine, except we allow some of those columns to be null and under the hood, bookshelf converts those nulls to `undefined`.  Then when knex receives the `undefined` values, it blows a gasket.  I do not know why bookshelf insists on these hyper-specific queries, but...  well, it will be some effort to figure out what's happening and fix our code to work with the updated bookshelf, so for now, we'll just stay outdated.

- no visible changes, or else we messed up 😄 


mgwalker:
Switched to Snyk and not getting vulnerability notices anymore, so no obvious ***need*** to update dependencies.  So...  let's just not for right now.

Issue #394, opened 2018-04-03
Hook up frontend to API


mgwalker:
For the first pass, we're going to hardcode the APD ID based on the database seed and everyone will load and save the same APD.  That will get our frontend and backend hooked back together and we can go from there.

mgwalker:
Documenting data deficiencies as I find them

### APD
- no APD key personnel in the data model or API return
- no APD years in the data model or API return (captured by #462, I think)

### activities
- no activity types (ie., HIT, HIE, MMIS) in the data model or API return
- no short description/summary in the data model or API return
- no alternative approach in the data model or API return
- no cost allocation methodology in the data model or API return

### activity goals
- data model has multiple objectives instead of just one

### activity schedule/milestones
- :bug: can't save them, so can't fetch them... 😬

### activity expenses
- :bug: descriptions in the data model are attached to the years instead of the expense itself 😬
- :bug: data model has a name field, but not a category field

### activity contractor resources
- start/end dates are currently ISO 8601 date string, but the `<input type="date">` expects a `yyyy-mm-dd` format; should we update the API to use the HTML date spec, or format the ISO 8601 string into the HTML date?
  - advantage of the HTML date string is that we just grab it and bind it to our input elements
  - advantage of the ISO 8601 is that it doesn't tie our raw data to HTML
  - as a stopgap, I convert the ISO date string into an HTML date string in the activity reducer
  - it's possible that `yyyy-mm-dd` is a valid ISO 8601 date...  should check into that, because if it is, we should just use that throughout (the data model defines those columns as date types, which is probably good for queryability, but the API can do the conversion to simple date strings before returning them)

### activity standards and conditions
- the data model doesn't have them

### OpenAPI
- doesn't list state personnel as part of an activity
- doesn't list contractor resources as part of an activity

NAretakis:
For the items @mgwalker listed above, are there new issues that need to be created?

mgwalker:
Wow, I did a terrible job of keeping track of this issue.  All of those are fixed.  😛

NAretakis:
Closed is closed ;)

Issue #462, opened 2018-04-25
[Program Overview Section] Federal Fiscal Year back-end work 


NAretakis:
Allow the selection of FFYs within the program overview section drive all of the tables and content through out the rest of the APD.

### This task is done when...
 - [ ] FFY selection is available within the Overview section
 - [ ] FFYs selected within the Overview section drive all subsequent tables within the APD.

brendansudol:
This is done on the front-end @NAretakis, but i think there is still a little back-end work left

mgwalker:
I think this one is done with #605.

PR #544, opened 2018-05-15
Hookup APD frontend to backend

mgwalker:
https://hitech-apd-pr544.app.cloud.gov/

- makes the main page "protected" (meaning, you have to login)
- on successful login or auth check, go ahead and fetch the user's APDs and use the first one returned as the current APD
  - an "auth check" happens when you first load the page - if you're already logged in, it just proceeds; otherwise it drops you at the login screen
  - there's some future work indicated here: how to pick which APD the user should be looking at, maybe there's a landing page before this, etc.
- populates the application state with the data from the API
- adds a temporary save button that saves the application state to the API

Addresses #394

- can save and resume editing an APD

### DIFF
--- a/web/src/actions/apd.js
+++ b/web/src/actions/apd.js
@@ -22,7 +22,10 @@ export const fetchApd = () => dispatch => {
 
   return axios
     .get(url)
-    .then(req => dispatch(receiveApd(req.data)))
+    .then(req => {
+      const apd = Array.isArray(req.data) ? req.data[0] : null;
+      dispatch(receiveApd(apd));
+    })
     .catch(error => {
       const reason = error.response ? error.response.data : 'N/A';
       dispatch(failApd(reason));
--- a/web/src/actions/apd.js
+++ b/web/src/actions/apd.js
@@ -140,5 +140,8 @@ export const saveApd = () => (dispatch, state) => {
     });
   });
 
-  console.log(JSON.stringify(apd));
+  return axios
+    .put(`/apds/${updatedApd.id}`, apd)
+    .then(res => dispatch(saveSuccess(res.data)))
+    .catch(() => dispatch(saveFailure())); // TODO handle the error
 };
--- a/web/src/actions/apd.js
+++ b/web/src/actions/apd.js
@@ -84,7 +84,11 @@ export const saveApd = () => (dispatch, state) => {
         description: g.desc,
         objectives: [{ description: g.obj }]
       })),
-      schedule: null, // TODO
+      schedule: activity.milestones.map(m => ({
+        milestone: m.name,
+        plannedStart: m.start,
+        plannedEnd: m.end
+      })),
       statePersonnel: activity.statePersonnel.map(s => ({
         title: s.title,
         description: s.desc,
@@ -97,8 +101,8 @@ export const saveApd = () => (dispatch, state) => {
       contractorResources: activity.contractorResources.map(c => ({
         name: c.name,
         description: c.desc,
-        start: new Date(c.start).toISOString(),
-        end: new Date(c.end).toISOString(),
+        start: c.start,
+        end: c.end,
         years: Object.keys(c.years).map(year => ({
           cost: +c.years[year],
           year
--- a/web/src/actions/apd.js
+++ b/web/src/actions/apd.js
@@ -59,7 +59,15 @@ export const saveApd = () => (dispatch, state) => {
 
   const apd = {
     activities: [],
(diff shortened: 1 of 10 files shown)


web/src/routes.js
```diff
@@ -6,10 +6,10 @@ import PoC from './components/PoC';
 import Login from './containers/Login';
 
 const routes = [
-  { path: '/', component: PoC, exact: true, nonPrivate: true },
+  { path: '/', component: PoC, exact: true, nonPrivate: false },
```
mgwalker: You have to be authenticated in order to send the API requests added in this PR, so...  there's this.

web/src/actions/apd.js
```diff
@@ -38,3 +48,100 @@ export const fetchApdDataIfNeeded = () => (dispatch, getState) => {
...
+      description: activity.descLong,
+      alternatives: activity.altApproach,
+      costAllocationMethodology: activity.costAllocateDesc,
+      otherFundingSources: {
+        description: activity.otherFundingDesc,
+        amount: activity.otherFundingAmt
+      },
+      goals: activity.goals.map(g => ({
+        description: g.desc,
+        objectives: [{ description: g.obj }] // TODO - objective should mape 1:1 to goals, instead of being an array (needs backend change first)
```
mgwalker: See #550. Once we figure out how the API will send/receive goals and objectives, we should update this.

web/src/actions/apd.test.js
```diff
@@ -35,7 +35,7 @@ describe('state actions', () => {
 
     it('creates GET_APD_SUCCESS after successful APD fetch', () => {
       const store = mockStore({});
-      fetchMock.onGet('/apds').reply(200, { foo: 'bar' });
+      fetchMock.onGet('/apds').reply(200, [{ foo: 'bar' }]);
```
mgwalker: The `GET /apds` endpoint returns an array

web/src/components/Inputs.js
```diff
@@ -9,10 +9,21 @@ class RichText extends Component {
...
+      lastContent: props.content
     };
   }
 
+  shouldComponentUpdate(nextProps, nextState) {
+    if (nextProps.content !== this.state.lastContent) {
+      this.setState({
+        content: htmlToEditor(nextProps.content),
+        lastContent: nextProps.content
+      });
```
mgwalker: Don't automatically return true here.  Doing so causes a non-stop render loop.  Instead, trigger a state update and wait for the state change before updating the component.

web/src/reducers/activities.js
```diff
@@ -256,6 +258,96 @@ const reducer = (state = initialState, action) => {
...
+      const byId = {};
+      action.data.activities.forEach(a => {
+        byId[a.id] = {
+          id: a.id,
+          name: a.name,
+          types: ['HIT'], // TODO
+          descShort: '', // TODO
+          descLong: a.description,
+          altApproach: '', // TODO
+          costAllocateDesc: '', // TODO
```
mgwalker: These 4 `TODO`s are partially represented in #554. After that one is done, we'll need to come back and update these.

web/src/reducers/activities.js
```diff
@@ -256,6 +258,96 @@ const reducer = (state = initialState, action) => {
...
+          otherFundingDesc: a.otherFundingSources.description,
+          otherFundingAmt: +a.otherFundingSources.amount,
+          goals: a.goals.map(g => ({
+            desc: g.description,
+            obj: g.objectives[0] // TODO - don't assume there's one objective; tie objective directly to the goal instead of a mapped table
```
mgwalker: #550

web/src/reducers/apd.js
```diff
@@ -44,7 +44,15 @@ const reducer = (state = initialState, action) => {
         ...state,
         fetching: false,
         loaded: true,
-        data: { ...action.data[0] }
+        data: {
+          id: action.data.id,
+          years: ['2018'], // TODO
```
mgwalker: #462 probably; lots to digest on this one

mgwalker:
@brendansudol Sorry I added you as a reviewer earlier, but feeling pretty good this is ready to go now.

@NAretakis @jeromeleecms Note that the API enforces certain pieces of data integrity right now, and we don't have UI to indicate when the API rejects your save.  The biggest ones you're likely to run into are 1) activity names must not be empty and they must be unique within the APD; and 2) goals must not be empty (this one will be addressed in #550).

web/src/reducers/activities.js
brendansudol: 👍

Issue #529, opened 2018-05-11
Create a sticky header for the activities


lauraponce:
When users are scrolling through the activities section, they lose the reference point of the specific activity they are entering information for. The name of the activity should stick to the top of the page and stay there when they are seeing the corresponding subsections.

In the screenshot below, the "Program Activities > Activity 1: SLR Administration (HIT)" would be sticky.

### This task is done when...
 - [ ] A sticky header option has been explored
 - [ ] The sticky header has been implemented
 
![screen shot 2018-05-11 at 11 27 19 am](https://user-images.githubusercontent.com/21222704/39932444-2ca8b820-550e-11e8-93b5-eb5411f5b59b.png)

brendansudol:
done!

PR #547, opened 2018-05-15
sticky activity header

brendansudol:
- makes activity section header sticky
- adds this nifty library (https://github.com/airbnb/deline) which helps when there are a lot of atomic classes

Addresses: https://github.com/18F/cms-hitech-apd/issues/529

Preview:
https://cl.ly/rc70

### DIFF
--- a/web/src/components/Collapsible.js
+++ b/web/src/components/Collapsible.js
@@ -1,3 +1,4 @@
+import deline from 'deline';
 import kebabCase from 'lodash.kebabcase';
 import PropTypes from 'prop-types';
 import React, { Component } from 'react';
@@ -16,15 +17,21 @@ class Collapsible extends Component {
   };
 
   render() {
-    const { bgColor, children, id, title, onChange } = this.props;
+    const { bgColor, children, id, sticky, title, onChange } = this.props;
     const isOpen = onChange ? this.props.open : this.state.open;
     const contentId = kebabCase(title);
+
+    const btnClass = deline`
+      btn block col-12 py2 h3 line-height-1 left-align
+      ${isOpen ? 'border-bottom border-bottom-silver' : ''}
+      bg-${bgColor} ${sticky ? 'sticky-top' : ''}
+    `;
 
     return (
       <div id={id} className={`mb2 bg-${bgColor} border border-silver`}>
         <button
           type="button"
-          className="btn block col-12 left-align h3 py2 line-height-1"
+          className={btnClass}
           aria-expanded={isOpen}
           aria-controls={contentId}
           onClick={onChange || this.handleClick}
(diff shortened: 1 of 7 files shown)


web/src/containers/ActivityDetailAll.js
```diff
@@ -22,25 +22,36 @@ const activityTitle = (a, i) => {
...
-    <ActivityDetailStatePersonnel aId={aId} />
-    <ActivityDetailContractorResources aId={aId} />
-    <ActivityDetailExpenses aId={aId} />
-    <ActivityDetailCostAllocate aId={aId} />
-    <ActivityDetailStandardsAndConditions aId={aId} />
-    <DeleteActivity aId={aId} />
-  </Collapsible>
-);
+class ActivityDetailAll extends Component {
+  handleChange = id => () => {
```
brendansudol: for performance, better to memoize this method so it doesn't get recreated on each render

web/src/components/Collapsible.js
```diff
@@ -16,15 +17,21 @@ class Collapsible extends Component {
...
-    const { bgColor, children, id, title, onChange } = this.props;
+    const { bgColor, children, id, sticky, title, onChange } = this.props;
     const isOpen = onChange ? this.props.open : this.state.open;
     const contentId = kebabCase(title);
 
+    const btnClass = deline`
+      btn block col-12 py2 h3 line-height-1 left-align
+      ${isOpen ? 'border-bottom border-bottom-silver' : ''}
+      bg-${bgColor} ${sticky ? 'sticky-top' : ''}
+    `;
```
mgwalker: Neat

web/src/styles/app/layout.css
```diff
@@ -53,3 +53,8 @@ footer {
 .w-3 {
   width: 3rem;
 }
+
+.sticky-top {
+  position: sticky;
```
mgwalker: If we need to support IE, we can't rely on `position: sticky;`.  I think that's a product question.

@jeromeleecms @NAretakis - are there specific minimum versions of browsers that we should be targeting?  Should we informally poll states to ask what browsers and versions they use?  (Have we already and I forgot? 😛)

brendansudol: i see this particular feature as a progressive enhancement -- if your browser supports it, you get a nice sticky header; if not, the experience is still perfectly fine but you now have to scroll a little if you forget which activity you're working on

we should def ask states what browser and version they're using (i guess we can use google analytics too and see for ourselves)

NAretakis: I agree with Brendan. I can do some informal polling on browsers, but as long as they can go through the experience in some way, we should be ok.

mgwalker:
Here's the URL to have a look at it live, btw:  https://hitech-apd-pr547.app.cloud.gov/
Just needs design approval!

PR #565, opened 2018-05-16
add markdown component

brendansudol:
I did a bunch of experimenting with a good way to integrate our translation strings with markdown support. Wasn't super happy with any of the component APIs I was creating, but here's something to get the conversation started. I'm very happy with alternative ideas.

This keeps the translations strings and the convert-markdown-to-html steps separate, mainly because there are times when we want both i think.

Example usage of this component:
`<Md wrapper="h3" content={t('apd.helpText')} className="regular h1" />`

The wrapper represents the enclosing html element and any additional props (i.e.,  `className`) get passed along too. Why `md.renderInline` and not `md.render`? Because `render` adds paragraph tags around the text which we probably don't want. I would like to support passing in either an element string or a component, but that can come down the road. (i.e., `Md wrapper={SectionTitle}... `)

### DIFF
--- a/web/package-lock.json
+++ b/web/package-lock.json
@@ -772,7 +772,6 @@
       "version": "1.0.9",
       "resolved": "https://registry.npmjs.org/argparse/-/argparse-1.0.9.tgz",
       "integrity": "sha1-c9g7wmP4bpf4zE9rrhsOkKfSLIY=",
-      "dev": true,
       "requires": {
         "sprintf-js": "1.0.3"
       }
@@ -4486,8 +4485,7 @@
     "entities": {
       "version": "1.1.1",
       "resolved": "https://registry.npmjs.org/entities/-/entities-1.1.1.tgz",
-      "integrity": "sha1-blwtClYhtdra7O+AuQ7ftc13cvA=",
-      "dev": true
+      "integrity": "sha1-blwtClYhtdra7O+AuQ7ftc13cvA="
     },
     "enzyme": {
       "version": "3.3.0",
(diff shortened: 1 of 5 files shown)


mgwalker:
This is really cool.  It's going to be a huge win.

Some thoughts/questions for you...

1. What's the thinking around being able to specify a container element?  My naive assumption is that having the markdown rendered into a span or div that has a `className` would satisfy the 90% case.

2. I tried locally switching to just `render` instead of `renderInline` and there was another important difference: because header elements are block elements, `# headers` don't get rendered when you call `renderInline`.  What weirdness/undesirable behavior do you foresee with the `<p>` inserted around the text elements when using `render`?  Is that worth the tradeoff of losing headers?

3. In the slightly longer view, what would you think about having this text represented by a single string in the JSON file?

screen shot 2018-05-17 at 1 47 43 pm

Something like

```json
{
  "programSummaryFrontMatter": "# Program Summary\n\nCMS recommends that states reach out to their regional CMS HITECH contacts prior to developing an HIE funding request. HIE funding requests are complex, and early consultation with CMS facilitates a more efficient approval process."
}
```

It impacts our ability to target those individual textual pieces with styles, but I'm curious if that's really a concern.  🤔

brendansudol:
1. What's the thinking around being able to specify a container element? My naive assumption is that having the markdown rendered into a span or div that has a className would satisfy the 90% case.

**mainly just flexibility; i can see times where a span, div, p, or header may be appropriate. I guess we could enforce that headers be defined in the string (i.e., `# title...`), but my thinking was that markdown in the strings would be more of the exception vs. the rule so that content folks can focus more on the words than whether there needs to be a `##` or a `###` before a phrase (and then we could handle that with the appropriate container element).**

I tried locally switching to just render instead of renderInline and there was another important difference: because header elements are block elements, # headers don't get rendered when you call renderInline. What weirdness/undesirable behavior do you foresee with the <p> inserted around the text elements when using render? Is that worth the tradeoff of losing headers?

**i think there are a couple problems with the auto <p>s -- semantically, it's not good to have these extra p elements wrapping things like headers; but the main problem is that p is a block element that inherits some default styling (like margin bottom); and we shouldn't assume that all translated text should be block and have that same bottom margin (i.e., for a label). My thinking was that this is worth the tradeoff of losing headers because we can simply add those headers using the container element (the downside here is that the whole text would then be a header but our translated strings already seem to be at that lower level anyway)**

In the slightly longer view, what would you think about having this text represented by a single string in the JSON file?

**I'm torn -- on my hand, it's simpler to fully utilize the markdown language, but on the other, we lose some styling control and our content file becomes a bigger mix of content and presentational syntax.**

brendansudol:
appreciate your thoughts and questions @mgwalker!

mgwalker:
Okay, I completely buy into all of that.

Issue #515, opened 2018-05-08
Add dollar sign & percentage sign in budget tables


quarterback:
In the application, when you insert dollar amounts and dates in the calendar, the formats are displayed differently depending on what field you're using. This is confusing.

### This task is done when...
 - [x] dates & dollar fields are standardized.

quarterback:
I notice the dates are now fixed, so I removed that from the title.

brendansudol:
this is done!

PR #566, opened 2018-05-17
add form field masking

brendansudol:
### This pull request...
- adds this nifty library (https://text-mask.github.io/text-mask/) so we can have nicely formatted inputs as users type stuff
- adds dollar and percent `MaskedInput`s to the appropriate activity form fields 

Addresses: https://github.com/18F/cms-hitech-apd/issues/515

Preview:
https://cl.ly/rdPp

### DIFF
--- a/web/src/components/Inputs.js
+++ b/web/src/components/Inputs.js
@@ -4,6 +4,7 @@ import { Editor } from 'react-draft-wysiwyg';
 import TextMask from 'react-text-mask';
 
 import { EDITOR_CONFIG, htmlToEditor, editorToHtml } from '../util/editor';
+import MASKS from '../util/masks';
 
 class RichText extends Component {
   constructor(props) {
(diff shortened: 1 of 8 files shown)


web/src/containers/ActivityDetailContractorResources.js
```diff
@@ -26,7 +27,7 @@ class ActivityDetailContractorExpenses extends Component {
     const { activity, updateActivity } = this.props;
 
     const updates = {
-      contractorResources: { [index]: { years: { [year]: +value } } }
+      contractorResources: { [index]: { years: { [year]: value } } }
```
mgwalker: I'm not wild about this.  The text mask input sends a formatted string (e.g., `$1,000.00`) with the `onChange` event.  That's problematic for two reasons:

1. if we add that formatted string to our internal state, we can't easily use it for math later on
2. we want to store numeric data on the API for the same reason

This change will actually break #544 _right now_, so it's not a future worry.  Would it be sensible to further wrap the text mask with specific `<DollarInput>` and `<PercentInput>` components that de-format the text and send numbers in their `onChange` events?

brendansudol: yes, i hear you and agree -- i feel like this library should really handle this (returning the raw and formatted values) -- maybe it does, i'll check!

brendansudol: alright, see what you think!

web/src/components/Inputs.js
```diff
@@ -57,6 +59,44 @@ const Textarea = props => (
   <textarea className="m0 textarea" spellCheck="true" {...props} />
 );
 
+const makeMaskedInput = ({ mask, unmask }) => {
+  // TODO: split some of these components into their own file
+  // eslint-disable-next-line react/no-multi-comp
```
mgwalker: 👍

web/src/util/masks.js
```diff
@@ -0,0 +1,16 @@
+import createNumberMask from 'text-mask-addons/dist/createNumberMask';
+
+const toNum = x => (x.length ? +x.replace(/\D+/g, '') : '');
```
mgwalker: Only concern is I think this will convert `$123.45` to `12345`.

PR #571, opened 2018-05-18
A bit of deployment cleanup

mgwalker:
Run per-PR deployment cleanup every two hours instead of just once a night.  Don't install cf CLI plugins we don't use (autopilot, in this case).  Don't make lots of work for ourselves by manually calling the cloud.gov API to delete apps and their routes - use the CLI's built-in command instead.

- no visible changes


brendansudol:
CI stuck?

mgwalker:
Not exactly stuck, but it turns out CircleCI uses a more strict `cron` format, so it was choking on the `*/2` (even numbered hours).  I changed it to list out all the hours and now it seems happy.  :shrug:

PR #621, opened 2018-05-30
budget table driven by data

brendansudol:
### This pull request...
- aggregates activity data across type (HIT, HIE, etc) and years for summary budget table, so now the budget table is driven by real activity data, not that fakey fake stuff (though the fed and state share percents are still static)

i'm not happy with the current data aggregation abstractions, but alas, baby steps... 


web/src/util/index.js
```diff
@@ -1,5 +1,7 @@
 /* eslint-disable import/prefer-default-export */
 
+export const ACTIVITY_TYPES = ['HIT', 'HIE', 'MMIS'];
```
mgwalker: What do you think about calling this something like `ACTIVITY_FUNDING_SOURCES`?  Or `ACTIVITY_FFP_SOURCE` (Federal funding participation)?  When I see `types` elsewhere in the code, I think about something else altogether.  I'm not sure ***what*** I think, but I consistently have to stop and think about what it means.

web/src/containers/BudgetSummary.js
```diff
@@ -0,0 +1,228 @@
...
+    Object.keys(datum).forEach(key => {
+      obj[key] = (obj[key] || 0) + datum[key]; // eslint-disable-line no-param-reassign
+    });
+    return obj;
+  }, {});
+
+const getDataByActivityType = activities =>
+  ACTIVITY_TYPES.reduce((obj, id) => {
+    // filter by activity type (this isn't great because we'll be
+    // double-counting activities that belong to 2+ types
```
mgwalker: I noticed this playing with the pr-deploy.  Glad to see this comment here.  I think we should open an issue to capture that this needs to be addressed.  I wonder if this is another level of detail in cost allocation...

PR #622, opened 2018-05-30
APD versions model and initial endpoint

mgwalker:
- Adds an APD versions database table and model
- Adds an endpoint at `POST /apds/:id/versions` to create a new version - this will also set the APD's status to `in review`
- Updates the `userCanEditAPD` middleware to fail if the APD status is not `draft`

- no visible changes yet!

- [x] Tests have been updated (and all tests are passing) - needs tests for the new endpoint

- [ ] Product has approved the experience: cc @jeromeleecms @NAretakis


NAretakis:
If we end up with a version number, or version code thingy, it might be worth putting that at the top of the page somewhere. I remember testers interested in seeing that they are working on "Version 1" or something. We might need to come up with a version naming convention.

NAretakis:
How about:
[State Code][Program]APDv[version number]
or
TX-HITECH-APD-v6
Thoughts @jeromeleecms ?

mgwalker:
Would this version number be for states to see as they're drafting it, or for CMS staff to see as they're reviewing it?  Or both?  (Logical OR versus semantic "or"!)

NAretakis:
Version stuff is for both States and CMS I think.

PR #629, opened 2018-06-01
Hook up the budget summary table to the actual data in the form

mgwalker:
Well, hopefully anyway.  🤞

- changing an activity type, any expense value (state personnel cost, contractor cost, or non-personnel cost), or cost allocation updates the budget summary table with updated numbers


web/src/actions/activities.js
```diff
@@ -77,8 +77,9 @@ export const toggleActivitySection = id => ({
   id
 });
 
-export const updateActivity = (id, updates) => ({
```
mgwalker: Flag added so we don't trigger the budget recalculation unless numbers changed.

web/src/containers/BudgetSummary.js
```diff
@@ -59,9 +58,9 @@ class DataRow extends Component {
             const val = data[yr];
             return (
               <Fragment key={yr}>
-                <td className="mono right-align">{formatMoney(val)}</td>
-                <td className="mono right-align">{formatMoney(val * 0.9)}</td>
-                <td className="mono right-align">{formatMoney(val * 0.1)}</td>
+                <td className="mono right-align">{formatMoney(val.total)}</td>
+                <td className="mono right-align">{formatMoney(val.federal)}</td>
+                <td className="mono right-align">{formatMoney(val.state)}</td>
```
mgwalker: Doing these calculations in the budget reducer instead

web/src/reducers/budget.js
```diff
@@ -0,0 +1,149 @@
...
+  combined: getFundingSourcesByYear()
+});
+
+const initialState = () => ({
+  combined: getFundingSourcesByYear(),
+  hie: expenseTypes(),
+  hit: expenseTypes(),
+  mmis: expenseTypes(),
+  years: YEAR_OPTIONS
+});
```
mgwalker: These are all functions so that when I'm building the budget table from scratch, all of these properties point to unique objects.  If I'd done something like this, they'd point to the same instances:

```js
const expenseTypes = {
  statePersonnel: {},
  contractors: {},
  expenses: {},
  combined: {}
};

const initialState = {
  combined: {},
  hie: expenseTypes,
  hit: expenseTypes,
  mmis: expenseTypes
}

initialState.hie.statePersonnel === initialState.mmis.statePersonnel // true
```

That's totally fine for a pure "initial state" object, but it doesn't work when we're actually trying to fill it in with data.  And rather than try to figure out which piece of budget data changed, it made way more sense to me to just build the whole table each time.  So...  Here we are.  🙂

web/src/reducers/index.js
```diff
@@ -1,19 +1,22 @@
...
+    apd: apd(oldState.apd, action),
+    auth: auth(oldState.auth, action),
+    budget: {},
+    state: state(oldState.state, action),
+    user: user(oldState.user, action),
+    router: routerReducer(oldState.router, action)
+  };
+  newState.budget = budget(oldState.budget, action, newState);
+  return newState;
+};
```
mgwalker: I don't ***love*** this, but in order to make the budget reducer useful, it needs access to the `activities` state.  `combineReducers` doesn't allow for that.  But it's not hard to mimic what `combineReducers` does, so I did that.  I'd hoped I could just call `rootReducer(state, action)`, but redux actually throws out any state properties that aren't defined in `combineReducers`, so budget stuff was getting lost and we were getting a console error.

web/src/containers/BudgetSummary.js
brendansudol: 💥

web/src/reducers/budget.js
brendansudol: 👍

web/src/reducers/index.js
brendansudol: i don't love this either, though i do understand why we're doing it.

this is basically following bullet 2 from this FAQ: https://redux.js.org/faq/reducers#how-do-i-share-state-between-two-reducers-do-i-have-to-use-combinereducers 

i'm curious for your thoughts on bullets 1 and 3:
* 1 would basically be having `budget` be nested under the `activities` reducer
* 3 would be augmenting our `updateActivity` action and adding `thunk` here so that we can add the activities state into the action payload if `isExpense = true`

option 3 sounds kinda appealing to me in that we keep the budget reducer separate and don't have to change the combineReducer but I'd love your thoughts.

mgwalker: 🤔 It occurs to me there's another issue at play here that I forgot about when I wrote that earlier text - the `budgets` reducer relies on the updated `activities` state - that is, the `activities` reducer must run first, and its output is fed into the `budgets` reducer.  It's not great.

I do like bullet 3 though.  What do you think about doing something like this?

- `updateActivity` is called
- it dispatches an `UPDATE_ACTIVITY` action; the budget reducer no longer response to this action
- using thunkification, we wait for that action to finish
- if `isExpense` is true, dispatch an `UPDATE_BUDGET` action; the budget reducer responds

Then we get `combineReducers` back and we don't have to worry about the order reducers are run in response to an action - there's an explicit action for updating the budget data.

mgwalker: Did that and it's ***SO MUCH CLEANER***.  Thanks for this feedback!

brendansudol: awesome! this is great 🥇

Issue #532, opened 2018-05-11
Build incentive payments table


lauraponce:
Build out the table that was designed and approved in #503

NAretakis:
Latest version for reference: 
![image](https://user-images.githubusercontent.com/34281281/40800135-be80fd34-64cc-11e8-9611-7eafcf0c775b.png)

NAretakis:
Also, this table is blank, and states simply fill in the cells. This is not driven from any content.

PR #653, opened 2018-06-07
incentive payments table

brendansudol:
preview:
https://cl.ly/s9VY

Closes https://github.com/18F/cms-hitech-apd/issues/532

### DIFF
--- a/web/src/containers/IncentivePayments.js
+++ b/web/src/containers/IncentivePayments.js
@@ -1,0 +1,123 @@
+import PropTypes from 'prop-types';
+import React, { Component, Fragment } from 'react';
+import { connect } from 'react-redux';
+
+import { updateApd as updateApdAction } from '../actions/apd';
+import { Input, DollarInput } from '../components/Inputs';
+import { INCENTIVE_ENTRIES } from '../util';
+import { formatMoney, formatNum } from '../util/formats';
+
+const YEARS = ['2018'];
+const QUARTERS = [1, 2, 3, 4];
+
+class IncentivePayments extends Component {
+  handleChange = (key, year, quarter) => e => {
+    const { value } = e.target;
+    const incentivePayments = { [key]: { [year]: { [quarter]: value } } };
+    this.props.updateApd({ incentivePayments });
+  };
+
+  render() {
+    const { data, totals } = this.props;
+
+    return (
+      <div className="py1 overflow-auto">
+        <table className="h6 table-fixed table-bordered">
+          <thead>
+            <tr>
+              <th style={{ width: 200 }} />
+              {YEARS.map(year => (
+                <th key={year} className="bg-teal white center" colSpan="5">
+                  FFY {year}
+                </th>
+              ))}
+              <th className="bg-black white center">Total</th>
+            </tr>
+            <tr>
+              <th />
+              {YEARS.map(year => (
+                <Fragment key={year}>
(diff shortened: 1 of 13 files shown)


web/src/styles/app/color.css
```diff
@@ -7,22 +7,22 @@
   background-color: #e1f3f8;
 }
 
-.bg-light-blue {
+.bg-blue-light {
```
brendansudol: better/easier for interpolation (`${color}-light`)

web/src/reducers/apd.js
```diff
@@ -45,6 +56,7 @@ const reducer = (state = initialState, action) => {
         fetching: false,
         loaded: true,
         data: {
+          ...state.data,
```
mgwalker: Not sure about this.  It creates a scenario where the local data could supersede the server data.  I don't think it's something that should realistically happen, but...  🤷‍♂️?  The reason for the spread in the root object is that the `fetching`, `loading`, etc. properties are client-specific metadata and not represented in the server data at all.

What's your thinking here?

web/src/styles/app/color.css
mgwalker: 👍 One question, it looks like the background color for the individual activities was lost.  Does it need to be updated to reverse the color/modifier in the `className` too?

web/src/containers/IncentivePayments.js
```diff
@@ -0,0 +1,142 @@
...
+import { updateApd as updateApdAction } from '../actions/apd';
+import { Input, DollarInput } from '../components/Inputs';
+import { t } from '../i18n';
+import { INCENTIVE_ENTRIES } from '../util';
+import { formatMoney, formatNum } from '../util/formats';
+
+const QUARTERS = [1, 2, 3, 4];
+const COLORS = {
+  '2018': 'bg-teal',
+  '2019': 'bg-green'
```
mgwalker: 😬 hardcoded years!  Could this just be an array and have it map to the index of the years array instead of being tied to specific years?  I think the second argument to the `Array.map` callback is the index?  (Would have to look it up, but that seems right.)

web/src/styles/app/color.css
brendansudol: fixed

web/src/reducers/apd.js
brendansudol: yeah, it would supercede if the server data didn't include a particular key/value (the idea being so that we're guaranteed to have at least the initial keys in the apd/activity object), but i agree, it's still weird and not sure what i was thinking -- i updated :)

web/src/containers/IncentivePayments.js
brendansudol: good call, this is gross!

brendansudol:
thx @mgwalker -- incorporated feedback and created new issue for backend work (https://github.com/18F/cms-hitech-apd/issues/663)

web/src/containers/IncentivePayments.js
```diff
@@ -24,8 +21,8 @@ class IncentivePayments extends Component {
   render() {
     const { data, totals, years } = this.props;
 
-    const yearsWithColors = years.map(year => {
-      const color = COLORS[year] || 'bg-gray';
+    const yearsWithColors = years.map((year, i) => {
+      const color = `bg-${COLORS[i] || 'gray'}`;
```
mgwalker: 🎉

PR #654, opened 2018-06-07
Update the deploy bot's message when it redeploys a PR app

mgwalker:
Currently, once the bot has commented on a PR, it doesn't comment again.  This will change the bot's behavior so that it will delete its old comment and create a new one every time it deploys the app.  That way its comment will always be at the bottom and we'll know that it updated.

- The PR bot message will be the most recent comment after it finishes a deploy, so we can tell when it redeploys after changes are made

@jeromeleecms Feel free to just close this if you don't want the change in behavior, I won't mind.  😉


jeromeleecms:
@mgwalker If we make this change am I going to get a flood of email updates from the bot? I like knowing when the app has been deployed, but ... is there an unintended consequence of increases in email volume here?

PR #661, opened 2018-06-08
Fix: Deleting personnel, contractor, or expense rows triggers a budget table update

### LINKED ISSUE #645: Removing personnel, contractors, or other costs does not remove dollar amounts from budgets
The bug is the same for state personnel, contractors, or non-personnel costs.

If a user inputs dollar amounts into these fields, and then subsequently uses the X to remove the actual field (row), the dollar amounts appear to still be summed within the Summary Budget and subsequent sections. 

What should happen is that when one of the fields is removed, is that the resulting budget tables no longer sum the deleted values.

mgwalker:
Also fixes a bug where you couldn't add new personnel, contractor, or expense rows.  😬

Closes #645

- deleting any of those kinds of rows will update the budget table
- deleting an activity will update the budget table
- you can add more of those kinds of rows

- ~Tests have been updated (and all tests are passing)~ Big tests overhaul coming soon, this will get wrapped up into that

### DIFF
--- a/web/src/reducers/budget.js
+++ b/web/src/reducers/budget.js
@@ -36,6 +36,10 @@ const initialState = years => ({
 
 const activities = src => Object.values(src.activities.byId);
 
+// The default for from is to handle the case where an expense
+// block doesn't have any rows.  In that case, from is never
+// populated.  But that block doesn't conribute to the budget,
+// either, so easiest thing to do is just zero it out.
 const addBudgetBlocks = (into, from = { total: 0, federal: 0, state: 0 }) => {
   const out = into;
   out.total += from.total;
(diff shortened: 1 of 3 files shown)


web/src/reducers/budget.js
```diff
@@ -36,7 +36,7 @@ const initialState = years => ({
 
 const activities = src => Object.values(src.activities.byId);
 
-const addBudgetBlocks = (into, from) => {
+const addBudgetBlocks = (into, from = { total: 0, federal: 0, state: 0 }) => {
```
mgwalker: Handles the case where an expense type (state personnel, contractor, non-personnel) doesn't have any rows.

mgwalker:
Hang on to this one a sec, I think we also need to handle the case when a whole activity is deleted.  😛

Issue #611, opened 2018-05-25
Develop certify and submission experience


quarterback:
Need to have a certify and submit section that completes the APD creation and submission process.

### This task is done when...
 - [ ] Develop a submission experience
 - [ ] Design reviews upon implementation

quarterback:
Copy in this section is complete. What we need is a submit button at the end of the application beneath submit/certify where the copy now lives. @brendansudol @mgwalker // cc: @NAretakis

Issue #497, opened 2018-05-02
Design HIE activities section


quarterback:
Design the sections of HIE specific content after the walkthrough of HIE is done. Related to issue #473 

### This task is done when...
 - [x] Write content for each of the HIE sections.
- [x] Write activity specific content
- [x] Create a related issue on design of these sections once content is implemented in prototype to decide whether it can exist as-is or needs more design refinement.

NAretakis:
@quarterback so, Jerome and I talked this through and I think we have a good, low LOE, approach for getting this coded. Basically, what we'd like to do is use the "HIE Checkbox" 
![image](https://user-images.githubusercontent.com/34281281/40516906-0d204740-5f70-11e8-974d-5b7a80e3d2ec.png)

to drive a special rich text box that appears within the Activity Description section of the corresponding activity.

How this would work is that when a user checks the "HIE" box, a special chunk of helptext appears for "Detailed Activity Description" section. And this special helptext addresses [these HIE questions (link to MU, Value Proposition, & Sustainability)](https://app.mural.co/t/gsa6/m/gsa6/1524714012930/d34a6666c3cb4db8dfee9cfadc5e4c13e7a88f72?wid=0-1524756276698)

![image](https://user-images.githubusercontent.com/34281281/40516978-5b83111a-5f70-11e8-9ca2-45453e54cfa6.png)

The special helptext only needs to appear when the "HIE" checkbox is used, otherwise, the default description is just fine. Does this make sense?

quarterback:
@NAretakis @jeromeleecms yes, this makes sense!

quarterback:
I created a dev issue #607

quarterback:
(HIE specific text above HIE detailed activity section)

**Detailed Activity Description** (section title)
(help text draft)
If you list HIE activities below, describe the activity in detail to include links to meaningful use, value proposition to providers, and a sustainability plan. Additionally, include diagrams as needed if they provide more context for how you'll complete the activity outlined.

_What I need here, is for @NAretakis to take a look and make sense of this from an analyst perspective, then I'd be able to refine it more clearly. Do we need more detail than above?_

quarterback:
I need to talk to @brendansudol @NAretakis about this particular issue, so we can get the check boxes added.

quarterback:
@brendansudol @mgwalker we need a section for HIE activities. Rather than do a "toggle", the idea is just to add this as another subsection under activity list. 

[...]

Issue #471, opened 2018-04-25
[Certify and Submit] Design frontend for certify and submit section


NAretakis:
Need to have a certify and submit section that completes the APD creation and submission process.

### This task is done when...
 - [x] Wireframe for certify and submit section is constructed
 - [x] Design hands task off to Development.

NAretakis:
Lightweight. Envisioned to have a checkbox, or maybe one of those _type your name_ to certify the submission as complete.

It also needs a bit of language letting the users know that after submitting, the APD will be locked for editing until CMS responds. Then maybe a big green submit button. 

Big green button is what should trigger the versioning thingy that @mgwalker has been working on.

quarterback:
Working on this week.

quarterback:
I've sent up a PR address this and #611. What I'll need is for @mgwalker / @brendansudol to convert the check boxes in assurances and compliance from text to actual check box options since I couldn't do that in the yaml file. (I just did them as text so you'd know what should be in that space.) Also write the copy headers for those sections. Will address submit/certify in that one.

PR #665, opened 2018-06-08
Content Additions

quarterback:
Meant to address #471, #611, and #497. 
Comments in each issue for devs. Adds content sections for submit/certify, as well as assurances text boxes which need to be replaced to be actual checkboxes. Added submit text copy. HIE activity section added to yaml, design of the section (a text box with specific text) mockup included inside of issues.


web/src/i18n/locales/en.yaml
```diff
@@ -245,28 +256,28 @@
...
+      helpText: "If the APD is approved, these program budget tables will be referenced on the CMS approval letter."
     budgetTable: 
       title: "Program Budget Table"
       helpText: ""
   certifyAndSubmit: 
     title: "Certify and Submit"
-    helpText: "By clicking Submit, you agree that your funding request fulfills our standards and requirements."
-    certify: 
+    helpText: "Note, this application will be locked upon submission until after a CMS analyst reviews it."
+    certify: "Note, this application will be locked upon submission until after a CMS analyst reviews it."
```
mgwalker: This `certify` is a mapping entry, meaning it has children.  An entry cannot have both a text value and children.  So this is invalid:

```yaml
certify: "some text"
    title: "other text"
    subtitle: "bloop"
```

but this is valid:

```yaml
certify:
    name: "some text"
    title: "other text"
    subtitle: "bloop"
```

Which one is meant here?  It looks like `certify` and `helpText` have the same text value, so I'm guessing only `helpText` needs that content?

quarterback: Yeah, only help text needs the content.

web/src/i18n/locales/en.yaml
```diff
@@ -75,6 +75,17 @@
...
+title: "HIE Activity Details"	
+helpText: "Describe the activity in detail including images, diagrams, or other relevant information. If this an HIE activity, the activity details must also include a description of three items: 	
+- How the activity is linked to meaningful use;	
+- Value proposition to providers;	
+- Sustainability plan describing how the activity will be sustained after HIE DDI funding is exhausted.	
+Additionally, include diagrams and descriptions of other HIE linkages as needed if they provide more context for how you'll complete the activity."
+list:
+title: "Activity List"	
+title: "Program Activities"
+helpText: "Identify each of the activities referenced from the overview section. CMS will collect information for each activity."
```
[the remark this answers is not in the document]
quarterback: I can reconstruct that to make it easier. But yes, all of that belongs to HIE activities.

quarterback:
Fixing the indentation and redundancies, merging with other changes & will resubmit.

PR #668, opened 2018-06-11
HIE activities

quarterback:


web/src/i18n/locales/en.yaml
```diff
@@ -150,15 +154,15 @@
       removeLabel: "Remove resource"
       addContractorButtonText: "Add resource"
     statePersonnel: 
-      title: "State Personnel"
+      title: "Key Personnel"
       subheader: "Outline the personnel resources for this section. This includes anyone funded as a part of this activity. This does not include personnel listed in the contract section."
-      helpText: "Personnel costs must include salaries and benefits."
+      helpText: "Personnel costs must include salaries and benefits. FTE% will be generated automatically, after you list a salary amount in the table below."
```
mgwalker: Is this supposed to happen?  Because it doesn't happen now and I'm not sure how it would.  We combine salary with %FTE to calculate total cost, but we need two of those three, not just one.  😛

Issue #584, opened 2018-05-22
Build quarterly budget table (v2)


brendansudol:
refine work from https://github.com/18F/cms-hitech-apd/issues/535

PR #677, opened 2018-06-12
wire up quarterly budget summary

brendansudol:
- (see title)

closes https://github.com/18F/cms-hitech-apd/issues/584


brendansudol:
i broke this up into 3 tables (one for each funding source); if that's not right, please let me know!

NAretakis:
Can we combine HIT and HIE? Those actually need to be combined into one table, but MMIS still should be on its own.

jeromeleecms:
@NAretakis I thought we were going to make these fields all editable (though default to divide by 4)? Similar to the incentive table with fields.

NAretakis:
@jeromeleecms yes, we need to format and situate the table, and have it filled from the summary amounts. Then we can make it editable.

web/src/reducers/budget.js
```diff
@@ -31,7 +31,7 @@ const initialState = years => ({
   hie: expenseTypes(years),
   hit: expenseTypes(years),
   mmis: expenseTypes(years),
-  years: []
+  years
```
mgwalker: 😆 Why didn't I fix this when I was doing the tests?  Thank you.

web/src/containers/QuarterlyBudgetSummary.js
```diff
@@ -0,0 +1,120 @@
...
+  combined: 'Total Enhanced FFP'
+};
+
+const color = idx => `bg-${COLORS[idx] || 'gray'}`;
+
+const QuarterlyBudgetSummary = ({ budget }) => {
+  const { years } = budget;
+
+  // wait until budget is loaded
+  if (!years.length) return null;
```
mgwalker: Smart!

web/src/containers/QuarterlyBudgetSummary.js
```diff
@@ -0,0 +1,120 @@
...
+                        <th className={`right-align ${color(i)}-light`}>
+                          {t('table.subtotal')}
+                        </th>
+                      </Fragment>
+                    ))}
+                    <th className="bg-gray-light" />
+                  </tr>
+                </thead>
+                <tbody>
+                  {Object.keys(data).map(key => (
```
mgwalker: Does this assume that `combined` will always be the last key on the data object?  I reckon since the budget data comes from the reducer, we can make that assertion, but if so, we should probably put some comments in the reducer noting that the property order matters.  What do you think?

NAretakis:
Sweet, @brendansudol for the comment on merging HIT and HIE into one table, as well as @jeromeleecms 's comment to allow that table to be editable.

web/src/containers/QuarterlyBudgetSummary.js
brendansudol: good point, i don't like this reliance -- will tweak

brendansudol:
sweet, coming up :)

brendansudol:
that turned out to be more involved that i thought, but it's now ready for re-review @mgwalker @jeromeleecms @NAretakis (once new deployment completes)

web/src/reducers/budget.test.js
```diff
@@ -362,6 +408,32 @@ describe('budget reducer', () => {
...
+          '1933': { federal: 159553.7, state: 33000, total: 207630 },
+          total: { federal: 198848.8, state: 117295.1, total: 637193 }
+        },
+        statePersonnel: {
+          '1931': { federal: 9540.5, state: 11540.5, total: 211081 },
+          '1932': { federal: 26576.9, state: 69576.9, total: 215769 },
+          '1933': { federal: 157973.66, state: 33000, total: 206034 },
+          total: { federal: 194091.06, state: 114117.4, total: 632884 }
+        }
+      },
```
mgwalker: I'm sorry you had to do this.  😆

NAretakis:
This is fantastic! But here's my nitpick issue, when I move the slider for HIT/HIE it also moves the slider for MMIS. Can these be independent?

brendansudol:
wasn't sure about that; yes they can be -- give me 20 mins :)

NAretakis:
A++

jeromeleecms:
@brendansudol Dev Magic :100:

Issue #652, opened 2018-06-07
Build out Results of Previous Activities section 


lauraponce:
Corresponds to #534 

There should be 3 subsections: previous activities summary, HITECH table, and MMIS table. There is no wireframe for the MMIS table, but it should work similar to the HITECH table.

NAretakis:
@mgwalker before we can finish up this section, we need to construct the table below, and put it underneath the previous activities summary. This is just a blank table, to be filled out by a state user.
table for reference
![image](https://user-images.githubusercontent.com/34281281/41314486-1fffd13a-6e4a-11e8-90c8-df90aaa88a57.png)

mgwalker:
What FFYs should be shown here?

NAretakis:
Ah. Kind of a moving target. Ideally it would be the FFY's approved in the previous CMS approval letter. So, if we can't make it variable (add/delete), we could go with current year + two previous FFYs.

jeromeleecms:
The rendering issue we saw in Firefox during user testing appears to have popped up in Chrome now. 

![image](https://user-images.githubusercontent.com/34279968/42108862-a0fdcbcc-7b90-11e8-9a71-bdc7a2001217.png)

No vertical scroll bar, either.

Just something to monitor during the refinement.

mgwalker:
That one is a data bug.  😬   See #731

NAretakis:
This is done for HIT/HIE. We now just need an MMIS table inserted to close this issue.

PR #691, opened 2018-06-14
Previous activity table

mgwalker:
- previous activity table will exist

### New backend needs
- data model for collecting previous activity approved/actual data (#694)

### DIFF
--- a/web/src/containers/ApdPreviousActivityTable.js
+++ b/web/src/containers/ApdPreviousActivityTable.js
@@ -2,13 +2,14 @@
 import React, { Fragment } from 'react';
 import { connect } from 'react-redux';
 
+import { defaultAPDYearOptions } from '../util';
 import { formatMoney } from '../util/formats';
 import { DollarInput } from '../components/Inputs';
 
 const ApdPreviousActivityTable = () => {
   const sections = { hit: 'HIT', hie: 'HIE', hitAndHie: 'HIT + HIE' };
   const colors = ['aqua', 'blue', 'navy'];
-  const years = ['2018', '2019', '2020'];
+  const years = defaultAPDYearOptions.map(y => +y - 2); // ['2018', '2019', '2020'];
 
   const borderClass = i => `border-left border-${i < 0 ? 'gray' : colors[i]}`;
 
--- a/web/src/containers/ApdPreviousActivityTable.js
+++ b/web/src/containers/ApdPreviousActivityTable.js
@@ -1,0 +1,103 @@
+import PropTypes from 'prop-types';
+import React, { Fragment } from 'react';
+import { connect } from 'react-redux';
+
+import { formatMoney } from '../util/formats';
+import { t } from '../i18n';
+
+const ApdPreviousActivityTable = ({ updateApd }) => {
+  const sections = { hit: 'HIT', hie: 'HIE', hitAndHie: 'HIT + HIE' };
+  const colors = ['aqua', 'blue', 'navy'];
+  const years = ['2018', '2019', '2020'];
+
+  const borderClass = i => `border-left border-${i < 0 ? 'gray' : colors[i]}`;
+
+  return (
+    <table
+      className="h6 table-fixed table-condensed"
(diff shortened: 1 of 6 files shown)


mgwalker:
The styling isn't quite perfect with the design but I was spending too much time fiddling to try to get it right.  @lauraponce what do you think about creating another issue to refine the styling on this table to get it just right?

web/src/containers/ApdPreviousActivityTable.js
```diff
@@ -0,0 +1,162 @@
...
+          ))}
+          <th colSpan="2" className={`border-right ${borderClass(2)}`}>
+            Federal + state<br />Grand total computable
+          </th>
+        </tr>
+        <tr>
+          <th />
+          {Object.values(programs).map((name, i) => (
+            <Fragment key={name}>
+              <th className={`bg-${colors[i]}-light ${borderClass(i)}`}>
```
brendansudol: these dynamic classes (which I've done a bunch of) feel quite messy to me; fine for now but I'm making a mental note - and actual note here :) - to revisit and clean up.

web/src/containers/ApdPreviousActivityTable.js
```diff
@@ -0,0 +1,162 @@
...
+                    label={`actual state share for ${program}, FFY ${year}`}
+                    hideLabel
+                    value={previousActivityExpenses[year][program].stateActual}
+                    onChange={handleChange(year, program, 'stateActual')}
+                  />
+                </td>
+              </Fragment>
+            ))}
+            <td className={`bg-${colors[2]}-light ${borderClass(2)}`}>
+              {formatMoney(27000)}
```
brendansudol: should this be hardcoded?

web/src/containers/ApdPreviousActivityTable.js
```diff
@@ -0,0 +1,162 @@
...
+            </td>
+            <td>{formatMoney(3000)}</td>
```
brendansudol: should this number be hardcoded?

mgwalker: Oh, whops.  Nooope.

NAretakis:
@mgwalker looks good. But the totals are not adding up, should they?

mgwalker:
Updated to compute all the totals values.  Took the opportunity to extra labels into the resource file too.

jeromeleecms:
@mgwalker Both column headers say HIT and HIT (vs HIE) - and yes, the total for HIT+HIE column is the sum of the previous items.

Few additional items (@NAretakis may be we need to discuss?)
1) If we have an HIT and HIE table, shouldn't we have a MMIS table as well?

2) Math works now (I heard computers are good at that!), but can we also have the state share auto-populated as a suggestion based on what the states put in at the 90% share? Or maybe good for MVP and we just need to add that as part of the front end validation.

3) Are the years hard coded? If we keep it this way, may be we should include some additional help text to tell the users to only go back as far as the previous APD (that way we are not making the states go grab data from something that closed a couple years ago when they really don't need to do that). 

4) (Styling Comment) Was it a styling thing with the Federal Share (newline) 90% FFP and State Share (no newline) 10% FFP? If we want to be consistent with our letters -- Federal Share (new line) 90% <-- centered in the column header. Happy to table that for the new styling issue. 

5) (Editorial comment) Last column: Federal + S(can we capitalize "S" for consistency)tate Grand Total Computable

mgwalker:
(Fixed the HIE/HIT headers.)

2. We can do that.  I'd suggest making that another issue and prioritizing, though.  I have questions about how we'd do the math here - Do we still have input boxes for both state and federal pieces, and as they fill in one, we compute the other?  Do we only have input boxes for the state share and always compute the federal?  Only input the federal share and always compute the state?  Etc.

3. Years are based on current FFY and previous two, per https://github.com/18F/cms-hitech-apd/issues/652#issuecomment-397450794, for now.

4. Whoops, those are both supposed to have newlines!  Messed up the resource file.

5. Fixed.

Issue #685, opened 2018-06-14
Change cost allocation section to fix Other Funding and Medicaid match rates


NAretakis:
Need to change how the cost allocation section works. In order to feed the Program Budget tables correctly, we need to:

1. Move the "Other Funding Amount" field into each of the FFY cards
2. Calculate and display the math associated with "Total Cost" minus "Other Funding Amount" = "Medicaid Share"
3. Remove the adjustable percentage allocation boxes
4. Insert a drop down menu that allows for 90/10, 75/25, and 50/50 allocations. 
5. Calculate and display the math showing the "Medicaid Share" broken into Federal and State shares.

Once this is completed, PR #681 can be resumed and will be able to have data fed from Cost Allocation into the Program Budget Tables.

Example flow is below:

![image](https://user-images.githubusercontent.com/34281281/41421452-1ab7ac28-6fb4-11e8-9979-59e23c6752fc.png)

brendansudol:
done!

PR #692, opened 2018-06-18
rework cost allocation

brendansudol:
This PR reworks the cost allocation as per https://github.com/18F/cms-hitech-apd/issues/685.

In summary
* one can specify a per-activity other amount, which comes off the top before computing the federal and state amounts
* the federal / state percentages can now we only 1 of 3 things (90/10, 75/25, 50/50)

The budget summary data munging methods are adjusted to reflect this new math/logic. 

preview:
https://cl.ly/sKrg

### DIFF
--- a/web/src/containers/ActivityDetailCostAllocate.js
+++ b/web/src/containers/ActivityDetailCostAllocate.js
@@ -11,7 +11,7 @@ import { t } from '../i18n';
 
 const ActivityDetailCostAllocate = props => {
   const { activity, updateActivity } = props;
-  const { costAllocateDesc, otherFundingDesc } = activity;
+  const { costAllocationDesc, otherFundingDesc } = activity;
 
   const sync = name => html => {
     updateActivity(activity.id, { [name]: html });
@@ -28,8 +28,8 @@ const ActivityDetailCostAllocate = props => {
           reminder="activities.costAllocate.methodology.reminder"
         />
         <RichText
-          content={costAllocateDesc}
-          onSync={sync('costAllocateDesc')}
+          content={costAllocationDesc}
+          onSync={sync('costAllocationDesc')}
         />
       </div>
       <ActivityDetailCostAllocateFFP aId={activity.id} />
--- a/web/src/containers/ActivityDetailCostAllocate.js
+++ b/web/src/containers/ActivityDetailCostAllocate.js
@@ -5,7 +5,7 @@ import { connect } from 'react-redux';
 import ActivityDetailCostAllocateFFP from './ActivityDetailCostAllocateFFP';
 import { updateActivity as updateActivityAction } from '../actions/activities';
 import { Subsection } from '../components/Section';
-import { DollarInput, RichText } from '../components/Inputs';
+import { RichText } from '../components/Inputs';
 import HelpText from '../components/HelpText';
 import { t } from '../i18n';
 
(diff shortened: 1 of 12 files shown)


web/src/containers/ActivityDetailCostAllocate.js
```diff
@@ -28,8 +28,8 @@ const ActivityDetailCostAllocate = props => {
           reminder="activities.costAllocate.methodology.reminder"
         />
         <RichText
-          content={costAllocateDesc}
-          onSync={sync('costAllocateDesc')}
+          content={costAllocationDesc}
+          onSync={sync('costAllocationDesc')}
```
mgwalker: Out of curiosity, do you have a tool or process for renaming things throughout an app, or do you just do it very carefully?  😛

web/src/containers/ActivityDetailCostAllocateFFP.js
```diff
@@ -1,96 +1,105 @@
...
-      amount: total * percent / 100
+      amount: totalNetOther * ffp[id] / 100
     }));
 
-    return { year, total, ffpLeft, ffpData };
+    return { year, total, totalNetOther, allocations };
   });
 
-  return { byYearData };
+  return { byYearData, costAllocation };
```
mgwalker: As I'm reading this, one of the things I really like about redux and `mapStateToProps` is that I don't have to just *know* what these objects look like - I can go look at the appropriate reducer.  So nice.

web/src/containers/ActivityDetailCostAllocateFFP.js
```diff
@@ -1,96 +1,105 @@
...
-                  <div>{ffpDisplay[id]}</div>
-                  <div className="bold mono">{formatMoney(amount)}</div>
+                <div>{t('activities.costAllocate.ffp.totalNetOther')}</div>
+                <div className="h3 bold mono">{formatMoney(totalNetOther)}</div>
+                <hr />
+                <Select
+                  name={`ffp-${year}`}
+                  label={t('activities.costAllocate.ffp.labels.fedStateSplit')}
+                  options={['90-10', '75-25', '50-50']}
+                  onChange={this.handleFFP(year)}
```
mgwalker: Should there be a `value` prop here so it gets set when loading?  Okay with opening a new issue for that, though.

web/src/reducers/budget.js
```diff
@@ -62,6 +63,18 @@ const initialState = years => ({
...
 
+const budgetInput = (type, value = v => +v, target = null) => ({
+  type,
+  value,
+  target: target || type
+});
+
+const budgetInputs = [
+  budgetInput('statePersonnel', year => +year.amt * +year.perc / 100),
+  budgetInput('contractorResources', undefined, 'contractors'),
```
mgwalker: Maybe we should have a cleanup issue to change this to `contractors` on the server side too, make it consistent with the other two?

web/src/containers/ActivityDetailCostAllocate.js
brendansudol: just `ack` and find/replace and patience :)

web/src/containers/ActivityDetailCostAllocateFFP.js
brendansudol: whoops, good catch; will update

web/src/reducers/budget.js
brendansudol: agreed

brendansudol:
given that nick and jerome are out for a few days @mgwalker, i propose we merge stuff once we feel good able that and then can create follow-up issues for anything we might have missed once CMS is back and takes a look... thoughts?

NAretakis:
This looks great! We'll need to figure out how to talk through this via helptext or examples. It would be nice to visually show how the math is working, but it is definitely working :tada:

Issue #655, opened 2018-06-08
Saving an APD provides feedback on success/fail of saving


jeromeleecms:
Saving an APD by pressing the "Save APD" button does not report whether or not the save was successful or failed. This behavior should be updated to tell the user whether or not the save succeeded or failed.

### This task is done when...
 - [x] Clicking the Save APD notifies user that the APD was "Saved" or "Failed" 
 - [x] If APD failed, app should report a meaningful reason to the user 
- [x] Code has been reviewed by someone else
- [x] Product has approved experience

brendansudol:
maybe having an UI like this? https://material-ui.com/demos/snackbars/

mgwalker:
I think this one was finished by #698?  We do need to make sure we track getting the error messages to make sense, but at least there are success/failure notifications now.

NAretakis:
Yes, I definitely am getting saving and submitting notifications. Closing out.

PR #698, opened 2018-06-21
add notification / snackbar system

brendansudol:
This adds a fairly robust, accessible notification / snackbar system.

To display a notification, one simply has to dispatch a `notify` action with a `message` param. i.e., `dispatch(notify('Success!'))`

The notification auto-hides after a configurable number of seconds, currently 3.

It can also handle consecutive notifications (i.e., triggering one while another is still visible); per [Google's guidelines](https://material.io/archive/guidelines/components/snackbars-toasts.html#snackbars-toasts-usage), when a second snackbar is triggered while the first is displayed, the first should start the contraction motion downwards before the second one animates upwards. I essentially followed this [implementation](https://material-ui.com/demos/snackbars/#consecutive-snackbars).

The impetus of this new component was https://github.com/18F/cms-hitech-apd/issues/655 and that issue should now be resolved.

preview:
https://cl.ly/sSn9


web/src/actions/apd.js
```diff
@@ -184,6 +186,16 @@ export const saveApd = () => (dispatch, state) => {
...
+    .then(res => {
+      dispatch(notify('Save successful!'));
+      dispatch(saveSuccess(res.data));
+    })
+    .catch(error => {
+      // TODO [bren]: map backend error message to content in resource file
+      const { response: res } = error;
+      const reason = (res && (res.data || {}).error) || 'not-sure-why';
+
+      dispatch(notify(`Save failed (${reason})`));
```
mgwalker: Ooh, nice.  I like how this works, will be easy to use all over.

web/src/containers/Notification.js
```diff
@@ -0,0 +1,65 @@
...
+import { processQueue, closeNotification } from '../actions/notification';
+
+class Notification extends Component {
+  handleClose = (_, reason) => {
+    if (reason === 'clickaway') return;
+    this.props.closeNotification();
+  };
+
+  handleExited = () => {
+    this.props.processQueue();
```
mgwalker: And then pop the queue only after the previous notification has transitioned away.  Slick!

jeromeleecms:
@brendansudol I can't get my screen reader to read anything in the app (much less the pop-up). :( Also, can't navigate via keyboard when it pops up.

@nkkl Could you try it out?

web/src/reducers/notification.js
```diff
@@ -0,0 +1,42 @@
...
+      return {
+        ...state,
+        open: false
+      };
+    case PROCESS_QUEUE:
+      return {
+        open: true,
+        messageInfo: state.queue[0],
+        queue: state.queue.slice(1)
+      };
```
mgwalker: This threw me for a loop for a second because it looked like we could end up in an exception state - if we get `PROCESS_QUEUE` but the `queue` is already empty.  (Technically that's still possible...) but the action checks the state and doesn't dispatch the message if the queue is empty.

I'm curious if that logic belongs in the action creator or the reducer.  We've got similar logic in other action creators too, but I guess I hadn't noticed the separation until now.  What do you think?

web/src/containers/Notification.js
```diff
@@ -0,0 +1,65 @@
...
+import React, { Component } from 'react';
+import { connect } from 'react-redux';
+import Snackbar from '@material-ui/core/Snackbar';
+
+import { processQueue, closeNotification } from '../actions/notification';
+
+class Notification extends Component {
+  handleClose = (_, reason) => {
+    if (reason === 'clickaway') return;
+    this.props.closeNotification();
```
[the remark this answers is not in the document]
brendansudol: yep!

web/src/reducers/notification.js
brendansudol: right, i added that logic to the action creator (https://github.com/18F/cms-hitech-apd/pull/698/files#diff-a49e614a3472c0ef0f58f5c1c69f6847R12) so that you can only get to `PROCESS_QUEUE` in reducer if the queue has entries, but i think it'd be fine to be a little defensive in the reducer as well so i might add that conditional

brendansudol:
i'm gonna merge this on -- @jeromeleecms we'll do a more exhaustive accessibility review soon and make sure this is re-checked

PR #709, opened 2018-06-25
Assurances Check Text

quarterback:
Added check box text for Assurances and compliance sections

### This pull request changes...
- [ ]  The experience requires review to ensure it's in the correct format.

PR #712, opened 2018-06-25
Assurances and compliance sections

### LINKED ISSUE #609: Implement Assurances and Compliance section
This is the assurances and compliance section of the application. 

### This task is done when...
 - [x] A working version of the design in #470 is implemented.

mgwalker:
Displays all the regulations that states have to assure their compliance with and provides them radio buttons to indicate yes/no and a place for explainer text.

The list of regulations is stored separately from the other resource text.  The reason is that it is used in multiple places - we must know everything that's in the resource file, but this approach allows us to add or remove regulations without having to tweak the code.

Closes #609.
Folds in changes from #709.

- assurances and compliance section exists

### Needs backend support
- need API to store assurance and compliance info
- API should return assurance and compliance info too


quarterback:
Small typo: Soule Source should read Sole Source.

jeromeleecms:
@mgwalker @quarterback Can we merge the four cards into a single accordion with headers (Instead of having 4 separate accordions? 

Also, styling behaves oddly  when I select it -- can we have:
1) Line Spacing is consistent between line items
2) Spacing for the radio buttons and the text field is all over the place when selecting or reselecting

image

3) I don't think the empty field would pass 508c compliance - it would need to have some sort of prompt about what to put there.

mgwalker:
@jeromeleecms Yep, all do-able.  Any thoughts about what the accordion text would be?

Regarding the input without a label - it actually does have one, but it's marked as `screen-reader only`, so sighted users won't see it but screen readers will still announce it.  The label text is terrible, now that I look at it again, but there's at least something there.  I'll fiddle with that too.

jeromeleecms:
@mgwalker Title of the accordion "Federal Citations and Justification" (certainly open to other suggestions).

Everything else: :1st_place_medal:

NAretakis:
We can use this link for the SMM references: https://www.cms.gov/Regulations-and-Guidance/Guidance/Manuals/Paper-Based-Manuals-Items/CMS021927.html

mgwalker:
Updated to just one accordion, fixed spacing issues (@brendansudol please don't cry), added SMM link.

web/src/containers/AssurancesAndCompliance.js
```diff
@@ -0,0 +1,125 @@
...
+
+  const handleExplanationChange = (section, index) => e => {
+    updateApd({
+      assurancesAndCompliance: {
+        [section]: { [index]: { explanation: e.target.value } }
+      }
+    });
+  };
+
+  console.log(apdSections);
```
brendansudol: ✂️

web/src/containers/AssurancesAndCompliance.js
```diff
@@ -0,0 +1,125 @@
+import React, { Fragment } from 'react';
+import PropTypes from 'prop-types';
+import { connect } from 'react-redux';
+
+import { t } from '../i18n';
+import { Input } from '../components/Inputs';
+import { Section, Subsection } from '../components/Section';
+import { updateApd as updateApdAction } from '../actions/apd';
+
+import regLinks from '../data/assurancesAndCompliance.yaml';
```
brendansudol: my rule of thumb, documented nowhere of course :), is to alphabetize these by path level and then directory/filename... since these are all on the same level (`../`), it would be:

```
import { updateApd as updateApdAction } from '../actions/apd';
import { Input } from '../components/Inputs';
import { Section, Subsection } from '../components/Section';
import regLinks from '../data/assurancesAndCompliance.yaml';
import { t } from '../i18n';
```

if there were files in the same directory as current file, those would go above, and if there were some in `../../foo/bar` those would go below. whatcha think? feel free to ignore

mgwalker: I like it!  Will reorder them.

web/src/containers/AssurancesAndCompliance.js
```diff
@@ -0,0 +1,125 @@
...
+                  <td colSpan="4">
+                    <h3>{t(`assurancesAndCompliance.headers.${name}`)}</h3>
+                  </td>
+                </tr>
+
+                {apdSections[name].map(
+                  ({ title, checked, explanation }, index) => {
+                    const link = regulations[title];
+                    return (
+                      <tr style={{ height: '4.5em' }}>
```
brendansudol: i can clean up these inline styles l8er

Issue #715, opened 2018-06-26
All activities are key activities


NAretakis:
For some reason all activity sections are showing up as "Key". "Key" activity sections should only show up for the first activity "Program Administration". All subsequent added activities should not have the "Key" nomenclature.

### This task is done when...
 - [ ] "Key" nomenclature is removed from all activities except for the fist one, Program Administration
 - [ ] Data model/backend is updated to note the difference between activity 1 (a Key activity) and all other activities.
 
![image](https://user-images.githubusercontent.com/34281281/41921086-49d612be-791f-11e8-816c-ea5ae4fe1f33.png)

PR #717, opened 2018-06-26
removes "Key" from activity sections that are not key

brendansudol:
fixes: https://github.com/18F/cms-hitech-apd/issues/715

i don't like how i implemented this @mgwalker -- all ears on a better approach...


web/src/util/index.js
```diff
@@ -161,3 +161,6 @@ export const arrToObj = (array = [], initialValue = 0) =>
 export const addObjVals = obj => Object.values(obj).reduce((a, b) => a + b, 0);
 
 export const titleCase = str => str.replace(/\b\S/g, t => t.toUpperCase());
+
+export const isProgamAdmin = activity =>
+  activity.name === 'Program Administration' || activity.id === 1;
```
mgwalker: I'm wondering if `isKey` should be a property on the activity state object.  `key` could be defined as the activity with the smallest ID (since the IDs will change over time because of the way we're kind of abusing the database 😬).  Then you could pull it off the activity state in all those components instead of asking a helper?

Got me further wondering if `isKey` should be in the data model as well.  At some point we probably ought to change the API behavior to not delete and re-add everything every time, but once we do that, we can't guarantee ordering anymore (and we should probably address that, too, but I've got an idea about that).

Anyway...

NAretakis:
We'll probably need some helptext to better explain how the first activity should capture Key information as compared to any subsequent activity. But this is another issue and a different PR. @quarterback for visibility.

web/src/i18n/locales/en.yaml
```diff
@@ -133,7 +133,7 @@
     namePrefix: "Activity"
     namePrefixAndNum: "Core Activity {{number}}"
     description: 
-      title: "Key Program Administration Information"
+      title: "Activity Overview"
```
NAretakis: I changed this to Activity Overview because that is a better name for this. Does this break anything that is doing "key" stuff? @brendansudol

quarterback: Good call @NAretakis

mgwalker:
Change looks good to me @NAretakis!

PR #725, opened 2018-06-27
Initiate APD with state profile from state model, update state model when APD is saved with state profile

### LINKED ISSUE #700: Save APD-specific state profile data
Because state profile data can change over time, we need to save the "current" profile data as part of the APD in addition to attaching it to the state itself.

Question for @jeromeleecms @NAretakis: We have a state data model that also includes identical profile information.  When we save an APD for a state, in addition to saving the state profile in the APD model, should we also update the state data model?

My thinking is that once the state data model is set, new APDs can just inherit profile info from there.  As long as the info doesn't change, they don't have to input anything.  But if the profile info does change, we need to save it for that APD model.  If we also save it for the state model, then future APDs will inherit the updated info.

Does that make sense?

### LINKED ISSUE #701: If an APD does not have state profile data in it yet, pre-populate it with the data on record for that state
When an APD is created, we should grab the state profile info from the state's data model and fill that in for the APD.  That way if the state profile hasn't changed, users can just use the existing info and not have to enter anything.

mgwalker:
Modifies the API behavior in two ways:
- when a new APD is created, if the state already has a profile configured, the profile is copied into the APD
- when an APD is saved, if it has state profile information, the state model's profile info is updated with the info from the APD

Closes #700
Closes #701

- new APDs will have state profile information, if any is defined

### DIFF
--- a/api/db/apd.js
+++ b/api/db/apd.js
@@ -104,8 +104,6 @@ module.exports = () => ({
 
     static: {
       updateableFields: [
-        'status',
-        'period',
         'previousActivitySummary',
         'programOverview',
         'narrativeHIE',
--- a/api/db/state.js
+++ b/api/db/state.js
@@ -2,6 +2,17 @@ module.exports = () => ({
   state: {
     tableName: 'states',
 
+    format(attributes) {
+      const out = { ...attributes };
+
+      // Stringify.  If undefined, set that instead.  These
+      // must be set to something or else bookshelf will throw
+      out.medicaid_office = JSON.stringify(out.medicaid_office);
+      out.state_pocs = JSON.stringify(out.state_pocs);
+
+      return out;
+    },
+
     apds() {
       return this.hasMany('apd');
     }
--- a/api/middleware/synchronize.js
+++ b/api/middleware/synchronize.js
@@ -54,6 +54,11 @@ const synchronizeAll = getSpecifics =>
   );
 
 /**
+ * @callback SyncEventHandler
+ * @param {object} request The request object.
+ */
+
+/**
  * @description Middleware to synchronize a single model instance
  * based on incoming request data
  * @param {function} getModelToSync Function that returns the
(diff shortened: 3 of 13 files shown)


api/db/apd.js
```diff
@@ -104,8 +104,6 @@ module.exports = () => ({
 
     static: {
       updateableFields: [
-        'status',
-        'period',
```
mgwalker: I don't even know what `period` refers to, but `status` shouldn't be settable directly.  It should be inferred by other actions (e.g., submitting an APD, sending it back to a state for responses, etc.).

api/db/state.js
```diff
@@ -2,6 +2,17 @@ module.exports = () => ({
...
+    format(attributes) {
+      const out = { ...attributes };
+
+      // Stringify.  If undefined, set that instead.  These
+      // must be set to something or else bookshelf will throw
+      out.medicaid_office = JSON.stringify(out.medicaid_office);
+      out.state_pocs = JSON.stringify(out.state_pocs);
+
+      return out;
+    },
```
mgwalker: This stringification was being done in the state route handler, but that's not very reusable.  Since we now update a state in the APD route handler too, I figured this should be more centralized.

api/middleware/synchronize.js
```diff
@@ -65,13 +70,30 @@ const synchronizeAll = getSpecifics =>
...
+ *    a promise.  If an event handler throws, the request will
+ *    terminate with a 500 server error unless the thrown error
+ *    has a statusCode property.
+ * @param {SyncEventHandler} events.afterSync Called after the model has
+ *    been synced.
+ * @param {SyncEventHandler} events.beforeSync Called before the model is
+ *    synced.  A good opportunity to massage the request object.
  */
-const synchronizeSpecific = getModelToSync =>
+const synchronizeSpecific = (getModelToSync, { afterSync, beforeSync } = {}) =>
```
mgwalker: Add some lifecycle-type events.  We need this for the APD `put` route in order to update the state model before the response is sent.  I think ideally the callbacks would also happen inside the db transaction, but...  I dunno.  This seems okay?

api/routes/apds/post.js
```diff
@@ -12,6 +19,12 @@ module.exports = (app, ApdModel = defaultApdModel) => {
         state_id: req.user.state,
         status: 'draft'
       });
+
+      const state = await StateModel.where({ id: req.user.state }).fetch();
+      if (state.get('medicaid_office')) {
+        newApd.set({ stateProfile: state.get('medicaid_office') });
+      }
```
mgwalker: Using the `medicaid_office` property on the state model as the full state profile for now.  Probably ought to clean up the model, but that touches a lot more places so I left it alone for now.

api/db/apd.js
brendansudol: i think period was originally intended to encompass the year range for an APD (i.e., 2018-2020)

api/middleware/synchronize.js
brendansudol: nice, seems legit to me

PR #728, opened 2018-06-27
table (start) and touchups

brendansudol:
continuation of design refinement; this one has some misc touchups and tweaks, and starts the work on upgrading our table styles.

there is still more to do and refine with our tables (primarily supporting multiple colors), but they're already looking a lot better so it felt like a good breakpoint to get stuff merged.

preview:
image

### DIFF
--- a/web/src/components/HelpText.js
+++ b/web/src/components/HelpText.js
@@ -9,7 +9,7 @@ const HelpText = ({ text: textResource, reminder: reminderResource }) => {
 
   return (
     (text || reminder) && (
-      <div className="mb2" style={{ whiteSpace: 'pre-line' }}>
+      <div className="mb2">
         {text && <div>{text}</div>}
         {reminder && <div className="red">{reminder}</div>}
       </div>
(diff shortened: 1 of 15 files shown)


NAretakis:
Looking good. Something to think about is perhaps only showing the MMIS table, if an MMIS activity was selected. Otherwise, the default might just be the HIT/HIE table. I can put that concept in our refinements bucket.

web/src/components/HelpText.js
```diff
@@ -9,7 +9,7 @@ const HelpText = ({ text: textResource, reminder: reminderResource }) => {
 
   return (
     (text || reminder) && (
-      <div className="mb2" style={{ whiteSpace: 'pre-line' }}>
+      <div className="mb2">
```
mgwalker: Should this have the `pre-line` class?  I *think* this was to support help text that had newline characters but I can't remember exactly.

web/src/containers/AssurancesAndCompliance.js
```diff
@@ -43,7 +43,7 @@ const AssurancesAndCompliance = ({ sections: apdSections, updateApd }) => {
                   ({ title, checked, explanation }, index) => {
                     const link = regulations[title];
                     return (
-                      <tr style={{ height: '4.5em' }}>
+                      <tr key={title} style={{ height: '4.5em' }}>
```
mgwalker: 👍

Issue #747, opened 2018-06-29
Add placeholder text and update help text for Program Introduction


lauraponce:
There was a tendency to paste in a chunk of information in the Program Introduction – would be useful to let people know they will have the opportunity to add in more detail in the following sections.

Should also clarify how the Program Overview is different than (and similar to) the executive summary.

Issue #753, opened 2018-06-29
Language around key personnel percentages needs clarification


lauraponce:
From round 5 of research, there was some confusion about the %FTE box. As per our research share out, Nick mentioned that this could be referred to as "%FTE attributable to this activity"

quarterback:
Made this change last week and it's already implemented.

Issue #749, opened 2018-06-29
Add placeholder text and update help text for Brief Activity Summary


lauraponce:
There was a tendency to paste in a chunk of information in the Brief Activity Summary – need to let people know they will have the opportunity to add in more detail in the following sections. Placeholder text could also help here.

Issue #748, opened 2018-06-29
Rework 280 character limit phrasing in Program Activities


lauraponce:
280 character limit prompted questions and created anxiety – need to re-work the phrasing of that character limitation

NAretakis:
We changed the helptext a bit, but the section is still limited to 280 characters. This still might create some anxiety, but I think we can help by explaining that the brief summary is supposed to be very short because we are going to use the Brief Summary in other section. While the activity details section can be  big as states need it to be.

quarterback:
I'll re-work this again beyond what I changed last week on the fly during sprint planning.

PR #763, opened 2018-07-03
Content updates

quarterback:
Closes issues #747, #748 #749, #742, #753


NAretakis:
Looks good, but I am not seeing program intro helptext #747 , and I am not sure if there is anything to look at for #742, is there?

PR #767, opened 2018-07-05
Add points-of-contact

### LINKED ISSUE #702: Let users set points-of-contact for the APD
Part of the old state profile included points of contact.  To keep things moving along, we left points of contact out of #679, but presumably we should add it back.

Question for @jeromeleecms @NAretakis: Should points-of-contact be treated the same as the other profile information, and stored in both the APD data model as well as the state data model? (More detail in #700 and #701.)  Or are points-of-contact more specific to each APD?

### LINKED ISSUE #769: New APDs don't load because federal citations property is empty
When we create a new APD, the federal citations is empty.  The frontend expects it to be non-empty so it fails.

### This task is done when...
 - [ ] frontend survives empty federal citations

Longer term, @brendansudol: we should spend a few minutes figuring out what a new APD should look like instead of creating a completely empty one.

mgwalker:
Adds UI for collecting point of contact information as well as the server-side piece.  All wired up in one nice, nifty PR.

### TODO
* when creating an APD, initialize its points of contact with the data from the state
* when saving an APD, update the state's points of contact with the data from the APD

Closes #702 
Closes #769

- adds a UI for collecting point of contact info
- saves/restores points of contact for an APD


lauraponce:
This is minor, but what do you think about updating the button copy to "Add another contact" so that it's clearer that it's _another_ and you don't have to click that to "add" the first one?

PR #778, opened 2018-07-06
Adds MMIS previous activity table

### LINKED ISSUE #652: Build out Results of Previous Activities section

mgwalker:
Just what it says on the tin!  Updates the data model and everything accordingly.

Closes #652 

- shows the MMIS previous activity table


web/src/containers/ApdPreviousActivityTableMMIS.js
```diff
@@ -0,0 +1,144 @@
...
+
+const mapStateToProps = ({ apd: { data: { previousActivityExpenses } } }) => ({
+  previousActivityExpenses
+});
+
+const mapDispatchToProps = { updateApd };
+
+export default connect(mapStateToProps, mapDispatchToProps)(
+  ApdPreviousActivityTable
+);
```
mgwalker: This should probably be abstracted in such a way we can use one component for both tables (I don't think it should be TOO hard), but for the sake of just getting it knocked out, I duplicated it and removed the unnecessary bits.

web/src/containers/ApdPreviousActivityTableMMIS.js
```diff
@@ -0,0 +1,144 @@
+import PropTypes from 'prop-types';
+import React from 'react';
+import { connect } from 'react-redux';
+
+import { formatMoney } from '../util/formats';
+import { DollarInput } from '../components/Inputs';
+import { updateApd } from '../actions/apd';
+import { t } from '../i18n';
+
+const i18nBase = { scope: 'previousActivities.actualExpenses.table' };
```
brendansudol: neat!

web/src/containers/ApdPreviousActivityTableMMIS.js
brendansudol: yeah, makes sense, can you make an issue to revisit / potentially consolidate?

web/src/containers/ApdPreviousActivityTableMMIS.js
```diff
@@ -0,0 +1,144 @@
...
+    approved: previous.mmis.federalApproved + previous.mmis.stateApproved
+  };
+  return total;
+};
+
+const ApdPreviousActivityTable = ({
+  previousActivityExpenses,
+  updateApd: dispatchUpdateApd
+}) => {
+  const years = Object.keys(previousActivityExpenses);
```
brendansudol: one area for future refinement: moving this to `mapStateToProps` and incorporating the `const total = rollup(previousActivityExpenses[year])` logic there so that the render method gets simpler

jeromeleecms:
Can we add the additional FFP rates to the table (i.e. 75/25 and 50/50)? Here's a quick hack/mockup:

image

jeromeleecms:
@mgwalker Nice! Does the default values load to $0 like the HIT/HIE table above? If not, will it cause problems when making creating new APD?

mgwalker:
@jeromeleecms It does, yes, but if you already have a saved APD, it'll use the saved values, which are all `null`.  :)

Issue #765, opened 2018-07-03
Add in all of the nav items into the left sidebar


lauraponce:
Navigation items that should go in the left nav:

- **State Profile Contact Information**
- **Program Summary**
-- Program Overview
- **Results of Previous Activities**
-- Prior Activities Outline
-- Actual Expenses
- **Program Activities**
-- Activity List
-- Activity 1
-- Activity 2 etc etc
- **Activity Schedule Table**
-- Summary Schedule Table
- **Proposed Budget**
-- Summary Budget Table
-- Federal Share by FFY Quarter
-- Incentive Payments by FFY Quarter
- **Assurances and Compliance**
-- Federal Citations and Justification
- **Executive Summary**
-- Executive Summary Overview
-- Program Budget Table
- **Certify and Submit**
-- Submit APD

brendansudol:
done via https://github.com/18F/cms-hitech-apd/pull/779

PR #779, opened 2018-07-06
Add all nav items in the sidebar

mgwalker:
Adds the missing major sections to the sidebar and adds subsection links (per #765).

Addresses #765.

One big ugly thing:  if the display height is less than the navbar height, the "save apd" button gets stuck on top of the navbar links.  It's not clear to me how it's positioned (absolutely inside a rel that's pinned by vertical height?) - should we fix that now or in a separate PR?  Maybe we should move that save button into a sticky header or something, but eventually the goal is to automatically save the APD regularly too, right?

- makes the nav bar full of links


jeromeleecms:
Couple items + some styling questions/comments (happy to move that to another issue if out of scope for this one):

- The second "Activity Schedule Table" doesn't match the current header "Summary Schedule Table". Can we rename the Nav Bar option to "Summary Schedule Table"?

image

image

- Executive Summary Overview: Doesn't line up with the header for the section. Worth renaming the header to line up with the Nav?

image

image

- What do you think about having an "on hover" event to highlight the link you are on? If can mirror what happens when you select the link (with the left border) or just highlight in white bold so the user can track where they are:

image

- Also, is the indent between list items seems really shallow (i.e. one character) - does it make it visually hard to distinguish between items?

- For the "Save APD" Button -- for now can you just stick it next the "Save to PDF" button?

image

- Autosaving -- will document that with our new "Future Enhancement" process. Will probably require some discussion if/how to implement?

lauraponce:
Is it possible to keep the subsections collapsed by default and open the one they are currently on?

mgwalker:
@lauraponce That's doable.  @jeromeleecms would it be okay to put Laura's suggestion into a separate PR?  It's not hard but the amount of code change would make this harder to review/reason about.

jeromeleecms:
@lauraponce Can you spin up an issue with the details of how you want it to behave?

lauraponce:
Created https://github.com/18F/cms-hitech-apd/issues/781 for the nav interaction

jeromeleecms:
Dev magic! :1st_place_medal: 

The activity list names don't seem to be populating? Tried it with a new and existing APD.

image

mgwalker:
Uhh...  oops.  Got a little carried away cleaning out text.  😬

jeromeleecms:
Last bit to fix the indentation:

image

But I'll trust that you fix it and give it the :+1:

web/src/components/SidebarLink.js
```diff
@@ -1,27 +1,42 @@
...
-  </li>
+      <a
+        href={`#${anchor || '!'}`}
+        className="inline-block white text-decoration-none truncate"
+        {...rest}
+      >
+        {children}
+      </a>
+    </li>
+    {sub.map(l => (
```
brendansudol: neat

PR #790, opened 2018-07-10
Collapsible sidebar links

### LINKED ISSUE #781: Collapse subsections by default
When you first load the app, the subsections should all be collapsed by default. The active section/subsection should be expanded with the yellow bar indicator next to it.

Each section should have the carrot next to it indicating there is more in that section.

e.g. if Activity 1 was open, this is what it would look like:

screen shot 2018-07-06 at 10 10 33 am

mgwalker:
Starts the sidebar with major sections collapsed and makes them expandable.

@lauraponce When you click on a section in the sidebar, should that section expand as well as navigate the page?  (from https://github.com/18F/cms-hitech-apd/issues/781#issuecomment-403986690)

Closes #781

### Still needs:
- reducer and action tests

- the major sections in the sidebar are collapsed by default
- clicking the up/down arrows will expand/collapse the link section without navigating the page

### DIFF
--- a/web/src/components/SidebarLink.js
+++ b/web/src/components/SidebarLink.js
@@ -1,9 +1,36 @@
 import PropTypes from 'prop-types';
 import React, { Fragment } from 'react';
+import Icon from '@fortawesome/react-fontawesome';
+
+import { faChevronDown, faChevronUp } from './Icons';
 
 const addIf = (condit, yes, no = '') => (condit ? yes : no);
 
-const SidebarLink = ({ anchor, children, depth, hash, sub, ...rest }) => (
+const icon = (expanded, handler) => {
+  if (expanded === true || expanded === false) {
+    return (
+      <button className="btn block right white" onClick={handler}>
+        {expanded ? (
+          <Icon icon={faChevronUp} size="sm" />
+        ) : (
+          <Icon icon={faChevronDown} size="sm" />
+        )}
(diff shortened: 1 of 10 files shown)


web/src/components/SidebarLink.js
```diff
@@ -13,33 +40,39 @@ const SidebarLink = ({ anchor, children, depth, hash, sub, ...rest }) => (
     >
       <a
         href={`#${anchor || '!'}`}
-        className="block white text-decoration-none truncate"
```
mgwalker: Not sure if removing `block` is the right thing to do here.  The reason I did it is that the up/down icon was being floated right and ended up on the next vertical line.

brendansudol: it's probably fine -- it means the hit target is now slightly smaller (doesn't extend the full width), but not a big deal

web/src/components/SidebarLink.js
```diff
@@ -1,9 +1,36 @@
 import PropTypes from 'prop-types';
 import React, { Fragment } from 'react';
+import Icon from '@fortawesome/react-fontawesome';
+
+import { faChevronDown, faChevronUp } from './Icons';
```
brendansudol: not necessary, but can import the `Icon` component directly from `./Icons` (as the default) and save a line

PR #791, opened 2018-07-11
no table data UI

brendansudol:
adds a tasteful 'no rows/data' message if you remove all the entries from the state personnel / contractor / expenses sections (instead of showing an ugly table with just columns and no rows)

preview:
image


brendansudol:
dang, that lines changed stat is so misleading...

Issue #738, opened 2018-06-28
Add "+" content into the summary budget table section


NAretakis:
The summary budget table has "+" signs for each line item. When clicked, users should see a row for each state, contract, or non-personnel entry that is being summed. 

### This task is done when...
 - [ ] Designs are wireframed to display detailed information when the "+" is clicked.

lauraponce:
@brendansudol do you need a wireframe for this? I'm picturing the text would just show up where you have the "Details will go here" – nothing fancy. Does that work? Let me know if you need something mocked up though and I can do that.

![screen shot 2018-07-09 at 10 42 02 am](https://user-images.githubusercontent.com/21222704/42466781-bc5fccea-8364-11e8-8904-cf9b4ee61c4e.png)

cc @jeromeleecms

brendansudol:
i'm good to iterate on this without a mock -- my one question is around what level of detail we want to show for the individual entries that make up these rolled-up category. i.e., do we need to show all of the numbers that we're showing at the category level? can/should we simplify that and just show entry name and totals? do we even need this detail section at all?

cc: @jeromeleecms @NAretakis

jeromeleecms:
I remember hearing one of the state testers mentioning this functionality, but I thought the initial intention of the + was to show any cost allocation (Reference #404 / #540). 

If we go this route (and not go the route of cost allocation), the "+" sign can show the total cost by Activity (i.e. Activity 1 - FFY 2018: $100,000 / Activity 2 - FFY 2018: $100,000). I do like the idea of clicking "+" under contracts and seeing all the contracts associated with the APD.

Do we want to do a quick discussion on this?

brendansudol:
if we go the cost allocation route @jeromeleecms, would the only entries with the '+' be mmis where there's an option for diff allocations besides the standard 90/10?

re: total cost by activity by year, i'm actually not convinced we need this because it's the exact same info as what's in the exec summary.

my latest thinking is that we should simply remove the '+' and see if it's missed.

jeromeleecms:
1. Cost Allocation: Yes, MMIS would be the major use case for this option. 
2. Total Cost by Activity By Year: The only difference would be that this wouldn't be an entire summary of the activity, but that piece of it. For example: Under personnel costs category, clicking the plus button would show the activity costs for personnel costs only. Under the contract costs, clicking the plus button would show the list of contracts for the activity. 
[...]

Issue #404, opened 2018-04-09
Wireframe v1 of the budget section for the state submission


lauraponce:
### This task is done when...
 - [x] wireframe has been designed
 - [x] wireframe has been shared with the team

lauraponce:
![screen shot 2018-04-23 at 9 16 05 am](https://user-images.githubusercontent.com/21222704/39128896-0479e10c-46d7-11e8-87e7-1a430871240b.png)

jeromeleecms:
@lauraponce This is for the final summary proposed budget section, right? I like the feel of this -- some initial thoughts:

1) What would this look like if we had 50% or 75% FFP match (MMIS may have this now) + it would be helpful for scaling this concept if we took this to MMIS/E&E in the future and not have to rework budget tables in a future iteration of the product. For the MVP, most MMIS requests will be 90%, so if time is an issue, open to tabling this.

2) If you scroll/hover all the way to the right, would there be a total column (i.e. the totals from FFY 2017+FFY 2018 in this example)? Could the FFYs in the "Two-Year Budget"  be clickable and it would move to that section of the table instead of relying on a hover behavior (thinking from a 508 standpoint, would this be compliant or could we replicate that hover with a right/left arrow on the keyboard)? Alternatively, could the initial view of this table be just the total computable for the FFYs + the total of all FFY columns and when you click on the table header, it would expand the FFY to show the Federal and State share? Not sure what makes sense quite yet, but just throwing that out there as food for thought.

3) Small thing - Two-year budget should read FFY 2017 - FFY 2018 (for coherency in the wireframe)

4) Instead of the HIT administrative budget, would this be the sum of all HIT activities? If a user selects HIT/HIE/MMIS in the activities listing at the beginning of the app, how would that translate into this table? Maybe this is less of a design issue as opposed to understanding how the business/logic components of the app trickle down.

Happy to talk through this more, if needed!

lauraponce:
Here is an updated version based on our conversation! The changes in v2 are:

1. For the v2 of the table, if you were to hover over the second cell in FFY 2017, it would pop out like that with the info icon to expand with the additional info. 
2. Added asterisk in title text to indicate it's 90/10
3. Added viewing selection at the top to select which sections they want to see, if they only selected two, one of them would collapse
4. Added scroll bar to see what that would look like 
5. Added color to indicate groupings
6. Added way to collapse and expand rows
7. Added shadows to make it more obvious that the left column would lock in place

[...]

PR #795, opened 2018-07-11
budget summary row details

brendansudol:
addresses and closes https://github.com/18F/cms-hitech-apd/issues/738

Adds totals by activity (by year) when you click the `+` in budget summary table.

The style of this list right now is quite basic, so we may want to revamp soon)

preview:
image

### DIFF
--- a/web/src/actions/apd.js
+++ b/web/src/actions/apd.js
@@ -80,7 +80,6 @@ export const fetchApd = () => dispatch => {
     .then(req => {
       const apd = Array.isArray(req.data) ? req.data : null;
       dispatch(receiveApd(apd));
-      dispatch(updateBudget());
     })
     .catch(error => {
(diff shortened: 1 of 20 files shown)


jeromeleecms:
Is it too much scope for this issue to add the activity listing by Total/Federal/State as a collapsible table row vs a list? In that format, you can remove the "All years" since it will roll-up (sort of modeled in #404). Does that make sense?

brendansudol:
ah, i thought we were simplifying for that table in #404 to just be the totals per activity. do you think it's important to also have the federal and state amounts per activity, @jeromeleecms?

jeromeleecms:
@brendansudol I thought it would be easier for the users to see how the totals rolled up between Federal/State, but we can put it in as and have some states/territories look at it in Denver. If they have a different expectation, we can create a new issue. Does that work for you?

PR #800, opened 2018-07-12
Set federal share percentage per activity, per FFY quarter

### LINKED ISSUE #756: Add in a way to specify which quarters activity is funded in Cost Allocation
Users should be able to specify when/which quarter(s) the activity will be funded while entering the activity details in Cost Allocation

mgwalker:
Closes #756!

- removes the sliders from the federal share by FFY quarter summary table
- adds a new table to the activity cost allocation section
  - this table lets the user set the federal share percentage for each quarter of each FFY, for each expense type, and live updates the dollar amounts
- updates the summary table to take into account the percentages applied to each activity instead of applying a global percentage

### Needs...
- content (there's a rogue `moop moop` floating around the new cost allocation table because I wasn't sure what it should be) - pinging @jeromeleecms and @quarterback for that one
- budget calculation tests

### DIFF
--- a/web/src/actions/apd.js
+++ b/web/src/actions/apd.js
@@ -80,7 +80,6 @@ export const fetchApd = () => dispatch => {
     .then(req => {
       const apd = Array.isArray(req.data) ? req.data : null;
       dispatch(receiveApd(apd));
-      dispatch(updateBudget());
     })
     .catch(error => {
       const reason = error.response ? error.response.data : 'N/A';
--- a/web/src/containers/ActivityDetailCostAllocateFFP.test.js
+++ b/web/src/containers/ActivityDetailCostAllocateFFP.test.js
@@ -52,8 +52,8 @@ describe('the activities cost allocation FFP component', () => {
           statePersonnel: [
             {
               years: {
-                '1066': { amt: 100 },
-                '1067': { amt: 1000 }
+                '1066': { amt: 100, perc: 41 },
+                '1067': { amt: 1000, perc: 63 }
               }
             }
           ],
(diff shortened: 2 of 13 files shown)


lauraponce:
Looking good! It looks like the labels of the left column got a little jumbled –– should be "Project state staff + non-personnel" and "Contracted resources @mgwalker

jeromeleecms:
image

The text does some weird overruns for numbers >= 6 digits.

mgwalker:
The `Federal-State split` calculation wasn't accounting for %FTE on state staff.  Whoops!

quarterback:
I added some text for the cost allocation section and replaced the errant moops

web/src/containers/ActivityQuarterlyBudgetSummary.js
```diff
@@ -0,0 +1,197 @@
...
+                        <Fragment key={year}>
+                          {QUARTERS.map(q => (
+                            <td
+                              className={`mono right-align ${
+                                name === 'combined' ? `${color(i)}-light` : ''
+                              }`}
+                              key={q}
+                            >
+                              <PercentInput
+                                name={`ffp-ACTIVITYIDHERE-${year}-${q}-${name}`}
```
brendansudol: update with actual `ACTIVITYIDHERE`?

brendansudol: oops, sorry, i just saw the `WIP` in title, i will wait :)

web/src/actions/apd.js
```diff
@@ -80,7 +80,6 @@ export const fetchApd = () => dispatch => {
     .then(req => {
       const apd = Array.isArray(req.data) ? req.data : null;
       dispatch(receiveApd(apd));
-      dispatch(updateBudget());
```
mgwalker: Don't need to update the budget when APDs are *fetched*, but only when one is selected.

web/src/containers/ActivityDetailCostAllocateFFP.test.js
```diff
@@ -52,8 +52,8 @@ describe('the activities cost allocation FFP component', () => {
           statePersonnel: [
             {
               years: {
-                '1066': { amt: 100 },
-                '1067': { amt: 1000 }
```
mgwalker: Tweak the test data to exercise the allocation math.  Lots of update required as a result of this, but higher confidence that the math is right.

web/webpack.config.js
```diff
@@ -14,6 +14,7 @@ const config = {
     path: path.join(__dirname, 'dist'),
     filename: 'app.js'
   },
+  // devtool: 'cheap-module-eval-source-map',
```
mgwalker: Useful in dev for enabling breakpoints and debugging in Chrome (and VSCode, and probably Atom too!), but breaks the build.  Probably ought to configure it to only enable in dev builds, but it's not instantly clear to me how to do that so it can wait.

brendansudol: 👍

PR #806, opened 2018-07-16
fix rich editor copy and paste

brendansudol:
Now, we can copy and paste in sentences with bold and italics and yes, even lists, from Word or Google Docs into our rich text editors and it should now work. 🤞 

This was tricky to track down because the library we were using (https://github.com/jpuri/react-draft-wysiwyg) allows for this functionality and it works great on their demo page but it wasn't working for us. The reason: we were using the latest version of this package and they recently made some changes to fix other things that in turn broke this copy/pasting (and their demo page kept working fine because it never upgraded to the latest version). So all I had to do was revert to the version that's being used on their demo page and all is now good :)

Preview:
https://cl.ly/t0Lj

### DIFF
--- a/web/package-lock.json
+++ b/web/package-lock.json
@@ -4518,14 +4518,14 @@
       }
     },
     "draftjs-to-html": {
-      "version": "0.8.3",
-      "resolved": "https://registry.npmjs.org/draftjs-to-html/-/draftjs-to-html-0.8.3.tgz",
-      "integrity": "sha1-DKPCc+gC7Ts7vxLjkA67cP2VUX8="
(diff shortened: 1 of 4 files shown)


jeromeleecms:
Copy and Paste from Word with Bullets:
image

Am I getting greedy if I add tables to our wishlist for copy and paste functionality? For MVP, this should cover most everything (images appear to work as well).

jeromeleecms:
Makes sense. Let’s keep moving and take it as a future request if it comes up during user testing.

brendansudol:
Woohoo! that copy/paste test from Word scored a 💯 !

Issue #802, opened 2018-07-13
Refreshing the browser on APD page...


jeromeleecms:
Per @mgwalker "We need an issue to decide how to handle that - basically what's happening is the web app doesn't know which APD to load, so it doesn't load anything.  (Now that we can get proper data from the API, we've taken a lot of the dummy data out of the web app, so it can't just carry on.)"

Current suggestion is to use the Web Storage API (https://developer.mozilla.org/en-US/docs/Web/API/Web_Storage_API)

### This task is done when...
 - [ ] Decision is made on how to reload the current APD
 - [ ] Dev has implemented the decision

PR #817, opened 2018-07-17
fix apd blank screen

brendansudol:
fixes: https://github.com/18F/cms-hitech-apd/issues/802

if you land on `/apd` page, it redirects you to homepage so that you can select an apd of your choosing.


jeromeleecms:
Should it take you to the main dashboard or can it keep you on the APD that you are currently working on?

mgwalker:
Maybe we should leave #802 open.  This is just to eliminate the blank screen - the app will do *something*.  Getting the local storage stuff - so you can reload and be looking at the same APD - might take a little more time to do right (there are some issues with timing that might be hard to figure out).

Does this seem like a reasonable stopgap?

brendansudol:
right now it does the former @jeromeleecms; if we get annoyed by that and want to move to latter, we can certainly do that in time (it's a little trickier)

jeromeleecms:
Works for me! Let’s keep #802 open, though. At least the app does stop working now.

PR #807, opened 2018-07-16
Save activity quarterly FFP in the API

mgwalker:
Adds the backend pieces and tweaks the frontend data state pieces so we can save and retrieve activity quarterly FFP data.

### This pull request changes...
- when you push the "save APD" button, your activity quarterly FFP data gets saved too!
- when you reload the page, your activity quarterly FFP data is restored!

### This pull request is ready to merge when...
- [ ] Tests have been updated (and all tests are passing)
- [ ] This code has been reviewed by someone other than the original author
- [x] The change has been documented
  - [x] Associated OpenAPI documentation has been updated

PR #818, opened 2018-07-17
Change the way we identify activities and activity components within the app [skip deploy]

mgwalker:
The API is responsible for creating unique IDs for activities and activities components (goals, personnel, etc.).  The app was previously using those IDs if they were available or else creating temporary IDs to uniquely identify activities.  The result is that we have to treat IDs as app-generated because we don't know whether the API created them or not, so we can't make any assertions about their "truth."

Further, that means when we save an APD, we can't send any activity or component IDs along.  The practical result is that the API deletes all of an APD's activities from the database and then inserts new ones - every time the APD is saved.  That's slow and dirty.

This PR updates the way we handle activity and component identity within the app by assigning each one a unique `key` instead.  Then, if an activity has an ID, we can be sure that it came from the API.  Now when we save an APD, we also send activity and component IDs, meaning that they don't first get deleted and then inserted - they are just updated.

### Known limitation
If you create a new activity in the app and then save the APD, the app state is not updated with the activity's ID, so until you reload the page, every time you save the APD, that activity (and its components) will be deleted and reinserted.  Since the API sends back the whole APD object after a save, we should probably completely resync application state at that point (e.g., overwrite the data into the APD's `byId` state, then issue a `SELECT_APD` action).

- saves should be *very slightly* faster
- enables us to save activity quarterly FFP data (see https://github.com/18F/cms-hitech-apd/pull/807#issuecomment-405624158)


web/src/containers/ActivityQuarterlyBudgetSummary.js
```diff
@@ -175,14 +175,14 @@ class ActivityQuarterlyBudgetSummary extends Component {
 }
 
 ActivityQuarterlyBudgetSummary.propTypes = {
-  aId: PropTypes.oneOfType([PropTypes.string, PropTypes.number]).isRequired,
+  aKey: PropTypes.string.isRequired,
```
brendansudol: 🎉

PR #825, opened 2018-07-18
Refactoring: Utilities for serializing/deserializing data exchanged with API

mgwalker:
Our activity reducer and APD action creator both had a lot of logic in them unrelated to their primary responsibilities - they both converted data to/from API/redux format.  This PR pulls that conversion logic into utilities to do the serialization for us.

This should also simplify future work, such as loading a newly-created APD without reloading the page.

- no visible changes


web/src/reducers/activities.js
```diff
@@ -367,137 +368,11 @@ const reducer = (state = initialState, action) => {
...
-                  {}
-                )
-              : { ...arrToObj(action.apd.years, quarterlyFFPEntry()) },
-          meta: {
-            expanded: a.name === 'Program Administration'
-          }
-        };
+        // const activity = fromAPI(a, action.apd.years, {
+        //   quarterlyFFP: { ...arrToObj(action.apd.years, quarterlyFFPEntry()) }
+        // });
```
mgwalker: Don't have to deserialize activities in the reducer at all - they are deserialized along with the rest of the APD.  Now when we select an APD, we just have to copy the activities into the `byKey` map.

web/src/reducers/activities.test.js
```diff
@@ -589,343 +589,24 @@ describe('activities reducer', () => {
...
-                      year: 2019
-                    }
-                  ]
-                }
-              ],
-              years: ['2018', '2019']
+                { key: 'a' },
+                { key: 'b', name: 'Program Administration' },
+                { key: 'c' }
+              ]
```
mgwalker: Because we're just copying activities, that makes this test a lot simpler.  :)

web/src/util/serialization/apd.js
```diff
@@ -0,0 +1,192 @@
...
+ * @param {Object} defaults.hitNarrative - default HIT narrative
+ * @param {Object} defaults.incentivePayments - default value for incentive payments
+ * @param {Object} defaults.mmisNarrative - default MMIS narrative
+ * @param {Object} defaults.overview - default APD overview
+ * @param {Object} defaults.pointsOfContact - default points of contact
+ * @param {Object} defaults.previousActivityExpenses - default value for previousActivityExpenses
+ * @param {Object} defaults.previousActivitySummary - default activity summary
+ * @param {Object} defaults.stateProfile - default state profile
+ * @param {Object} defaults.years - default value for years
+ */
```
mgwalker: Added this JSDoc because it turns out VSCode is really good at reading it and turning it into a decently useable tooltip.

web/src/actions/apd.js
```diff
@@ -114,149 +114,9 @@ export const notifyNetError = (action, error) => {
...
-          q1: applyToNumbers(ffp[1], v => v / 100),
-          q2: applyToNumbers(ffp[2], v => v / 100),
-          q3: applyToNumbers(ffp[3], v => v / 100),
-          q4: applyToNumbers(ffp[4], v => v / 100),
-          year
-        })
-      )
-    });
-  });
+  const apd = toAPI(updatedApd, activities);
```
brendansudol: 😍

web/src/reducers/activities.test.js
brendansudol: 🙌

web/src/util/serialization/activities.js
```diff
@@ -0,0 +1,239 @@
...
+      id: e.id,
+      category: e.category,
+      description: e.desc,
+      entries: Object.keys(e.years).map(year => ({
+        amount: +e.years[year],
+        year
+      }))
+    })),
+    standardsAndConditions: {
+      businessResults: activityState.standardsAndConditions.bizResults,
```
brendansudol: at some point, we should unify all of these names (i.e., businessResults v. bizResults), would be greatly simplify this object

web/src/util/serialization/apd.js
brendansudol: whoa, cool

web/src/reducers/apd.js
```diff
@@ -127,106 +127,28 @@ const reducer = (state = initialState, action) => {
...
-                        ...previous,
-                        [year - 2]: getPreviousActivityExpense()
-                      }),
-                      {}
-                    ),
-              previousActivitySummary: previousActivitySummary || '',
-
-              stateProfile: replaceNulls(stateProfile),
-              years: (years || defaultAPDYears).map(y => `${y}`),
+              ...fromAPI(apd, {
```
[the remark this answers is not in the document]
mgwalker: Yep, will do.  Those are the defaults for any missing values, so probably `defaults` will be good.  😆 

In the future, I'd like for the API to always return a complete APD.  Then we wouldn't need defaults at all.  I've got a branch where I've played around with that.

PR #827, opened 2018-07-20
Switch vulnerability scan to CLI instead of web tool [skip deploy]

mgwalker:
The Snyk web tool ignores `package-lock.json` and instead inspects our dependencies based on the _highest version_ of a library that our `package.json` says is acceptable.  This leads to false negatives.

The Snyk CLI tool scans what is actually installed.  Because we install explicitly from `package-lock.json` in our CircleCI config (and our deploy as well), the CLI tool will scan against the versions of libraries that we actually use.

- no visible changes to the product
- should start seeing some failing tests

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -31,6 +31,20 @@ jobs:
           paths:
             - api/node_modules
 
+  web dependency vulnerability scan:
+    docker:
+      - image: node:8
+    steps:
+      - checkout
+      - restore_cache:
+          key: v1-cms-hitech-apd-{{ checksum "web/package-lock.json" }}
+      - run:
+          name: vulnerability scan
+          working_directory: ~/project/web
+          command: |
+            npx snyk auth $SNYK_TOKEN
+            npx snyk test
+
   web lint:
     docker:
       - image: node:8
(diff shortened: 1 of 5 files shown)


.circleci/config.yml
```diff
@@ -31,6 +31,20 @@ jobs:
...
+      - image: node:8
+    steps:
+      - checkout
+      - restore_cache:
+          key: v1-cms-hitech-apd-{{ checksum "web/package-lock.json" }}
+      - run:
+          name: vulnerability scan
+          working_directory: ~/project/web
+          command: |
+            npx snyk auth $SNYK_TOKEN
```
mgwalker: Right now this token is tied to my account, but the token is set in the CircleCI project settings so we can _*easily*_ switch it to someone else later with no code changes.  😄

bin/deploy.sh
```diff
@@ -32,7 +32,7 @@ rm -rf seeds/test
 rm seeds/development.js
 rm seeds/test.js
 rm seeds/shared/delete-everything.js
-npm ci
+npm ci --only=production
```
mgwalker: When we deploy the API, we should only push the production dependencies.  Our dev dependencies aren't useful and there's no good argument for putting unused code into production!

mgwalker:
🤔 Should deploys be held until vuln scans pass?  That seems like it would make sense...

bin/deploy.sh
brendansudol: nice, agreed

brendansudol: also a good test that our dependencies are in the right place (dev vs. prod) in package.json

PR #832, opened 2018-07-20
form field validation start

brendansudol:
This adds the capability for client-side form field validation, on a per field basis, with validation occurring after interacting with the field (and thereafter on field change, though this is configurable).

Behind the scenes, it utilizes [formik](https://github.com/jaredpalmer/formik) and [yup](https://github.com/jquense/yup), though it takes a bit of finagling to work on a per-field basis and jive with our redux store.

The input component API stays largely the same, though now there's one more prop, `validation`, which expects a `yup` schema instance. In time, we could perhaps store this yup schema for all fields in one central place (which could be shared with server-side) and then lookup the relevant field based on the name key but right now, i'm simply adding it in-line with the component. i.e.,

```
<InputValidate
    name="firstName"
    label='First Name'
    value={name} // i.e., from redux store
    onChange={this.handleChange('firstName')} // i.e., dispatch action to update store
    validation={Yup.string()
        .min(2, 'C’mon, your name is longer than that')
        .required('Name is required.')}
/>
```

Maybe I'm completely overthinking this and maybe we don't need `formik` -- we're definitely not using it in the way intended :) -- but perhaps that's fine (I like the `touched` and `errors` objects that it generates, though we could create these on our own...).

For illustrative purposes, there's one field with validation wired up in the app (so far), the medicaid director name, and you can also play around with this at `/temp/misc` page.

Let me know what you think, and also happy to scrap this approach and go with another if preferred.


brendansudol:
preview:
https://cl.ly/t8LW

web/src/components/temp/FormFieldTest.js
```diff
@@ -0,0 +1,42 @@
...
+import { updateApd as updateApdAction } from '../../actions/apd';
+
+const FormFieldTest = ({ overview, updateApd }) => (
+  <div className="app">
+    <h3>Form Fields w/ validation</h3>
+    <Input
+      name="email"
+      label="Email"
+      type="email"
+      validation={Yup.string()
```
mgwalker: Like it.  We've got `joi` somewhere on the API side but not consistently.  I'm 100% okay with switching.

web/src/components/InputsWithValidation.js
```diff
@@ -0,0 +1,113 @@
...
+
+  Wrapper.defaultProps = {
+    value: null,
+    validation: null
+  };
+
+  return Wrapper;
+};
+
+export const Input = makeValidated(InputRaw);
```
mgwalker: It's a little convoluted, but a) you can't see that complexity from the outside so that's cool, b) it doesn't fire redux actions left and right, and c) the error object that Formik puts together is pretty nice.

brendansudol: well said, i agree on all points!

PR #836, opened 2018-07-23
Fix place display [skip deploy]

mgwalker:
Sends back useful user information from the API when logging in or checking your current authentication status.  Captures that user information in the app and stores it in the user application state.  Then we can do cool stuff with it.

- the state name and graphic displayed in the sidebar will match the state associated with your user account

### DIFF
--- a/api/db/user.js
+++ b/api/db/user.js
@@ -28,7 +28,7 @@ module.exports = (zxcvbn = defaultZxcvbn, bcrypt = defaultBcrypt) => ({
 
     async apds() {
       logger.silly('getting user apds');
-      if (!this.relations.state || !this.relations.state.apds) {
+      if (!this.relations.state || !this.relations.state.relations.apds) {
         logger.silly('user apds are not loaded yet... loading them');
         await this.load('state.apds');
       }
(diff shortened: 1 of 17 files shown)


api/db/user.js
```diff
@@ -28,7 +28,7 @@ module.exports = (zxcvbn = defaultZxcvbn, bcrypt = defaultBcrypt) => ({
 
     async apds() {
       logger.silly('getting user apds');
-      if (!this.relations.state || !this.relations.state.apds) {
+      if (!this.relations.state || !this.relations.state.relations.apds) {
```
mgwalker: This has always been wrong.  Thankfully we never used it before.  😛

web/src/reducers/user.js
```diff
@@ -30,6 +37,8 @@ const user = (state = initialState, action) => {
       };
     case GET_USER_FAILURE:
       return { ...state, fetching: false, error: action.error };
+    case LOGOUT_SUCCESS:
+      return initialState;
```
mgwalker: It occurs to me that we should probably reset the state globally on logout.

mgwalker:
No PR deployment for this one because it requires API changes.

web/src/reducers/user.js
brendansudol: yes we should

Issue #522, opened 2018-05-09
Where to store external documents


jeromeleecms:
As part of the App experience, the user may have to attach documents to their submissions. We need to decide how/where the documents will be stored.

### This task is done when...
 - [ ] Decision made on where the attached documents will be stored
 - [ ] Decision has been documented in the Wiki

NAretakis:
Will update this issue with information from the CMS AWS team on file storage.

mgwalker:
Pending a decision on storage, we have some options:

1. We can assume that we'll be storing files in S3 and start building towards that.  We can use a tool like [localstack](https://github.com/localstack/localstack) for dev and testing, but we'll run into problems uploading in staging until we actually have S3 access.  Lowish risk, since S3 seems likely, but if we end up using something other than S3, all this work goes down the tube.

2. We can use a storage abstraction library (something like [abstract-blob-store](https://github.com/maxogden/abstract-blob-store) and its plugins with a big of a configurator on top would work well) to let us quickly switch out where we store files - we could just throw them away for now and easily switch to S3 in staging later with just a config change and restart.  Also lowish risk, since we should be able to adapt to whatever storage mechanism we're given, but we may end up having to take on some of the development/maintenance of the abstraction plugins.

I lean towards the second option personally, but I'm a proponent of reasonable abstraction layers - they keep us from being too tightly tied to any particular service (just like we should be able to switch out our database to MySQL or something else, we could also switch from S3 to another storage system).

NAretakis:
Quick convo with folks, S3 is definitely an option available to us as is NFS. We can dig in more on Thursday, but does this help point you to option 1 or 2?

NAretakis:
OBSOLETE

PR #843, opened 2018-07-24
Generic/pluggable file storage

mgwalker:
This PR adds a fairly light wrapper around [abstract-blob-store](https://github.com/maxogden/abstract-blob-store) to allow us to transparently support multiple different storage options.  Currently only two options are supported: local file and `null`, but adding more should be as simple as adding the plugin and adding two lines of code to use it if configured.

The `null` type writes to `/dev/null`, which just means it gets thrown away.  The `/dev/null` path also supports reading, though there's nothing there so it reads no bytes.  But that's fine, the code doesn't care what gets read, as long as the read streams open and close as expected.  😛

Configuration is through environment variables.  Two new variables are added for this purpose:
* **STORE_TYPE** - This indicates what kind of storage we're using.  It defaults to `null`, but also supports `fs` for local file storage.  If it's an unsupported value, we fall back to `null`.
* **STORE_PATH** - Where within the storage to put our files.  For local storage, this is a disk path.  For S3, it might be a bucket name (configuration for S3 will require more stuff; path might not even be relevant, but we'll see).

A step towards being able to support #522.

- no visible changes


api/store.js
```diff
@@ -0,0 +1,176 @@
...
+
+const isReadableStream = obj => obj instanceof stream.Readable;
+
+// Creates a filestream blob store that reads/writes to
+// /dev/null.  Can't use the fsBlobStore for this because
+// it tries to create /dev/null and...  well, that doesn't
+// fails, as we might reasonably expect.
+const nullBlobStore = () => ({
+  createReadStream: () => fs.createReadStream('/dev/null'),
+  createWriteStream: () => fs.createWriteStream('/dev/null'),
```
mgwalker: Creates a dependency on Linux-y systems for now.  There's almost certainly a better way to do this, but it gets us going for now.  Longterm, we might don't need this at all.  Once we've got access to S3, we could create a staging bucket and a production bucket, and bam, we don't have to just pretend like we're storing files.  😛

api/store.js
```diff
@@ -0,0 +1,176 @@
...
+    });
+  }
+
+  /**
+   * @async
+   * Gets the data from a specified blob.
+   * @param {String} key - Blob identifier
+   * @returns {Buffer} A buffer of the data in the blob.
+   */
+  async get(key) {
```
mgwalker: Not sure this method is actually useful, but it was helpful for testing.  My ***guess*** is that in practice, we'll want to get the read stream and pipe that directly to an HTTP response rather than reading it into memory.

api/store.js
```diff
@@ -0,0 +1,176 @@
...
+        writeStream.end(done);
+        writeStream.on('error', errHandler);
+      }
+    });
+  }
+}
+
+const def = new Store(
+  getBlobStore(process.env.STORE_TYPE || '', process.env.STORE_PATH || '')
+);
```
mgwalker: Made this a class to make testing easier - can inject a mock blob-store!  But also, created a default instance based on config.

api/store.js
```diff
@@ -0,0 +1,176 @@
...
+
+Store.exists = def.exists.bind(def);
+Store.get = def.get.bind(def);
+Store.getReadStream = def.getReadStream.bind(def);
+Store.remove = def.remove.bind(def);
+Store.write = def.write.bind(def);
+
+module.exports = Store;
```
mgwalker: The end goal is that you should be able to use it like this:

```javascript
const store = require('./store');
const myFile = await store.get('the-file-with-this-key');

// This one might be possible too, which would be amazing
await store.write(httpRequest)
```

api/store.js
```diff
@@ -0,0 +1,175 @@
...
+      }
+    });
+  }
+}
+
+const def = new Store(
+  getBlobStore(process.env.STORE_TYPE || '', process.env.STORE_PATH || '')
+);
+
+Store.exists = def.exists.bind(def);
```
brendansudol: i'm not sure i follow this class method binding -- can you walk me through what it accomplishes?

brendansudol: can we not just export both the `def` instance as well as the `Store` class?

mgwalker: The binding is a result of how the methods are being exported.  Without the binding, when they're called, their `this` is set to the `Store` class itself, not the instance, so `this.store` resolves `undefined` and everything breaks.

I'm not fundamentally opposed to two exports.  Do you mean something like this?

```javascript
module.exports = {
  default: def,
  class: Store
};
```

My thinking was just that it's nice and clean from a module consumer's perspective to be able to just import a thing and use it.  Ultimately the whole reason for having a class at all is so I can inject the store for testing, and then this import stuff is just to make the module easy to consume in actual use.  Ideally we'd never use the constructor except for the default and for testing.

I'm totally open to other approaches for this, though!

Issue #840, opened 2018-07-24
contractor resource component refinements


brendansudol:
* being able to delete a previously uploaded attachment
* adding a "document type" for a given attachment (not sure what the list of options is here)

PR #845, opened 2018-07-25
add doc category, remove functionality

brendansudol:
* adds accompanying document type for each uploaded file (right now options are "Contract" or "RFP")
* adds ability to remove an uploaded file

Addresses and closes https://github.com/18F/cms-hitech-apd/issues/840

### DIFF
--- a/web/src/containers/ActivityDetailContractorResources.js
+++ b/web/src/containers/ActivityDetailContractorResources.js
@@ -11,6 +11,7 @@ import {
 import NoDataMsg from '../components/NoDataMsg';
 import { Input, DollarInput, Textarea } from '../components/Inputs';
 import { Subsection } from '../components/Section';
+import Select from '../components/Select';
 import { t } from '../i18n';
 import { isProgamAdmin } from '../util';
 
@@ -22,7 +23,15 @@ Label.propTypes = {
   children: PropTypes.node.isRequired
 };
 
+const DOC_TYPES = ['Contract', 'RFP'];
+
 class ContractorExpenses extends Component {
+  state = { docType: DOC_TYPES[0] };
+
+  updateDocType = e => {
+    this.setState({ docType: e.target.value });
+  };
+
   handleChange = (index, field) => e => {
     const { value } = e.target;
     const { activity, updateActivity } = this.props;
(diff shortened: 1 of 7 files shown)


jeromeleecms:
@brendansudol overall looks nice (aka works as advertised)!  Just a couple updates:
1) Please add "Contract Amendment" to the dropdown? It looks like it's hard coded, but down the road can we think through how to simplify adding new categories of documents? 
2) Can we add a confirmation box when someone selects the delete button?

web/src/containers/ActivityDetailContractorResources.js
```diff
@@ -22,7 +23,15 @@ Label.propTypes = {
   children: PropTypes.node.isRequired
 };
 
+const DOC_TYPES = ['Contract', 'RFP'];
+
 class ContractorExpenses extends Component {
+  state = { docType: DOC_TYPES[0] };
```
mgwalker: What's this?  I've always seen state initialized in the constructor using `this.state`. Is this a way of initializing a class member variable?

web/src/containers/ActivityDetailContractorResources.js
```diff
@@ -41,22 +50,36 @@ class ContractorExpenses extends Component {
...
+    // reset document category if necessary
+    if (docType !== DOC_TYPES[0]) this.setState({ docType: DOC_TYPES[0] });
+  };
+
+  handleFileDelete = (cIdx, fIdx) => () => {
+    const { activity, updateActivity } = this.props;
+    const { files } = activity.contractorResources[cIdx];
+    const updatedFiles = files.filter((_, i) => i !== fIdx);
+
+    const updates = { [cIdx]: { files: updatedFiles } };
```
mgwalker: Out of curiosity, is there a reason to do the filtering here instead of the reducer?  Similarly with adding a file.  In the other places we add/remove list items, we do it in the reducer.

web/src/containers/ActivityDetailContractorResources.js
```diff
@@ -135,27 +158,46 @@ class ContractorExpenses extends Component {
...
+                        label="Document Type"
+                        wrapperClass="col-6 mr2"
+                        value={docType}
+                        onChange={this.updateDocType}
+                      />
                       <Dropzone
                         className="btn btn-primary"
-                        onDrop={this.handleDrop(i)}
+                        onDrop={this.handleFileUpload(i)}
                       >
```
mgwalker: Should we add a `multiple={false}` prop to `Dropzone` to match the handler behavior?

web/src/containers/ActivityDetailContractorResources.js
brendansudol: just another way to initialize state without needing to use constructor :)

web/src/containers/ActivityDetailContractorResources.js
brendansudol: mainly just so we didn't need to add additional actions and reducer code, though you're right that it's now not quite consistent; i guess i just like utilizing the versatile update action (and `updeep`)

web/src/containers/ActivityDetailContractorResources.js
brendansudol: YES!!

brendansudol:
made those updates @jeromeleecms and yes, we can think of a way to make adding additional document categories easier down the line

Issue #735, opened 2018-06-28
Linking attached contracts to backend (Activity & Contract resource)


NAretakis:
Need to define and construct how contract attachments will be linked to the database, activity and/or contract resource.

### This task is done when...
 - [ ] criteria 1
 - [ ] criteria 2

mgwalker:
I'm leaning towards a simple `files` table that isn't *directly* linked to anything.  It would look something like this:

|column|type|
|---|---|
|id|(auto) integer|
|key|string - identifies the file in a way that the storage mechanism understands (e.g., a filename)|
|size|integer - file size in bytes|
|metadata|json - for any additional information we want to store about the file; this could be vary based on what kind of file it is|

Then we'd have several many-to-many mapping tables.  For example: A `contract_files` table that has two columns, `file_id` (foreign key into the `files` table) and `contract_id` (foreign key into the `activities_contractor_resources` table).  A nearly identical table for `activity_files`, maybe even `apd_files`.

Curious for others' thoughts.  @jeromeleecms @brendansudol

jeromeleecms:
So this is structure is essentially a master file listing and it would abstract out file location + meta-data into this table right? Seems that this would enable flexibility in the off case we'd have to have track another set of files. Would you need to have a field that denotes how the file is being stored (blob vs physical location)?

mgwalker:
Yep, that's right.

For your second question, no.  That's a possibility, but in the short term, I think it's smart to just assume all of our files will be stored the same way.  We can modify the database later if that turns out to be incorrect.

mgwalker:
#852 updates the database and data models.  The next step will be adding API endpoints to upload and download files, and then I think this one will be done.

mgwalker:
Closed by #865 and #852

PR #852, opened 2018-07-26
Add generic files database model

mgwalker:
This pull request creates a generic database model for files.

- creates a `files` table that stores basic information about the file
  - `id` is used to identify the file within the database
  - `key` is used to identify the file with the data store
  - `size` is (ostensibly) the size of the file in bytes
  - `metadata` is an extensible JSON object where we can store anything else we want to about the file; e.g., its original name (could also put size in here instead of having a dedicated column for it...  thoughts?)

- creates an `activity_files` mapping table to allow us to attach files to activities; this is a many-to-many relationship, so ostensibly one file could be associated with multiple activities, and one activity can have multiple associated files

- creates an `activity_contractor_files` mapping table, which is also many-to-many

- updates the APD, activity, and contractor models to fetch and return file information

Addresses #735.

- no visible changes


api/migrations/20180726153748_files-and-relations.js
```diff
@@ -0,0 +1,28 @@
+exports.up = async knex => {
+  await knex.schema.createTable('files', table => {
+    table.increments();
+    table.text('key');
+    table.integer('size');
```
mgwalker: Should `size` be a distinct column, or should we wrap it into the metadata?  One nice thing about having it be distinct is that we can be sure it'll exist.  If we put it in metadata, there's no longer a guarantee that it exists because the metadata field is free-form.

brendansudol: i like it distinct

brendansudol:
so is the idea that whenever a file gets added that's related to a contractor, we also add a link to just the activity too? or not necessarily? and then there could be other files that are just related to an activity but not a contractor? just want to make sure i understand

mgwalker:
I'm thinking we'd only link to the thing the contract is explicitly associated with.  So if uploading a contract, it's probably associated with a contractor.  If we need to get an associated activity, we can do that via the contractor.

PR #873, opened 2018-07-30
fix 508 issues

brendansudol:
finally used our cool `tota11y` tool and fixed the flagged issues. there were not nearly as many issues as i expected, which was a fun surprise.

### DIFF
--- a/web/src/components/SidebarLink.js
+++ b/web/src/components/SidebarLink.js
@@ -7,18 +7,17 @@ import { faChevronDown, faChevronUp } from './Icons';
 const addIf = (condit, yes, no = '') => (condit ? yes : no);
 
 const icon = (expanded, handler) => {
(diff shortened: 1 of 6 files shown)


NAretakis:
@brendansudol Does this take care of #877 ?

web/src/components/SidebarLink.js
```diff
@@ -7,18 +7,17 @@ import { faChevronDown, faChevronUp } from './Icons';
...
-      </button>
-    );
-  }
-  return null;
+  if (expanded !== true && expanded !== false) return null;
+
+  return (
+    <button
+      className="btn block right white"
+      aria-label={expanded ? 'Collapse' : 'Expand'}
```
mgwalker: Was going to suggested `aria-expanded` here, but it looks like this `aria-label` already takes care of that.  Nice.

Issue #850, opened 2018-07-26
Make all rich text editing fields resizable


NAretakis:
All of the fields that have the rich text editor need to be able to be expanded by the user. No changes are needed for fields that do not utilize the rich text editor.

### This task is done when...
 - [ ] all rich text editor fields are re-sizable

PR #889, opened 2018-07-31
resizable rich text

brendansudol:
Addresses and closes https://github.com/18F/cms-hitech-apd/issues/850

@NAretakis -- this is maybe the best we can do here... as the user adds more text, the box becomes larger --there's no explicit resize box button, it just happens as the content fills the space -- is that kewl?

preview:
https://cl.ly/tMsw

### DIFF
--- a/web/src/styles/app/editor.css
+++ b/web/src/styles/app/editor.css
@@ -26,8 +26,9 @@
   overflow: inherit;
 }
 
-.DraftEditor-root {
-  height: 400px;
+.DraftEditor-root,
+.public-DraftEditor-content {
+  min-height: 240px;
 }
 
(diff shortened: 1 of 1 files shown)


mgwalker:
We could also set the `resize` property on the editable div:

```css
.public-DraftEditor-content {
    resize: both;
}
```

It gives us more standard behavior in Chrome and Firefox (the draggable icon for vertical and horizontal resizing).  Safari shows the draggable corner but it doesn't actually work.  I don't have access to IE right now but can try it out later - though I'm dubious that i'll respect the `resize` property.  Maybe both?  🤔

NAretakis:
Yes. Auto resizing based on input or drag-gable is fine by me.

Issue #851, opened 2018-07-26
Remove "Key" Activity label from UX, & "Key Person" checkbox


NAretakis:
We currently have the word "Key" included for every component under the first activity (program administration). We need to remove the word "Key" from every item so that the first activity (program administration) looks the same as any newly added activity.

However, we still need a way to flag if a person specified under the State Personnel component is a "Key Person". This should function as a checkbox that sits under the title/name space within the Personnel section. Also, this checkbox needs to exist for all activities, not just the first one, Program Administration.

### This task is done when...
 - [ ] "Key" word is removed from all components of Program Administration
 - [ ] "Key Personnel" checkbox is added to the Personnel section for all activities.
 
![image](https://user-images.githubusercontent.com/34281281/43267187-83ace628-90aa-11e8-9f1d-47196e65a54c.png)

quarterback:
I'll tackle the removal of "key" inside the yaml file.

NAretakis:
Excellent. Probably then just need to write out some helptext on how to identify/use the Key Person checkbox that the Devs can implement. After we get the helptext done, let’s spin up an issue for the Devs to tackle.

quarterback:
@brendansudol @NAretakis If I'm not mistaken, I don't see the "key" sections in the yaml file anymore (besides Key Personnel requirement) There used to be, but we'd removed them at one point and so, it appears the "key" sections that exist now are just hard coded?

quarterback:
Help text: _"If the personnel identified is a key member of the project, select the Key Personnel checkbox beneath their name"_ 

If we need more detailed text than this, than I can go there. It'll depend partially on how the experience is modeled.

PR #899, opened 2018-08-01
"key" personnel changes

brendansudol:
Addresses and closes https://github.com/18F/cms-hitech-apd/issues/851

Preview:
image


NAretakis:
Looks good to me. I gather that the checkbox sets the "is key" value, which is something we'd be able to extrapolate later? As in, would I be able to pull out (identify) all of the persons with Key Personnel checked?

web/src/util/serialization/activities.test.js
```diff
@@ -600,6 +605,7 @@ describe('APD activity serializer', () => {
             id: 'eugene',
             title: 'Boss',
             description: 'friend of the show',
+            keyPersonnel: false,
```
mgwalker: should this be `isKeyPersonnel`?

brendansudol: on the backend right now, this field is called `keyPersonnel`

brendansudol:
yes exactly @NAretakis -- with this change, it is now trivial to pull out all of the key personnel across activities using this checkbox flag.

web/src/util/serialization/activities.test.js
mgwalker: Oh, whoops.  Should have pulled the code to look at this in context.  😬

PR #900, opened 2018-08-01
Automatically open new APDs + some frontend cleanup

### LINKED ISSUE #876: Create a valid APD data structure upon creation
Devs can self assign whoever wants to tackle this.

Created per #854

mgwalker:
When you click the "create an APD" button, it is created and then opened.  Just one click!

Also removes a lot of APD initialization and defaulting code since the API generates a valid APD now.  There's no need for the frontend code to do it.

Closes #876 

- click the "create an APD" button, get taken to your fancy-pants 👖 new APD

### DIFF
--- a/web/src/actions/apd.js
+++ b/web/src/actions/apd.js
@@ -51,16 +51,22 @@ export const updateApd = updates => dispatch => {
   }
 };
 
+export const selectApd = (id, pushRoute = push) => (dispatch, getState) => {
+  dispatch({ type: SELECT_APD, apd: getState().apd.byId[id] });
+  dispatch(updateBudget());
+  dispatch(pushRoute('/apd'));
+};
+
 export const createRequest = () => ({ type: CREATE_APD_REQUEST });
 export const createSuccess = data => ({ type: CREATE_APD_SUCCESS, data });
 export const createFailure = () => ({ type: CREATE_APD_FAILURE });
-export const createApd = () => dispatch => {
+export const createApd = ({ pushRoute = push } = {}) => dispatch => {
   dispatch(createRequest());
   return axios
     .post('/apds')
     .then(req => {
       dispatch(createSuccess(req.data));
-      dispatch(selectApd(req.data.id));
+      dispatch(selectApd(req.data.id, pushRoute));
     })
     .catch(error => {
       const reason = error.response ? error.response.data : 'N/A';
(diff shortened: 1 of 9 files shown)


web/src/actions/apd.js
```diff
@@ -51,15 +51,27 @@ export const updateApd = updates => dispatch => {
...
   dispatch(createRequest());
-  return axios.post('/apds').catch(error => {
-    const reason = error.response ? error.response.data : 'N/A';
-    dispatch(createFailure(reason));
-  });
+  return axios
+    .post('/apds')
+    .then(req => {
+      dispatch(createSuccess(req.data));
+      dispatch(selectApd(req.data.id, pushRoute));
```
mgwalker: It's so cool that we can just...  use what we already have.  ❤️ redux

web/src/actions/apd.test.js
```diff
@@ -90,13 +90,29 @@ describe('apd actions', () => {
...
+      const expectedActions = [
+        { type: actions.CREATE_APD_REQUEST },
+        { type: actions.CREATE_APD_SUCCESS, data: newapd },
+        { type: actions.SELECT_APD, apd: { hello: 'world' } },
+        { type: actions.UPDATE_BUDGET, state },
+        { type: 'FAKE_PUSH', pushRoute: '/apd' }
+      ];
 
-      store.dispatch(actions.createApd()).then(() => {
+      return store.dispatch(actions.createApd({ pushRoute: push })).then(() => {
```
mgwalker: If you don't `return` this promise, the test finishes before the action actually happens, so your test will fail but the testrunner will say everything passed.  Don't ask me how I found out.  😳

web/src/reducers/activities.js
```diff
@@ -399,8 +399,10 @@ const reducer = (state = initialState, action) => {
     case SELECT_APD: {
       const byKey = {};
       ((action.apd || {}).activities || []).forEach(a => {
-        byKey[a.key] = a;
-        byKey[a.key].meta = { expanded: a.name === 'Program Administration' };
+        byKey[a.key] = {
+          ...a,
+          meta: { expanded: a.name === 'Program Administration' }
```
mgwalker: This started throwing an immutability error.  Kind of weird that it wasn't already doing that, in hindsight. 🤷‍♂️

web/src/util/serialization/apd.js
```diff
@@ -3,6 +3,19 @@ import {
...
+  (acc, [name, regulations]) => ({
+    ...acc,
+    [name]: Object.keys(regulations).map(reg => ({
+      title: reg,
+      checked: false,
+      explanation: ''
+    }))
+  }),
+  {}
+);
```
mgwalker: We should think about this one a little bit, but right now, by design, the assurances list is entirely client-driven.  It's a JSON field on the backend, so it defaults to empty but will happy accept *whatever*.  Since the frontend depends on these values, it made sense to me for the frontend to create them.

This feels acceptable to me right now, but interested in what you think.

web/src/actions/apd.test.js
brendansudol: 😬

web/src/util/serialization/apd.js
brendansudol: this makes sense to me

Issue #826, opened 2018-07-19
Apply sticky header to all sections


lauraponce:
Dev implementation issue for #758

lauraponce:
How does this seem for the interaction with the sticky header? @NAretakis @jeromeleecms @brendansudol 


![screen shot 2018-07-24 at 9 53 59 am](https://user-images.githubusercontent.com/21222704/43153967-9861a6f0-8f27-11e8-81e2-930d62d2fd8d.png)

Once you scroll past main section (e.g. Submission):

![screen shot 2018-07-24 at 9 54 15 am](https://user-images.githubusercontent.com/21222704/43153968-99ed21fc-8f27-11e8-80b4-f048c83fb31a.png)

Once you scroll past subnav (e.g. Certify and Submit)
![screen shot 2018-07-24 at 9 54 28 am](https://user-images.githubusercontent.com/21222704/43153970-9b5c4950-8f27-11e8-9932-9d5450930410.png)

NAretakis:
Perfect! Ship it!

PR #912, opened 2018-08-03
sticky everything!

brendansudol:
This adds that gooey sticky awesomeness to every section (not just the activity ones like before).

One thing of note: we now have "nested" stickies -- a sticky section within a sticky section -- for these cases, the child header sticks right below the parent.

This is not nearly as elegant / beautiful as mocks laid out here (https://github.com/18F/cms-hitech-apd/issues/826), which I think we can work up to eventually, but this is hopefully a nice intermediate step on the way there.

Preview:
image

Video: https://cl.ly/tSU9


NAretakis:
Here I am, just scrolling up and down, enjoying the stickiness. This is fantastic!!

PR #920, opened 2018-08-09
Update dependencies [skip deploy]

mgwalker:
Updates dependencies.  This resolves the vulnerability identified by snyk as well as a bunch of vulnerabilities identified by npm.  The only remaining identified vulnerability is via `webpack-dev-server`, so I'm not worried about that at all.  😛

`tap` is updated by a major version; everything else is minor or patch.  The only breaking change in `tap` is how it compares sets, but since we're not using sets, no problem for us!

- nothing visible


web/package.json
```diff
@@ -56,19 +56,19 @@
...
-    "html-to-draftjs": "^1.4.0",
+    "html-to-draftjs": "1.4.0",
     "i18n-js": "^3.0.3",
     "lodash.kebabcase": "^4.1.1",
     "markdown-it": "^8.4.1",
     "numeral": "^2.0.6",
     "prop-types": "^15.6.0",
     "react": "^16.4.1",
-    "react-dom": "^16.4.1",
-    "react-draft-wysiwyg": "^1.10.7",
```
mgwalker: Remove the version range specifier.  Let's lock this puppy (and the other draftjs-related deps) down.

PR #928, opened 2018-08-10
Dirty state tracking / incremental saves

### LINKED ISSUE #956: Incremental/partial APD saving
We should be able to save only the parts of the APD that have changed.  The API supports it, now we need the frontend to support it.

mgwalker:
Updated to track which pieces of the APD are changed as the user works, and only save those pieces.  Why?  For one, it means we send less data on each save, which should make it possible to put more content in the rich text fields (you just have to remember to save after you edit the rich text field - which might quickly become an argument for automatic saving).  It also means that there's less churn in the database, which should make auditing better (and also possibly make saves faster, but in my experimenting, saves are between 500 ms and 1000 ms either way).

Closes #956 

- no visible changes


web/src/actions/activities.js
```diff
@@ -21,7 +22,20 @@ export const UPDATE_ACTIVITY = 'UPDATE_ACTIVITY';
...
+  dispatch(actionWithYears(ADD_ACTIVITY));
+  dispatch(updateBudget());
+
+  const newKey = getState().activities.allKeys.filter(
+    k => !oldActivities.includes(k)
+  )[0];
+  dispatch({
+    type: ADD_ACTIVITY_DIRTY,
+    data: getState().activities.byKey[newKey]
+  });
```
mgwalker: Sends a separate action to trigger the dirty reducer when we add an activity, because it wants access to the new activity state - but the new activity state isn't available until after the `ADD_ACTIVITY` action is processed.

web/src/reducers/dirty.js
```diff
@@ -0,0 +1,167 @@
...
+      );
+
+    case ADD_ACTIVITY_CONTRACTOR:
+    case REMOVE_ACTIVITY_CONTRACTOR:
+      return u(
+        {
+          dirty: true,
+          data: {
+            activities: {
+              byKey: { [action.key]: { contractorResources: true } }
```
mgwalker: We only need to know that this is dirty - we don't care in what ways.  So just say `true`.  😄

web/src/reducers/dirty.js
```diff
@@ -0,0 +1,167 @@
...
+        action.updates.federalCitations ||
+        action.updates.incentivePayments ||
+        action.updates.narrativeHIE ||
+        action.updates.narrativeHIT ||
+        action.updates.narrativeMMIS ||
+        action.updates.programOverview ||
+        action.updates.previousActivityExpenses ||
+        action.updates.previousActivitySummary ||
+        action.updates.stateProfile
+      ) {
```
mgwalker: These are the APD fields that are filtered on save, so we need to explicitly mark them as dirty when they're changed.

mgwalker:
This got rebased onto master a couple of times, and I think that was a mistake because now old changes are showing up as part of this PR...  😬

web/src/actions/activities.js
```diff
@@ -21,7 +22,20 @@ export const UPDATE_ACTIVITY = 'UPDATE_ACTIVITY';
 const actionWithYears = (type, other) => (dispatch, getState) =>
   dispatch({ type, ...other, years: getState().apd.data.years });
 
-export const addActivity = () => actionWithYears(ADD_ACTIVITY);
+export const addActivity = () => (dispatch, getState) => {
+  const oldActivities = getState().activities.allKeys;
```
brendansudol: slightly more efficient way to do so this to avoid the O(n*n) filter (and take advantage of `find` method):

```
const oldKeys = new Set(getState().activities.allKeys);

...

const newKey = getState().activities.allKeys.find(k => !oldKeys.has(k))

...
```

mgwalker: One question about using a `Set` - do you know if it'd cause us problems with IE support?  I get that it's supported in babel-polyfill but I'm not sure how we're pulling that in, so...  😄 

In any case, `find` is definitely better than `filter`.  👍

brendansudol:
should we merge @mgwalker ?

mgwalker:
Hang tight, going to address your comments first!

mgwalker:
Added tests for the new reducer.  This one should be good to go now @brendansudol

brendansudol:
love it

Issue #935, opened 2018-08-13
Large budget figures don't fit in tables


quarterback:
Program Budget table & federal cost allocation tables do not accommodate 7+ figure amounts neatly.

### This task is done when...
 - [ ] When the spacing in the budget tables sections are cleaned up.

<img width="453" alt="screenshot 2018-08-13 14 13 55" src="https://user-images.githubusercontent.com/4461062/44058471-0a6e5b7c-9f03-11e8-9dbe-12210fe1f370.png">

PR #940, opened 2018-08-14
fixes number overflow

brendansudol:
fixes: https://github.com/18F/cms-hitech-apd/issues/935

preview:
screen shot 2018-08-14 at 3 18 02 pm


web/src/containers/activity/CostAllocateFFP.js
```diff
@@ -37,23 +37,25 @@ class CostAllocateFFP extends Component {
...
                   <DollarInput
+                    wrapperClass="lg-col-8 mb2"
                     name={`cost-allocate-other-${year}`}
                     label={t('activities.costAllocate.ffp.labels.other')}
                     value={costAllocation[year].other}
                     onChange={this.handleOther(year)}
                   />
                   <div>{t('activities.costAllocate.ffp.totalNetOther')}</div>
-                  <div className="h3 bold mono">
+                  <div className="h4 bold mono truncate">
```
mgwalker: Question for @lauraponce @quarterback @NAretakis @jeromeleecms:

Is truncating these dollar numbers okay if they get too big?  (It looks like `$123,345,679...`)  It looks like the cost allocation component can handle into the hundreds of millions before it truncates so ***hopefully*** that should be fine.  (Goodness I hope you don't see many hundred million dollar APDs...)

brendansudol: i think the solution for really big numbers just not be truncating but utilizing SI prefix (https://en.wikipedia.org/wiki/Metric_prefix#List_of_SI_prefixes) -- for example, `$123.45B` or `1.21M`

brendansudol:
@mgwalker i'll fix those budget tables ones here too, thanks for heads up about that

brendansudol:
ok, i adjusted this to better accommodate big money numbers (right now that threshold is a million dollars so it's very easy to adjust).

if the number is less than a cool mil, it gets the standard treatment: `$999,999`
if the number is better, we utilize [SI prefixes](https://en.wikipedia.org/wiki/Metric_prefix#List_of_SI_prefixes) with at most 3 significant digits: `$1.23M` (if there are trailing decimal 0s, they get removed, as in `$8.1M` vs `$8.10M`)

preview:
image

mgwalker:
I like the postfixed dollar values a lot.  Definitely want product (cc: @jeromeleecms) to weigh in, though.  To be clear, this only changes how the dollar values are displayed, not what we're actually storing, so the analyst/reviewer view could still show full dollar values.

quarterback:
Good solution

Issue #938, opened 2018-08-14
Add in list entries to personnel, non-personnel, and contract resources


lauraponce:
Building off of #870 and #872

When this section is first loaded, it will look like how it's currently implemented:

<img width="664" alt="screen shot 2018-08-14 at 11 24 06 am" src="https://user-images.githubusercontent.com/21222704/44110485-ae641f44-9fb4-11e8-9a08-f0ecaa7fde83.png">

As soon as an entry is complete and "add another entry" is clicked, they will see the entry they just finished appear in the blue list. They can then edit the new one and when that is done, they can add another one by clicking add another entry and see their previous entry populate in the list. 

If you click on any of the items in the blue list, they would expand back out with all the fields to edit and make changes. From the blue list, you would just have a quick view of the details and can delete it.

If a user is editing an entry and they click away to another section, when they go back to this section, they would see the one they were editing in the open state where they can make additional changes.

<img width="665" alt="screen shot 2018-08-14 at 11 24 17 am" src="https://user-images.githubusercontent.com/21222704/44110684-3a6e357e-9fb5-11e8-8de5-45ae48d4af07.png">

lauraponce:
@brendansudol adding some more color to the list entry interaction. Let me know if you want to chat about this more or if this makes sense!

PR #955, opened 2018-08-21
add list row entries for expenses

brendansudol:
related to https://github.com/18F/cms-hitech-apd/issues/938, adds clickable list entries for expenses so you can see the relevant info at a glance and/or edit when needed.

### DIFF
--- a/web/src/containers/activity/Expenses.js
+++ b/web/src/containers/activity/Expenses.js
@@ -14,6 +14,7 @@ import Label from '../../components/Label';
 import { Subsection } from '../../components/Section';
 import Select from '../../components/Select';
 import { t } from '../../i18n';
+import { arrToObj } from '../../util';
 import { formatMoney } from '../../util/formats';
 
 const EXPENSE_CATEGORIES = [
@@ -138,7 +139,10 @@ class Expenses extends Component {
 
     if (lastExpense && !(lastExpense in state.showForm)) {
       return {
-        showForm: { ...state.showForm, [lastExpense]: true }
+        showForm: {
+          ...arrToObj(Object.keys(state.showForm), false),
+          [lastExpense]: true
+        }
       };
     }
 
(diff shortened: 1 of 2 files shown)


brendansudol:
Preview:
https://cl.ly/0f032N470e23

web/src/containers/activity/Expenses.js
```diff
@@ -24,26 +26,172 @@ const EXPENSE_CATEGORIES = [
...
+  handleChange: PropTypes.func.isRequired,
+  handleYearChange: PropTypes.func.isRequired,
+  handleDelete: PropTypes.func.isRequired
+};
+
+const getLastExpense = expenses =>
+  expenses.length ? expenses[expenses.length - 1].key : null;
+
 class Expenses extends Component {
+  static getDerivedStateFromProps(props, state) {
```
mgwalker: It looks like this is never used, but it'd almost satisfy [Laura's comment](https://github.com/18F/cms-hitech-apd/pull/955#pullrequestreview-148177788).  It looks like this code is intended to default the **last** expense to opened.

mgwalker:
@lauraponce Would it make sense to get rid of the remove button at the bottom of the expense form in favor of the `X` button at the far right of the expense header?

lauraponce:
@mgwalker totally! That would work perfectly (putting an X in the top right and getting rid of the remove button at the bottom).

web/src/containers/activity/Expenses.js
brendansudol: this is used -- when a new expense gets added, it adds that new expense key to the state and defaults it to be open

brendansudol:
changes coming up!

brendansudol:
updates made @lauraponce :)

lauraponce:
@brendansudol one last small thing — when you click "Add expense", can it collapse the previous entry into the list format? I could see people just continuing to add more without clicking to collapse so they wouldn't get the cool benefit of the high-level view list.

brendansudol:
good call @lauraponce -- i just made that change and the experience is a lot nicer with it :)

PR #972, opened 2018-08-23
Refactor budget reducer [skip deploy]

### LINKED ISSUE #883: Refactor / add more comments to budget reducer / munging methods
Devs can self assign whoever wants to tackle this.

Created per #854

mgwalker:
Refactors the budget reducer to be more straightforward (🤞).  The sequence of computing all the budget data is now almost linear, making it hopefully easier to track where things happen.  Also, lots of comments to explain what's going on.

Closes #883 

- no visible changes

### DIFF
--- a/web/.eslintrc.json
+++ b/web/.eslintrc.json
@@ -17,6 +17,7 @@
         }
       }
     ],
+    "no-param-reassign": [2, { "props": false }],
 
     "react/forbid-prop-types": ["error", { "forbid": ["any"] }],
     "react/jsx-filename-extension": 0,
--- a/web/src/reducers/budget.js
+++ b/web/src/reducers/budget.js
@@ -100,17 +100,29 @@ const buildBudget = bigState => {
   const { years } = bigState.apd.data;
   const newState = initialState(years);
 
+  /**
+   * Adds a given cost to the budget running totals.
+   * @param {String} fundingSource The CMS funding source of the cost
+   * @param {String} year As a four-character year string (e.g., '2018')
+   * @param {String} prop The property this cost comes from (e.g., "contractors", "statePersonnel")
+   * @param {Number} cost The cost value.
+   */
   const addCostToTotals = (fundingSource, year, prop, cost) => {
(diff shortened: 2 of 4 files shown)


web/.eslintrc.json
```diff
@@ -17,6 +17,7 @@
         }
       }
     ],
+    "no-param-reassign": [2, { "props": false }],
```
mgwalker: This allows us to update the props of a param, which feels to me like the 90% case.  This is an older rule meant to prevent people from modifying the `arguments` object, which we don't use (and heavens, nobody should anymore).

Anyway, now we can do things like this:

```js
arr.forEach(o => { o.value += 3; });
```

web/src/reducers/budget.test.js
```diff
@@ -287,7 +287,7 @@ describe('budget reducer', () => {
               },
               subtotal: {
                 state: { dollars: 3060, percent: 1 },
-                contractors: { dollars: 1800, percent: 1.0000000000000002 },
+                contractors: { dollars: 1800, percent: 1 },
```
mgwalker: I was hoping to do this refactor without changing the test file at all, but I feel okay changing this wild floating point values.

web/.eslintrc.json
brendansudol: 👍

web/src/reducers/budget.js
```diff
@@ -91,320 +90,331 @@ const initialState = years => ({
...
-          }
-        )
-      }),
-      {}
-    );
-    newState.activities[activity.key].quarterlyFFP.total = total;
-  });
+    activity.statePersonnel.forEach(person => {
+      Object.entries(person.years).forEach(([year, { amt, perc }]) => {
+        const cost = amt * perc / 100;
```
brendansudol: can we assume this calculation will work? could `amt` or `perc` be blank or null?

web/src/reducers/budget.test.js
brendansudol: yesssssss

PR #980, opened 2018-08-24
state personnel - small entries

brendansudol:
very similar to what was done with expenses

preview:
image

### DIFF
--- a/web/src/util/formats.js
+++ b/web/src/util/formats.js
@@ -1,8 +1,9 @@
 import { format } from 'd3-format';
 
-const isNumeric = n => !Number.isNaN(parseFloat(n)) && Number.isFinite(n);
-
 const BIG_NUMBER_THRESHOLD = 1000000;
+
+// eslint-disable-next-line no-restricted-globals
+export const isNumeric = n => !isNaN(parseFloat(n)) && isFinite(n);
 
 export const fmt = (val, spec = ',.2f') =>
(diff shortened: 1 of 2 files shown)


web/src/util/formats.js
```diff
@@ -1,9 +1,10 @@
 import { format } from 'd3-format';
 
-const isNumeric = n => !Number.isNaN(parseFloat(n)) && Number.isFinite(n);
-
 const BIG_NUMBER_THRESHOLD = 1000000;
 
+// eslint-disable-next-line no-restricted-globals
+export const isNumeric = n => !isNaN(parseFloat(n)) && isFinite(n);
```
mgwalker: Why the change from `Number.isNaN` to the older version?

brendansudol: our old fn was saying that `'0.00'` was not numeric -- this one solves that, though now I see that it's due to the `isFinite` change...

image

brendansudol:
sure thing @lauraponce

web/src/util/formats.js
mgwalker: What if we simplified to:

```js
export const isNumeric = n => Number.isFinite(Number.parseFloat(n));
```

It turns out that `NaN` is not considered finite, the `isFinite` check should be enough, and then we get to use the scoped methods instead of the global ones.  Since this PR is already merged, we can put this on the backburner if it makes sense and change it sometime in the future when we happen to be fiddling in the utils file again.

Issue #970, opened 2018-08-23
Build out updated cost categories section to combine non-personnel and personnel


lauraponce:
The design issue for this was #948 

This resulted from usability testing feedback and some whiteboarding sessions with CMS at MESC. Basically, it's just combining Personnel and Non-personnel into one subsection in Activities and then breaking down by # of FTEs instead of %.

![image](https://user-images.githubusercontent.com/21222704/44553186-8e4c0800-a6e1-11e8-878e-e1ad463e333b.png)

PR #983, opened 2018-08-28
combines personnel, expenses into 1 section

brendansudol:
closes: https://github.com/18F/cms-hitech-apd/issues/970

### DIFF
--- a/web/src/components/StateDash.js
+++ b/web/src/components/StateDash.js
@@ -154,10 +154,7 @@ const StateDash = () => (
               { day: 30, title: 'Performance Progress CoP' }
             ]}
           />
-          <Events
-            month="July"
-            events={[{ day: 7, title: 'Annual APD Due' }]}
-          />
+          <Events month="July" events={[{ day: 7, title: 'Annual APD Due' }]} />
         </Collapsible>
       </div>
     </div>
--- a/web/src/reducers/budget.js
+++ b/web/src/reducers/budget.js
@@ -239,7 +239,7 @@ const buildBudget = bigState => {
 
     activity.statePersonnel.forEach(person => {
       Object.entries(person.years).forEach(([year, { amt, perc }]) => {
-        const cost = n(amt) * n(perc) / 100;
+        const cost = n(amt) * n(perc);
         addCostToTotals(fundingSource, year, 'statePersonnel', cost);
         activityTotalByCategory[year].statePersonnel += cost;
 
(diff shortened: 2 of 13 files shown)


web/src/components/StateDash.js
```diff
@@ -154,10 +154,7 @@ const StateDash = () => (
               { day: 30, title: 'Performance Progress CoP' }
             ]}
           />
-          <Events
-            month="July"
-            events={[{ day: 7, title: 'Annual APD Due' }]}
-          />
+          <Events month="July" events={[{ day: 7, title: 'Annual APD Due' }]} />
```
brendansudol: prettier being prettier (PBP)

NAretakis:
For language purposes, let's change the section names: 

- "Cost Categories" change to "In-House Cost Categories"
- "Contractor Resources" change to "Private Contractor Costs"

Doing this aligns the sections with CMS' accounting practices. Does that make sense @brendansudol ?

brendansudol:
yep, coming up @NAretakis!

web/src/components/StateDash.js
mgwalker: Prettier don't think it be like it is but it do.

brendansudol:
@mgwalker -- i updated the fte percent field to now just represent the actual percentage (i.e., before we stored perc as `80` and then divided by 100 in the budget reducer; now we simply store `.8` and no longer need that divide by 100) -- what this means though is that all existing percent values stored in the database are now much too big; can/should we update them (and divide them all by 100) so that everything is on same page again?

web/src/reducers/budget.js
```diff
@@ -239,7 +239,7 @@ const buildBudget = bigState => {
 
     activity.statePersonnel.forEach(person => {
       Object.entries(person.years).forEach(([year, { amt, perc }]) => {
-        const cost = n(amt) * n(perc) / 100;
+        const cost = n(amt) * n(perc);
```
mgwalker: @brendansudol Just out of curiosity, how long did it to take to find this line to change?

brendansudol: 3 seconds

brendansudol: by the way, i'm also totally fine keeping this number as integer based, instead of float (and then doing the conversion to/from in the react component for display purposes

mgwalker:
I think we can fix it in the serializers.  The values in the database/API are already fractional, so 0.8 instead of 80.  The `toAPI` method divides the FTE by 100, and the `fromAPI` multiplies it by 100.  We could just take that math out, and that should fix t, right?

brendansudol:
ahh, yes, perfect -- will create new PR in a sec for that

PR #989, opened 2018-08-29
Upload file contracts [skip deploy]

### LINKED ISSUE #839: When a user adds a contract file, save it to the backend
i.e., uploading / saving and storing reference in database and then wiring that up in `ContractorResources` component

mgwalker:
Skipping deploy because it won't work until the API is also deployed, and the PR-based deploys only do the frontend, not the backend.

Closes #839

- when you create a new contractor, the APD is saved - that way the contractor exists in the database, so we can attach files to it later
- when you select a file to add to a contractor, the file is uploaded immediately and then attached to the contractor

### DIFF
--- a/api/db/file.js
+++ b/api/db/file.js
@@ -1,3 +1,5 @@
+const o = v => (typeof v === 'object' ? v : JSON.parse(v));
+
 module.exports = () => ({
   file: {
     tableName: 'files',
@@ -22,10 +24,9 @@ module.exports = () => ({
 
     toJSON() {
       return {
+        ...o(this.get('metadata')),
         id: this.get('id'),
-        key: this.get('key'),
-        size: this.get('size'),
-        metadata: this.get('metadata')
+        size: this.get('size')
       };
     }
   }
--- a/api/db/index.js
+++ b/api/db/index.js
@@ -17,6 +17,8 @@ const file = require('./file');
 
 const exportedModels = {};
 
+let knexObject;
+
 const setup = (
   knex = defaultKnex,
   bookshelf = defaultBookshelf,
@@ -38,13 +40,13 @@ const setup = (
   logger.silly(
     `setting up models using [${process.env.NODE_ENV}] configuration`
   );
-  const db = knex(config[process.env.NODE_ENV]);
+  knexObject = knex(config[process.env.NODE_ENV]);
 
   // Get bookshelf instance with our knex config, then
   // load the "registry" plugin - lets us map model
   // names to actual models to get rid of ciruclar
   // dependency graphs
-  const orm = bookshelf(db);
+  const orm = bookshelf(knexObject);
   orm.plugin('registry');
 
   const BaseModel = baseModel(orm, exportedModels);
@@ -75,6 +77,7 @@ const setup = (
 };
 
 module.exports = {
-  setup,
-  models: exportedModels
+  models: exportedModels,
+  raw: (...args) => (knexObject ? knexObject(...args) : null),
+  setup
 };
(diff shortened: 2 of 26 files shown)


api/db/file.js
```diff
@@ -1,3 +1,5 @@
+const o = v => (typeof v === 'object' ? v : JSON.parse(v));
+
```
mgwalker: Bookshelf is inconsistent with JSON columns.  When you assign to them, you have to assign them as strings.  When you fetch them from the database, though, they come back as objects.  If you read the property from the model object after saving it, but without fetching it again, then it's a string.  This just tries to normalize for that.

api/db/index.js
```diff
@@ -75,6 +77,7 @@ const setup = (
 };
 
 module.exports = {
-  setup,
-  models: exportedModels
+  models: exportedModels,
+  raw: (...args) => (knexObject ? knexObject(...args) : null),
```
mgwalker: Exposes the knex object directly because it can be a lot easier to manually create some of our data relationships than to use Bookshelf.  :sigh:

api/routes/openAPI/index.js
```diff
@@ -628,18 +628,14 @@ module.exports = {
           id: {
             type: 'number'
           },
-          key: {
-            type: 'string',
-            description:
-              'File store identifier, used to retrieve the actual file'
-          },
```
mgwalker: There's no reason to expose this key to the outside world.

api/store.js
```diff
@@ -150,6 +150,11 @@ class Store {
           done();
         });
 
+        data.on('end', () => {
+          writeStream.end();
+          done();
+        });
```
mgwalker: Some streams emit the `end` event, some emit the `close` event...  It'd be cool if they could all conform to the spec but I guess that's asking too much.

web/src/actions/activities.js
```diff
@@ -36,8 +42,112 @@ export const addActivity = () => (dispatch, getState) => {
...
+export const addActivityContractor = (key, { save = saveApd } = {}) => async (
+  dispatch,
+  getState
+) => {
+  const activity = getState().activities.byKey[key];
+  const oldIds = activity.contractorResources.map(c => c.id);
+
+  dispatch(actionWithYears(ADD_ACTIVITY_CONTRACTOR, { key }));
+
+  const newApd = await dispatch(save());
```
mgwalker: Add the contractor, then save.  That way we let the dirty reducer do the work of making sure the contractor gets added to the data we save.

api/db/index.js
brendansudol: i understand :)

api/routes/files/post.js
```diff
@@ -0,0 +1,94 @@
...
+            await file.save();
+            logger.silly('file was saved');
+
+            // Inserting a many-to-many link in bookshelf is...  I mean,
+            // the documentation is not clear on how it's supposed to
+            // work, and everything I tried just exploded.  So, drop to
+            // knex and do it manually.  (E.g., the "attach" method is
+            // the intended way, but that removes existing relations,
+            // which is definitely not what we want.)
+            await rawDB('activity_contractor_files').insert({
```
brendansudol: 👍

web/src/actions/activities.test.js
```diff
@@ -185,4 +234,170 @@ describe('activities actions', () => {
...
+      append: sinon.stub()
+    };
+    const FormData = sinon.stub().returns(formdata);
+
+    fetchMock.onPost('/files/contractor/contractor id').reply(200, {
+      id: 'new file id',
+      metadata: 'bloop bloop'
+    });
+
+    const notifyAction = sinon.stub().returns({ type: 'notification stub' });
```
brendansudol: ah, now i see why this was a param

PR #993, opened 2018-08-30
export view poc

brendansudol:
A PoC around a very lightweight way to make the APD form more print / export friendly.

Definitely still needs a bunch of refinement, but could potentially work (may need to move this `printMode` variable into redux store to further customize downstream components). Biggest thing still to do is a way to handle tables, but there are options there. One idea for that: doing something like [this](https://codepen.io/AllThingsSmitty/pen/MyqmdM) where the table collapses to a list for print mode. Could also just do a more explicit rendering of both a table view and list view for these components and toggle the visibility based on edit / print mode. 


mgwalker:
👀

quarterback:
!!!

PR #996, opened 2018-08-30
Put goals and objectives under little collapsible sections too!

### LINKED ISSUE #988: Build out the cleaned up Needs and Objectives section
As discussed in #943, the Needs and Objectives section needs a little bit of finetuning to match the styles of the other sections.

image

(Related: I think @quarterback is working on updating the content for the help text for this section)

mgwalker:
Closes #988 

- The goals and objectives under the "Performance Measures and Benchmarks" section are under the happy little blue collapsible sections just like personnel/non-personnel/contractor expenses.

### DIFF
--- a/web/src/containers/activity/Goals.js
+++ b/web/src/containers/activity/Goals.js
@@ -9,66 +9,149 @@ import {
 } from '../../actions/activities';
 import Btn from '../../components/Btn';
 import { RichText } from '../../components/Inputs';
+import MiniHeader from '../../components/MiniHeader';
+import NoDataMsg from '../../components/NoDataMsg';
 import { Subsection, SubsectionChunk } from '../../components/Section';
+import { arrToObj } from '../../util';
 import { t } from '../../i18n';
 
+const plaintext = txt => {
+  const e = document.createElement('div');
+  e.innerHTML = txt;
+  return e.innerText;
+};
+
+const GoalForm = ({ goal, idx, handleChange, handleDelete }) => (
+  <div className="mt2 mb3">
+    <Btn
+      kind="outline"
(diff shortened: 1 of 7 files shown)


mgwalker:
@lauraponce 👍 Should the x in the blue box be completely removed, or just when the goal is expanded?

And should we also update the expense sections to behave the same way?  (They all use the same blue bar element now, so if they need to behave differently, that's fine I just need to make sure to account for that.  😄)

lauraponce:
Good catch on the expenses section! Yes, let's remove that extra X there also

brendansudol:
bless you for abstracting the mini-header component!

web/src/containers/activity/Goals.js
```diff
@@ -9,66 +9,149 @@ import {
...
+        {goals.length === 0 ? (
+          <NoDataMsg>{t('activities.expenses.noDataNotice')}</NoDataMsg>
+        ) : (
+          <div className="mt3 pt3 border-top border-grey">
+            {goals.map((goal, i) => (
+              <div key={goal.key}>
+                <MiniHeader
+                  content={
+                    <div className="col-11 truncate">
+                      {plaintext(goal.description)}
```
brendansudol: what are your thoughts on using this `plaintext` method vs. the `Md` component here?

mgwalker: What would happen if the start of the description was an H1?  I'm open to giving it a try and seeing what happens - the `plaintext` method doesn't exactly fill me with joy.  😆

brendansudol: ohh, that's a good point -- alright, plaintext works for me then!

Issue #722, opened 2018-06-27
Build out the new state dashboard


lauraponce:
Use #676 as a reference (the only thing changing in the design since the last version posted is removing the priority column in the tasks section)

NAretakis:
@lauraponce can we spend just a bit of time on this to wireframe the dash based on the latest user feedback, and then Devs and run with it.

lauraponce:
Dashboard tweaks:

- Remove the Due Date column in the Tasks section
- Swap places of My Tasks section and the Statuses section
- Update Statuses states to be: `Drafting` **or** `Submitted`, `Received`, `Under Review` **or** `Pending State Response to CMS Comments`, `In Clearance`, and `Approved` **or** `Disapproved` **or** `Withdrawn` (So there are a total of 5 circles along the status line.)
- Add dates under the status states such as:

<img width="740" alt="screen shot 2018-09-07 at 1 45 37 pm" src="https://user-images.githubusercontent.com/21222704/45242311-51b50a80-b2a4-11e8-8a7b-c7e52e580529.png">

lauraponce:
^ These updates are based on the feedback from the last round of research: https://docs.google.com/document/d/1AqZpgtV_aqstOlKnfZLT_gzQpRI26TBT9KGtPHmCxDA/edit#

@NAretakis @jeromeleecms does this list look right as the next step in making some UI tweaks to the dashboard? [This](https://hitech-apd-pr1008.app.cloud.gov/) is the current state of the dash from @mgwalker's latest [PR](https://github.com/18F/cms-hitech-apd/pull/1008).

lauraponce:
I was thinking that maybe we don't need the `Received` status as a state since it should be a given once it's submitted, but I can see it giving people extra peace of mind seeing a greencheck there showing that their APD is now in the hands of CMS even if the actual review process hasn't begun. Open to other opinions on this though!

NAretakis:
I don't know how we'd calculate `Expected Due Date` unless we just did 60 days from the submission. That said, I am not sure how useful it would be because we definitely want to be under that target. I think I prefer not have expected date there (unless we actually had some metrics/math to make it realistic). And then it gets a date stamp once we mark it as approved. That work for you @lauraponce ?

mgwalker:
[...]

PR #1008, opened 2018-09-06
Implement state dashboard refinements

mgwalker:
Addresses #722.

- updates the dashboard
- hooks up the dashboard to the real list of APDs, at least nominally

### DIFF
--- a/web/src/components/SidebarLink.js
+++ b/web/src/components/SidebarLink.js
@@ -24,12 +24,24 @@ const getClasses = (depth, hash, anchor) => deline`
   ${depth ? 'pl2' : ''} ${hash === anchor ? 'sb-item-active' : ''}
 `;
 
+const clickHandler = (depth, toggleExpand, propsOnClick) => e => {
+  if (depth === 0) {
+    toggleExpand();
+  }
+  if (propsOnClick) {
+    propsOnClick();
+    e.preventDefault();
+    e.stopPropagation();
+  }
+};
+
 const SidebarLink = ({
   anchor,
   children,
   depth,
   expanded,
   hash,
+  onClick,
   sub,
   toggleExpand,
   ...rest
@@ -39,7 +51,7 @@ const SidebarLink = ({
       <a
         href={`#${anchor || '!'}`}
         className="white text-decoration-none truncate"
-        onClick={depth === 0 ? toggleExpand : undefined}
+        onClick={clickHandler(depth, toggleExpand, onClick)}
         {...rest}
       >
         {children}
@@ -61,6 +73,7 @@ SidebarLink.propTypes = {
   depth: PropTypes.number,
   expanded: PropTypes.oneOfType([PropTypes.bool, PropTypes.object]),
   hash: PropTypes.string,
+  onClick: PropTypes.func,
   sub: PropTypes.array,
   toggleExpand: PropTypes.func
 };
@@ -70,6 +83,7 @@ SidebarLink.defaultProps = {
   depth: 0,
   expanded: null,
   hash: null,
+  onClick: null,
   sub: [],
   toggleExpand: () => {}
 };
--- a/web/src/components/StateDashboard.js
+++ b/web/src/components/StateDashboard.js
@@ -43,6 +43,27 @@ Events.propTypes = {
   events: PropTypes.array.isRequired
 };
 
+const TaskRow = ({ btnIcon, btnText, status, type }) => (
+  <tr
+    className="align-middle bg-grey-light border-top border-bottom border-white"
+    style={{ borderWidth: '0.5rem' }}
+  >
+    <td>{type}</td>
+    <td>{status}</td>
+    <td>
+      <Btn size="small" extraCss="col-12 p1 bg-blue white">
+        {btnText} <Icon icon={btnIcon} className="ml1" />
+      </Btn>
+    </td>
+  </tr>
+);
+TaskRow.propTypes = {
+  btnIcon: PropTypes.object.isRequired,
+  btnText: PropTypes.string.isRequired,
+  status: PropTypes.string.isRequired,
+  type: PropTypes.string.isRequired
+};
+
 const TaskTable = () => (
   <table className="table-fixed">
     <thead>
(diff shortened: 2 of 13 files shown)


web/src/components/SidebarLink.js
```diff
@@ -24,12 +24,24 @@ const getClasses = (depth, hash, anchor) => deline`
...
+const clickHandler = (depth, toggleExpand, propsOnClick) => e => {
+  if (depth === 0) {
+    toggleExpand();
+  }
+  if (propsOnClick) {
+    propsOnClick();
+    e.preventDefault();
+    e.stopPropagation();
+  }
+};
```
mgwalker: This `depth` piece already existed, but the gist of it is that if this link is at the top of the tree, then a click should expand it to show the links underneath it.

The `propsOnClick` thing is new - the idea is that if the thing creating this link hooks up its own click handler, we'll use that and suppress the default click event.

web/src/components/StateDashboard.js
```diff
@@ -0,0 +1,171 @@
...
+              events={[{ day: 7, title: 'Annual APD Due' }]}
+            />
+          </DashboardSection>
+        </div>
+      </div>
+    </div>
+  </div>
+);
+
+export default StateDashboard;
```
mgwalker: Github says this is a new file, but this is just a renamed file with a bunch of changes, so...  I guess that's about the same?

Several of the components in here are essentially placeholders - they're not hooked up to any real data, but we want users to be able to see what the intent is, so we went ahead and built them out.  They're not necessarily super great (especially several components in one file 😬), but I expect they'll get split out over time as they get hooked up to real data.

web/src/containers/StateDashboardCurrentDocuments.js
```diff
@@ -0,0 +1,205 @@
...
+    case 'approved':
+      states.one.done = true;
+      states.two.done = true;
+      states.three.done = true;
+      actionButtonText = 'Approval letter';
+      break;
+
+    default:
+      break;
+  }
```
mgwalker: I'm really not wild about this `switch` statement to map a status into state info, but I wasn't sure offhand of a better way...

web/src/containers/StateDashboardCurrentDocuments.js
```diff
@@ -0,0 +1,205 @@
...
+          <DocumentItem
+            key={apd.id}
+            name={`APD for ${apd.years.join(', ')}`}
+            status={apd.status}
+            buttonClick={this.open(apd.id)}
+          />
+        ))}
+        <DocumentItem name="Example one" status="reviewed" />
+        <DocumentItem name="Example two" status="in review" />
+        <DocumentItem name="Example three" status="approved" />
```
mgwalker: These are here to show off the various states that the progress bar could be in, for usability testing.

mgwalker:
There's still work to do to get it up to Laura's design (e.g., the "approval letter" button needs to be a link; tables need to be sortable - or at least have a sorting dropdown that doesn't do anything; fonts need to be tweaked), but I thought this would be close enough we could move forward with it.  Let me know if it's not and I'll dive back in.  😄

jeromeleecms:
@mgwalker If we merge this, do we need to spin up a separate issue for the content @lauraponce included in #722 this Friday? The link in #722 to the viewable version of the PR says it no longer exists so I can't check what it looks like...

NAretakis:
Wait, is HiC-Bot talking to CMS's AWS?

mgwalker:
Yep.

mgwalker:
@jeromeleecms I think I can roll them into this.

web/src/containers/StateDashboardCurrentDocuments.js
hbillings: The only thing that comes to mind is creating an object that maps these to the status data, like:
```
words = {
  'draft': {
    'actionButtonText': 'View',
    'actionButtonIcon': faEdit,
  }
}
```
and then you could just return something like `{ words[status].actionButtonText, words[status].actionButtonIcon }`. I'm not sure it's _better_, just different.

The other thing that might come in really handy is a `constants` file where you can define all your text. CALC does this [here](https://github.com/18F/calc/blob/develop/frontend/source/js/data-explorer/constants.js), and it's saved me a world of typo pain. (Also makes it _super_ easy when someone decides that text needs to be tweaked slightly.)

web/src/containers/StateDashboardSidebar.js
```diff
@@ -0,0 +1,80 @@
...
+class Sidebar extends Component {
+  createApd = () => this.props.createApd();
+
+  pickApd = id => () => this.props.selectApd(id);
+
+  render() {
+    const { apds, place } = this.props;
+
+    return (
+      <div className="site-sidebar bg-teal relative">
```
hbillings: This should probably be an `aside` element.

web/src/containers/StateDashboardCurrentDocuments.js
mgwalker: Oooh, I like it.  This button text will eventually come from a resource file, but the key can live in a constants file.  I'll make that change!

For the statuses, those are read-only from the API, so we could massage them into `Symbol`s also defined in a constants file, to make more guarantees about using them correctly elsewhere.  🤔  Probably shouldn't tack more onto this PR, but I love this as a future cleanup.  The statuses aren't used much right now, so good time to make that change.

web/src/containers/StateDashboardCurrentDocuments.js
```diff
@@ -0,0 +1,252 @@
...
+  //     done: false,
+  //     started: false,
+  //     title: 'In Clearance'
+  //   },
+  //   five: {
+  //     done: false,
+  //     started: false,
+  //     title: 'Approved'
+  //   }
+  // };
```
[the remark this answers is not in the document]
mgwalker: Oopsie.  Yes!

web/src/components/StateDashboard.js
```diff
@@ -0,0 +1,166 @@
...
+          (ex.) {e.title}
+        </div>
+      </div>
+    ))}
+  </div>
+);
+
+Events.propTypes = {
+  month: PropTypes.string.isRequired,
+  events: PropTypes.array.isRequired
```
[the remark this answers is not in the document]
mgwalker: Interesting!  That makes total sense.  I wonder why we've never done it that way before.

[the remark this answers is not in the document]
mgwalker: Oh, I see.  Our linter complains if we set `defaultProps` for required props.  Interesting...  https://github.com/yannickcr/eslint-plugin-react/blob/master/docs/rules/default-props-match-prop-types.md

Issue #1005, opened 2018-09-05
Revert yaml help text


quarterback:
Right now, yaml file has help text constructed as reminder text. With future changes to styling, it'll be necessary to revert it back to help text. 

### This task is done when...
 - [ ] Reconstruct help text file content

quarterback:
This issue closes with PR #1010

quarterback:
For this issue, I wrote reminder text to clarify the difference between key personnel.

_ Key personnel are stakeholders involved globally with activities contained with this APD. Individuals with specific responsibilities on an activity basis only, should be listed individually in Program Activities section of the eAPD._

@jeromeleecms @NAretakis

quarterback:
Merged

PR #1010, opened 2018-09-10
YAML Fixes + other content changes

quarterback:
- Reverted YAML changes to help reminder text back to help text. Other content changes. (#1005)


web/src/i18n/locales/en.yaml
```diff
@@ -67,12 +67,12 @@
...
       title: Program Overview
-      helpText:
-      reminder: >-
-        Provide a brief, high level summary of the state program including history,
+      helpText: >-
+       Provide a brief, high level summary of the state program including history,
         and future vision. Include history and summary of future
-        vision of the state program. Indicate which federal fiscal years this
-        APD covers.
+        vision of the state program.
```
jeromeleecms: The previous sentence sounds like exactly the same guidance.  If I'm reading this correctly in GH:
"Provide a brief, high level summary of the state program including history, and future vision. Include history and summary of vision of the state program."

The second sentence seems redundant.

web/src/i18n/locales/en.yaml
```diff
@@ -193,31 +194,30 @@
...
       summary: 
         title: Brief Activity Summary
       detail:
         standard:
           subheader: 
         hie:
           subheader:
-          helpText:
+          helpText: >-
+           Detail linkage to meaningful use, including value proposition to providers, and a sustainability plan describing how the activity will be sustained after HIE DDI funding is exhausted.
```
jeromeleecms: Just poking around the YAML, do we need to define acronyms within the text somewhere? I did a search for DDI in the YAML and its never defined. Other examples like HIT, HIE, etc.

jeromeleecms:
Just a few comments @quarterback. 

Thoughts about the CircleCI test failures?

web/src/i18n/locales/en.yaml
quarterback: I do have the glossary in the wiki but this is a good question that it might be worth defining acronyms, I had the sense that people would be familiar with the terms but probably not good to fiat that. I'll make a change.

quarterback: Also on CircleCI, let me go back and look at the code changes there is probably an error somewhere this happened once before one of the perils of the most recent changes is it sometimes caused these.

web/src/i18n/locales/en.yaml
```diff
@@ -367,13 +366,17 @@
         payers who may benefit must contribute their share from the outset.
         The absence of ancillary payers is not sufficient cause for
         Medicaid to be primary payer.
+
+      reminder: >-
+      Include any participating non-Medicaid FTEs.
```
mgwalker: I think this line needs to be indented a bit further to get the CircleCI build to go.

quarterback: Thanks! I'll fix it now.

quarterback:
YAML checks out as valid now.

quarterback:
Gonna close this and push a different version.

PR #1022, opened 2018-09-17
Add help text id to brief activity summary

quarterback:
The help text was missing from brief activity summary because it didn't have a help text identifier beneath it. This fixes that.

- Adds help text id and help text (tbd) to the brief activity summary subsection.


quarterback:
@jeromeleecms okay so this still didn't fix it.  Should spin up a bug issue for when Greg gets back to take a look at why the text isn't showing up there. cc: @jameshupp

jameshupp:
@quarterback What should we see that we're not seeing? Looking at the files changed and the bot-created version, it looks correct to me. (Apologies for using the PR discussion on a text change to educate myself about possible issues with tech underpinning the product.)

quarterback:
screenshot 2018-09-18 13 26 03

@jameshupp For HIT or MMIS, the 2nd box has no help text content, but for HIE it does. But in typing this I think I figured out how to fix it. 

screenshot 2018-09-18 13 29 21

jeromeleecms:
Spun up #1024 to track this bug.

mgwalker:
@quarterback The HIT and MMIS activity summary second-box help texts both come from the same value in the resource file.  The keys are `activities.description.detail.standard.helpText` and `activities.description.detail.standard.reminder`, so you'd add them alongside `subheader` here:

https://github.com/18F/cms-hitech-apd/blob/d7e0692a6b8fee28c6599437710b9ae8ac7ede65/web/src/i18n/locales/en.yaml#L222

I think this is a ***really*** good example of where having multiple resource files would be helpful.  How on earth would have ever figured out those are shared, and it's called "standard"?  The only way to know is to look in the code to see what it's referencing.  😬

mgwalker:
One thing that might be helpful, in development you can add `#18n` to the end of the URL then reload the page.  That will disable the text resource and instead display the key.

Unfortunately, all of our deploys are built in production/staging mode, so that functionality is disabled.

quarterback:
@mgwalker Thanks for explaining! #moopmoop

mgwalker:
Closing and opening another branch with these changes, but in the split-out files.

Issue #1004, opened 2018-09-05
Explore a different structure to yaml file


quarterback:
The current structure of the yaml file is unwieldy. Consider a universe where we split it into separate files, if possible.

### This task is done when...
 - [ ] 
 - [ ]

NAretakis:
Done!

PR #1036, opened 2018-09-25
Split giant resource file into several small resource files

mgwalker:
Instead of one ginormous, hard-to-manage YAML file, now we have 20 small YAML files.

Do you think this is more manageable @jameshupp and @quarterback?  Is this too many files?  We can combine or break out files to your heart's content.

For #1004.

- no visible changes


web/src/i18n/locales/en/app.yaml
```diff
@@ -0,0 +1,20 @@
+# These are shared throughout the app.
+
+# This is the title displayed in the sidebar.  First is when logged in, second
```
mgwalker: Would it be useful to have comments in all of the files describing which part of the site they apply to, as well as some descriptions of the sections of the file?

jameshupp: I think that'd help a lot. Does it make most sense to have it be the first line, above the `title`?

mgwalker:
How about if we add the comments as we go?  So as we change the text, throw in some comments if that's helpful for the future?

PR #1046, opened 2018-10-11
Update content (from #1022)

mgwalker:
Carried over from #1022.  It was easier to do this than to "fix" the other branch.  😄

- Adds help text id and help text (tbd) to the brief activity summary subsection.

### DIFF
--- a/web/src/reducers/activities.test.js
+++ b/web/src/reducers/activities.test.js
@@ -19,12 +19,12 @@ describe('activities reducer', () => {
     name: '',
     start: '',
     files: [],
-    years: { '2018': 0, '2019': 0 },
+    years: { '2019': 0, '2020': 0 },
     hourly: {
       useHourly: false,
       data: {
-        '2018': { hours: '', rate: '' },
-        '2019': { hours: '', rate: '' }
+        '2019': { hours: '', rate: '' },
+        '2020': { hours: '', rate: '' }
       }
     }
   });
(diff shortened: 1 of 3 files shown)


web/src/reducers/activities.test.js
```diff
@@ -19,12 +19,12 @@ describe('activities reducer', () => {
     name: '',
     start: '',
     files: [],
-    years: { '2018': 0, '2019': 0 },
+    years: { '2019': 0, '2020': 0 },
```
hbillings: Should we be setting these to the current year/year +1 dynamically?

mgwalker: That'd be more robust, but I'm hesitant because it makes the tests nondeterministic.  I think in this case it's simple enough it'd probably be fine, but ideally I'd probably want to use Sinon's time mocking and force everything into a known year, regardless of the actual current year.  Then the tests would pass forever, and they'd be deterministic too.

I'm 100% on board if you think we should make this test more robust before we merge it, though.

web/src/reducers/activities.test.js
```diff
@@ -1,12 +1,32 @@
...
+// 13, 1972.  (As everyone knows, December is the 11th month.)  FFY 1973.  Set
+// this clock before we import anything, so that the stuff we import will use
+// our faked-out clock.
+const mockClock = sinon.useFakeTimers(new Date(1972, 11, 13).getTime());
+
+// Need to use require() here instead of import, because apparently Babel will
+// reorder code and put imports before everything else.  The upshot is that
+// the code under test would get loaded before the clock is faked, and the util
+// library computes years as soon as it's loaded, so...  require() is left
+// alone, so it will come after the clock fakery.  :pixar-joy:
```
hbillings: This blew my mind.

PR #1051, opened 2018-10-16
Round everything to whole dollars when calculating budgets

### LINKED ISSUE #788: When calculating budget tables, round values to whole dollars
CMS expects dollar values to be in whole dollars, so when we do calculations to fractional values, we need to round them.  Currently we round the dollar amounts when we display them, but they are stored with up to two decimal places.

@jeromeleecms Should we do a standard rounding operation?  Always round up?  Always round down?

### This task is done when...
 - [ ] data model is updated to store whole numbers
 - [ ] if necessary, calculations are updated to round to whole numbers
   - updating the data model may actually be sufficient - let the database do the rounding for us!

mgwalker:
When calculating all the budget values, rounds everything to whole dollars while making sure we maintain the original sums so everything should check out.  This is only used where we deal with percents:

- we sum up total costs for each funding source (HIE, HIT, MMIS), by cost category (personnel, contractor, non-personnel)
- we compute the federal and state shares of the totals
  - this is computed by subtracting any other funding amount and then using the percentages the user provided in the cost allocation section: `(total - other) * percent`
  - we use our magic rounding method on this one, to guarantee that the federal + state shares will always sum to the total
- we compute what percent each cost category contributes to the total
  - the values provided by the user implicitly include part of the "other funding" amount
  - this percent is `category-amount / total`
  - the cost categories contribute the same percentage to the federal and state shares as computed above
- we compute the federal and state shares of the individual cost categories
  - we use our magic rounding again, this time to take the total share and spread it across the cost categories, based on what percentage each one contributed to the total
  - e.g., `federal-total * state-personnel-percent` would be the federal share of state personnel costs

Uses a slight modification of the [largest remainder method](https://en.wikipedia.org/wiki/Largest_remainder_method)

Closes #788.  There's a change identified at the end of the conversation in #788 that we should capture elsewhere, but that one will require keeping track of the Medicaid grand total, which is the original grand total minus any other funding.  I think that already exists in the code, we just need to make sure it goes out to the rest of the app, and some of our tables will need to be adjusted to use it.

- Maybe no visible changes?  Maybe some.  It really depends on whether the numbers you've put in resulted in any weird rounding situations.  😄


web/src/util/roundedPercents.test.js
```diff
@@ -0,0 +1,14 @@
+import roundedPercents from './roundedPercents';
+
+describe('rounding percentages util', () => {
+  test('spreads a number across a set of percentages', () => {
+    expect(roundedPercents(224, [0.15, 0.19, 0.3, 0.36])).toEqual([
```
mgwalker: Picked these values because if you split 224 into 15%, 19%, 30%, and 36%, round all those pieces up, and then add them together, you get 225.  So...  It's useful to know our magic thingy doesn't do that.

web/src/util/roundedPercents.test.js
```diff
@@ -0,0 +1,14 @@
...
+      34,
+      42,
+      67,
+      81
+    ]);
+
+    expect(roundedPercents(100, [1 / 3, 1 / 3, 1 / 3])).toEqual([33, 33, 34]);
```
mgwalker: And these because 1/3 is funky.  This test actually found a bug where I was taking the floor of a number instead of rounding it, so I was actually summing to 99.

web/src/reducers/budget.test.js
```diff
@@ -578,24 +578,24 @@ describe('budget reducer', () => {
             },
             '1933': {
```
mgwalker: The only changes in this file should be rounding numbers one way or the other.

web/src/reducers/budget.js
```diff
@@ -250,113 +254,172 @@ const buildBudget = bigState => {
...
+      [['contractors', 'expenses', 'statePersonnel'][i]]: value
+    });
 
     // Now loop back over the years and compute state and federal shares
     // of all the costs.
     years.forEach(year => {
-      const other = getOtherFunding(year);
+      const totalCost = activityTotals.data.combined[year];
+      const totalGovShare = totalCost - n(allocation[year].other);
+      const costShares = getAllocation(year, totalGovShare);
```
mgwalker: 100% open to better names for these variables.  `totalCostShare` should maybe be `totalMedicaidShare`, since it represents the total amount that Medicaid is responsible for, whether at the state or federal level.

`costShares` is the Medicaid total share, broken into state and federal shares.

web/src/reducers/budget.js
```diff
@@ -250,113 +254,172 @@ const buildBudget = bigState => {
...
+        // need to track subtotals, not individual cost categories
+        [ffpLevel, 'combined'].forEach(prop => {
+          newState.mmisByFFP[prop][year].federal += costShares.fedShare;
+          newState.mmisByFFP[prop].total.federal += costShares.fedShare;
+
+          newState.mmisByFFP[prop][year].state += costShares.stateShare;
+          newState.mmisByFFP[prop].total.state += costShares.stateShare;
+
+          newState.mmisByFFP[prop][year].total += totalCost;
+          newState.mmisByFFP[prop].total.total += totalCost;
```
hbillings: I was going to make an argument for making this less redundant, but it being so explicit makes it way easier to read (for me, anyway)!

web/src/util/roundedPercents.js
```diff
@@ -0,0 +1,52 @@
...
+
+  if (difference !== 0) {
+    // We want to spread the difference among the rounded numbers in order
+    // of their remainders.  That is, the real number with the largest
+    // decimal portion reflects which rounded number we would add one to
+    // first.  Then we'll pick the next number and add one again.  Until
+    // all of the difference is accounted for.
+    const indicesByLargestRemainder = [...Array(realNumbers.length)]
+      .map((_, i) => i)
+      .sort((a, b) => (realRemainders[a] > realRemainders[b] ? -1 : 1));
```
hbillings: mind-blown.gif

web/src/reducers/budget.js
```diff
@@ -250,113 +254,172 @@ const buildBudget = bigState => {
...
+      const costCategoryPercentages = [
+        activityTotals.data.contractors[year] / totalCost,
+        activityTotals.data.expenses[year] / totalCost,
+        activityTotals.data.statePersonnel[year] / totalCost
+      ];
+
+      // Compute the actual cost of each cost category, based on the federal
+      // and state shares and the percentages calculated above.
+      const fedShare = roundedPercents(
+        costShares.fedShare,
```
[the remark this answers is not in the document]
mgwalker: The comment on this piece of code isn't great (probably made sense when I was in the weeds, but now, not so much), but yes, it's okay here.  The `costShares` is being divvied up into cost categories, and those cost categories will sum back to `costShares`.  It's...  confusing, because we're calculating a percent of a percent.  But I think a tree helps me understand it:

```
100%                                _____ $1,500 ____
                                  /                  \
                                 /                    \
90%/10%                   $1,350                        $150
federal/state            /   |  \                     /   |  \
                        /    |   \                   /    |   \
23%/42%/35%        $310.50  $567  $472.50        $34.50  $63  $52.50
personnell/        $311     $567  $472           $35     $63  $52
contractor/
other
```

(that took longer than I want to admit...)

Anyway, what's happening is that `costShares` isn't being rounded again - it's being used as the basis for rounding the cost of each cost category, such that the cost categories sum up to the `costShares`, which in turn sum up to the total Medicaid cost.

It's a little convoluted, but it's because of the "other funding."  It comes off the top, not out of any particular cost category, but CMS wants to see where their dollars are going.  In order to be able to show that, we have to deduct the "other funding" amount proportionally from each cost category.  😬  We actually used to do it that way - we figured the proportions, divided up the "other funding" accordingly, and subtracted it from each cost category total.  But it turns out that once you have the proportions, it's a lot simpler to take the other funding off the total and then divvy up the Medicaid total proportionally instead.

I don't know if this is making anything clearer.  The best I can offer is that we can spend a half hour some time walking through all the numbers in the app to show where they came from, if that'd be helpful.  There are a bunch of them!

web/src/reducers/budget.js
mgwalker: Oops, yes, the latter.  I'll go make these changes!

web/src/reducers/budget.js
mgwalker: Back in the Before Times™ (first iteration), there were lots of loops and abstractions, but then when we needed to change how something was calculated it took days and days to figure out where that happened.  Literally, days.

Issue #1027, opened 2018-09-20
Focused review of the Help Text within the Activity Portion of the App


jeromeleecms:
Target areas which may interrupt flow / data entry integrity + evaluate feedback from usability testing for the Activity portion of the App

### This task is done when...
- [x] Inventory the section (in progress)
- [x] Compare with usability testing feedback
- [x] Review my inventory with SMEs (Nick or Jerome)
- [x] Edit the yaml and work through via PR until consensus

That checklist was originally in a comment on #953. That was closed because it was not an actively managed issue and what it was addressing on should emerge from this work, as well.

jameshupp:
A couple things discussed for this that aren't in #1056, just noting here so I have a quick reference fo creating subsequent issues on them:
- whether the "Add" button language works properly for the look and function of the buttons
- eliminating start dates for milestones and/or adjusting the label away from "milestones" because milestones don't feel like things with ranges to me, but again does that shift require design changes?
- section titling: how do we feel about existing case choices (when it's title, when it's sentence, etc.)

jameshupp:
Done!

PR #1056, opened 2018-10-19
For review/discussion - basic language updates to activity section

jameshupp:
This addresses #1027, done as much to make sure I understand what’s in this section as anything else.

- reconciles language around goals, benchmarks, objectives, etc. I settled on `goals` and `benchmarks` because the distinction between those words ought to be more generally obvious.
- shifted out of a couple uses of the word `expenses` in favor of `costs` (there is one left as it was because I think it’s playing a different role)

There are a couple outstanding items left from the initial inventory of the section, but they’re more complex than simple language things so I want to raise them with other designers. Will open separate issues on them.

- [x] Visual check shows the changes aren’t causing any issues with rendering
- [x] These changes have been reviewed by someone other than the original author


web/src/i18n/locales/en/activities/goals.yaml
```diff
@@ -1,16 +1,19 @@
...
+reminder: >-
+  Examples of goals include:
+  improved health outcomes for benficiaries, transformation of public health services,
+  enabling implementation of other initiatives, etc.
 noDataNotice: There are no goals set for this activity.
 goal:
   title: 'Goal #{{number}}'
 objective:
-  title: Goal roadmap
+  title: Benchmarks
```
NAretakis: Suggest that we go with "Measures" in place of "Benchmarks" as I think that is more CMS-ish language. That seem ok?

web/src/i18n/locales/en/activities/goals.yaml
```diff
@@ -1,16 +1,19 @@
-title: Performance Measures and Benchmarks
+title: Performance Goals and Benchmarks
```
NAretakis: Benchmarks --> Measures

web/src/i18n/locales/en/activities/goals.yaml
```diff
@@ -1,16 +1,19 @@
...
 noDataNotice: There are no goals set for this activity.
 goal:
   title: 'Goal #{{number}}'
 objective:
-  title: Goal roadmap
+  title: Benchmarks
   helpText: >-
-    Briefly describe how state will measure progress
-    toward each identified goal.
+    Briefly describe how state will measure progress toward the identified goal.
```
NAretakis: `Briefly describe how **the** state will measure progress toward the identified goal."

web/src/i18n/locales/en/activities/goals.yaml
NAretakis: Nevermind. I spent some time thinking about this, and the definition for benchmark works better here than measure.

web/src/i18n/locales/en/activities/goals.yaml
NAretakis: Nevermind. I spent some time thinking about this, and the definition for benchmark works better here than measure.

PR #1058, opened 2018-10-19
Restores missing helpText to the Standard Activity Overview section

### LINKED ISSUE #1024: Help Text is Missing
During review of the app / user testing, we discovered that help text for the Activity Overview section was missing:

image

Missing text was addressed in #1022, but now there is no text in the boxes:

image

### This task is done when...
 - [ ] Appropriate updates made to app to pull in tagged YAML data
 - [ ] Someone else has approved of the update.

jeromeleecms:
Digging through an old en.yaml file, this adds back in the text that was missing from the standard helptext activity overview. If this works, this should close #1024.

Note: Also using this as a learning experience to see how to edit the YAML files. 

@jameshupp Feel free to edit the language. I took a look at #1056 and this change does not overlap with your proposed changes.

- Adds text "Include as much detail necessary to explain the activity." to the helpText in the overview section

### DIFF
--- a/web/src/i18n/locales/en/activities/overview.yaml
+++ b/web/src/i18n/locales/en/activities/overview.yaml
@@ -12,7 +12,6 @@ summary:
 detail:
   # Standard is for HIT and MMIS
   standard:
-    subheader
     helpText: >-
       Include as much detail necessary to explain the activity.
   hie:
--- a/web/src/i18n/locales/en/activities/overview.yaml
+++ b/web/src/i18n/locales/en/activities/overview.yaml
@@ -12,7 +12,7 @@ summary:
 detail:
   # Standard is for HIT and MMIS
   standard:
-    ? subheader
+    subheader
     helpText: >-
       Include as much detail necessary to explain the activity.
   hie:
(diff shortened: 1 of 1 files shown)


web/src/i18n/locales/en/activities/overview.yaml
```diff
@@ -12,8 +12,8 @@ summary:
 detail:
   # Standard is for HIT and MMIS
   standard:
-    ? subheader
-
+    helpText: >-
+      Include as much detail necessary to explain the activity.
```
jameshupp:
```suggestion
      Include as much detail as is necessary to explain the activity.
```
`as is` creates a complete sentence.

This language is fine if state users are going to know what it means to explain the activity and I'll approve it regardless because this change is good! But one note to consider: we're asking to distinguish between summarizing in the first text box and explaining in this one. Is that distinction going to be clear enough, or do we need to offer more about what they're explaining?

jeromeleecms: @jameshupp If you have some suggestions for changing it, maybe we can bake it into #1027?

Issue #1049, opened 2018-10-15
Can't properly type percentages into the Quarterly break outs at the activity level


jeromeleecms:
When typing percentages into the quarterly breakdown at the activity level, the placeholder 0 persists, making it impossible to enter % < 10% or % that end with a number other than 0.

### This task is done when...
 - [x] User can enter any number in the quarterly breakout in the field
 - [x] Math checks out.
 
 ** Don't forget to fill out the LABEL and MILESTONE before submitting. If the issue is for the CURRENT SPRINT, add it to the PROJECT. If the issue is a BUG, FUTURE ENHANCEMENT, or OUTSIDE OF THE CURRENT SPRINT, do not add it to the PROJECT (then delete this line) **

mgwalker:
This bug applies to several of our formatted text boxes, like some of the dollar inputs.  Whatever the issue is, it's probably the same in all of them, so hopefully a one-time, across-the-board fix.

mgwalker:
The percentage one is especially weird.  There are a couple of things at play here, and it might be useful to have @lauraponce weigh in on how it should behave to make the user experience pleasant.

1. The input does not allow leading zeroes, so if you try to type anything behind the zero, the input will just reject it.  We can reconfigure this one really easily.  If we change this one, then typing behind the zero will give you the value you want.  E.g., if the box says `0%` and your cursor is after the 0, when you type `3`, the box will change to `3%`.

2. The value you see in the input is disconnected from the value that you type.  When you type a value, it gets sent to the activity reducer.  Then the budget reducer runs.  Then the value in the input is updated with the value from the budget reducer - what you type is not what gets put in the box.  The upshot is that the UX ***does not*** allow the user to clear out these boxes.  You can delete everything in them, but the reducers will convert that empty value into a `0`, and that's what the input box will display.

Fixing 1 is really easy and probably necessary.  Fixing 2 is harder, but I think we can make it happen, if that'd make the UX better.

lauraponce:
Is it possible to have the % and $ symbols in the input field, but not part of the actual input? For example: 

![formatting](https://user-images.githubusercontent.com/21222704/47188125-94c6bc80-d2eb-11e8-99ce-800cab059602.gif)

mgwalker:
I'm prototyping switching to a different library for adding the masks, and this new one seems to behave  a lot better.  It still leaves the $ and % in the input, but it's a lot smarter about how it handles the cursor and everything.

[...]

PR #1064, opened 2018-10-24
Fix number inputs bug

mgwalker:
Addresses #1049

Changes out the library we use to [react-number-format](https://www.npmjs.com/package/react-number-format), which seems to have a lot better handling around cursor location, etc.  Should fix the primary issue described in #1049.

Also makes the inputs only accept whole numbers.

- can set percentage fields as expected
- accept whole numbers throughout the app

- [ ] Product has approved the experience (with the acknowledgement that the PR deploy behaves strangely for Jerome, so there might be a little extra effort so he can make sure this works as expected).


web/src/components/Inputs/RichText.js
```diff
@@ -0,0 +1,58 @@
+import PropTypes from 'prop-types';
+import React, { Component } from 'react';
+import { Editor } from 'react-draft-wysiwyg';
```
hbillings: I had to go google this -- _oooh_

web/src/components/Inputs/InputWrapper.js
```diff
@@ -0,0 +1,47 @@
...
+    label,
+    type,
+    hideLabel,
+    wrapperClass,
+    ...rest
+  }) => (
+    <div className={wrapperClass || 'mb2'}>
+      <label
+        htmlFor={name}
+        className={hideLabel ? 'sr-only' : 'block mb-tiny truncate'}
```
hbillings: I kinda think we should move toward defining classes as variables, like:
```
const labelClasses = {
  hidden: 'sr-only',
  visible: 'block mb-tiny truncate',
}

...

className={hideLabel ? labelClasses.hidden : labelClasses.visible}
```

That'd let us find and edit classes easier. But it's an optimization and an opinion, and def not something that needs to happen now!

mgwalker:
I'll try to do a better job of describing where in the app changes apply!

Issue #1023, opened 2018-09-19
Formatting for help text in app is not working consistently


jeromeleecms:
During user testing, we discovered some formatting issues throughout the app where word wrapping was occurring too early.

![image](https://user-images.githubusercontent.com/34279968/45773689-c6bb0500-bbff-11e8-9cc2-a6e4dba92f15.png)

![image](https://user-images.githubusercontent.com/34279968/45773749-f9fd9400-bbff-11e8-8010-57fbea718790.png)

vs. expected normal:

![image](https://user-images.githubusercontent.com/34279968/45773908-709a9180-bc00-11e8-91fc-36269f55ce3a.png)

### This task is done when...
 - [x] Root cause of the bug is discovered
 - [x] Product/Design has reviewed app to ensure no other formatting issues have occurred.

hbillings:
It looks like the cause of this is the YAML formatting. Our renderer is treating line breaks within the YAML as places where line breaks should be. For example, the program overview example above looks like this in the YAML:
<img width="664" alt="screen shot 2018-10-19 at 10 48 48 am" src="https://user-images.githubusercontent.com/509309/47229091-7f559f00-d38c-11e8-92b7-5917a8002de9.png">

and renders this code:
```
<div class="mb2">
Provide a brief, high level summary of the state program including history,
<br>
and future vision. Include history and summary of future
<br>
vision of the state program.
</div>
```

I believe the easiest way to fix this is simply to remove the `>-` block [("chomping indicator")](https://yaml-multiline.info/) prior to the text. Someone with more YAML experience (@jameshupp?) might know better, though?

jameshupp:
Per discussion offline with @hbillings and @mgwalker sounds like the issues is the indentations at the start of new lines, as shown in that image. PR coming momentarily that should address this.

jameshupp:
After #1065 this is done.

PR #1065, opened 2018-10-24
Remove hanging indentations in copy from yaml

jameshupp:
This pull request addresses #1023

### This pull request changes… 
These indentations are leading to improper line breaks. Removing them lets the text break at its natural point instead.

 - [x] Someone other than the original author can verify that the line breaks are not occurring too early anymore.


mgwalker:
We should add a YAML linter/formatter to our build scripts.  🤔  @jameshupp what do you use to edit the YAML files?

Issue #976, opened 2018-08-23
Implement print preview functionality that has page breaks and hides helptext


NAretakis:
This issue is for the devs to implement whichever library/code is necessary to allow for export of an eAPD.

We need both PDF and Word exports. For Word, rich text format would probably be good enough for MVP.

NAretakis:
This issue is for actually implementing exporting functionality based on #974

NAretakis:
tagging in @hbillings so that we can implement the ability to hide "Helptext" and "Reminders" from the print preview display. By hiding these elements, we'll be left with a consolidated single APD view, that can be printed to PDF, and/or exported to docx (maybe).

NAretakis:
rescoped issue to represent the thinking of #974 

TL:DR:
Potential workflow is:

1. click the "print preview" button
2. click the browser's "print" button (we could also initiate this with some Javascript)
3. save/print to PDF

NAretakis:
closing this issue because #1160 is a better issue that has more detailed information.

Issue #1077, opened 2018-10-29
Clarify interaction of add/edit/remove item component


jenniferthibault:
The pattern of creating a list of one or many items with a title & description is one that happens in a few places of the app. These interactions are often nested, which adds complexity. At least two new members of the 18F team interacted with the component, and didn't  understand it's features or capabilities.

The interaction must allow the user to:
- enter at least one item
- add another item with details
- remove an item and its detail fields
- expand/collapse the details of an item you've added
- name an item, and see the title associated with the description (and see it adjust when changed)

Seen here: https://hitech-apd.app.cloud.gov/apd
![screen shot 2018-10-29 at 2 27 02 pm](https://user-images.githubusercontent.com/11636908/47671794-bf840100-db86-11e8-9429-36b5355da11b.png)

Identified challenges with the current design:
- there are two adjacent buttons for removing an item
- the title field looks and acts like a paragraph field, so it doesn't give users clues to what an appropriate input is
- it's not clear that you can collapse items
- visual grouping of elements is unclear

### This task is done when...
 - [ ] We have proposed a new design for the interaction
 - [ ] We have tested the interaction with users, 
   - [ ] who successfully use it to complete their tasks

jenniferthibault:
Looked at this with other designers, and leaving some notes here for our future selves:

- The purpose of the blue bar (where info is often recapped from what's entered) isn't clear. We should find out if it really serves a purpose (such as capturing info that's brought into tabular data elsewhere in the form) and clarify that, or possibly eliminate it
- Make the pattern consistent and predictable between instances where it's applied, so that all "Add" and "Remove" interactions feel the same, no matter what they're applied to
- Make it more clear what information is the unit of one add/remove instance


And found a few related examples for future inspiration. None of these are quite exactly like our context, but show a few possible different approaches:

![screen shot 2018-11-27 at 4 05 16 pm](https://user-images.githubusercontent.com/11636908/49118509-836dba00-f272-11e8-8046-669900c605c8.png)
<img width="740" alt="screen shot 2018-11-27 at 2 14 59 pm" src="https://user-images.githubusercontent.com/11636908/49118513-8799d780-f272-11e8-9f68-9b3dc1220f91.png">
<img width="749" alt="screen shot 2018-11-27 at 4 27 02 pm" src="https://user-images.githubusercontent.com/11636908/49118530-a4360f80-f272-11e8-9da5-9b02fd1703af.png">
[...]

PR #1078, opened 2018-10-29
Consolidate blue lists [skip deploy]

mgwalker:
There are five sections that use lists where the list items have blue headers.  This PR extracts that list into a reusable component and updates the five sections to use that component instead of each one defining the list themselves.

Instead of the list items being removed from the DOM when they are collapsed, they are now given a `display: none` style (class `d-none`).

- no visible changes
- makes it easier to update these later
- makes it possible to have these expanded for PDF creation (see #976)


web/src/components/CollapsibleList.js
```diff
@@ -0,0 +1,92 @@
...
+
+    return (
+      <Fragment>
+        <div className="mb1 h5 flex justify-between">
+          <button
+            type="button"
+            className={`btn btn-no-focus p1 flex-auto left-align bg-blue-light ${
+              deleteItem ? 'rounded-left' : 'rounded'
+            }`}
+            onClick={this.toggle}
```
hbillings: I feel like there should probably be ARIA attributes (`aria-controls` on the button and a corresponding `id` on the collapsible panel) on this somehow. But the structure of this may be changing in #1077, so it's probably not worth stressing over too much yet.

web/src/components/CollapsibleList.js
```diff
@@ -0,0 +1,92 @@
...
+            <button
+              type="button"
+              onClick={() => deleteItem()}
+              className="btn btn-no-focus p1 bg-blue-light rounded-right"
+            >
+              ✗
+            </button>
+          )}
+        </div>
+        <div className={childClass}>{children}</div>
```
hbillings: We should throw `aria-expanded={this.state.expanded}` on here. (In fact, if you want, you could even scope the styles to that using `[aria-expanded="false"] { display: none; }`, and obviate the need for a CSS class at all!)

web/src/components/CollapsibleList.js
```diff
@@ -0,0 +1,92 @@
...
+};
+
+ListItem.defaultProps = {
+  children: null,
+  deleteItem: null,
+  initialExpanded: false
+};
+
+const CollapsibleList = ({ content, deleteItem, getKey, header, items }) => (
+  <div>
```
hbillings: I don't know that much about when to use `Fragment`s, but is this a case where that might be handy? This `div` doesn't seem to do much except keep React from complaining about not having a container div.

web/src/components/CollapsibleList.js
mgwalker: Definitely yes on adding `aria-expanded`!  Also, I love the idea about using a selector on the attribute.  Could we still override it for the print preview mode?  Eg, `.print [aria-expanded="false"] { display: inherit; }`?  That's how it's working now, so we're able to enable the print mode without doing much in the code - just add a class to the top level element and that overrides display styles.

web/src/components/CollapsibleList.js
mgwalker: Good point.  Do you think we should make an issue just so we remember that the `aria-controls` is missing?

web/src/components/CollapsibleList.js
mgwalker: That's a good point.  We should probably come up with some rules for when to use one or the other.  I was thinking the `div` was a nice way to group the things in the DOM, but I don't know if that's actually useful _at all_.  😛  Certainly might be useful if we wanted to apply styles all the way down, I guess?  I think I'd suggest punting on changing anything until #1077 for this as well, but I think it's a good idea to talk about which we use, and when.  Maybe we could write it down.  In documentation somewhere.  🤔

PR #1098, opened 2018-11-06
Create scrolly tables for budgetary displays

### LINKED ISSUE #1037: Build out design updates to the table
Based on lots of suggestions and conversations with Aviva and Heather for #860, the updates to start with for the table are:

1. Adding a persistent horizontal scrollbar (currently it only appears after you start scrolling)
2. Making the leftmost column sticky as you scroll to the right
3. Un-fix the table cell widths and allow them to grow/shrink with the numbers in there

hbillings:
Closes #1037.

- Changes `overflow: auto` to `overflow-x: scroll` on all budgetary tables. This should have the effect of always showing scrollbars at the bottom of those tables.
- Removes the `min-width` from tables to make columns more able to size themselves. This means they'll be smaller when there's less in them. ~(@jenniferthibault @lauraponce, I'd love your eyes on the `Summary Budget Table`  -- the only table where this is currently implemented -- to see if you like this approach.)~ I left the tables with inputs as fixed tables to provide a consistent input size.
- Implements a frozen left pane on most scrolling tables. Notable exceptions: 
~1) Cost allocation (in Program Activities). The header cells on this table are special snowflakes, making it harder to implement properly, and the headers are so big relative to the content area that it seems like fixing the headers might make it awfully annoying to enter data.~ Screw it, I figured out the table structure, so I implemented this.
2) Executive summary. This seems like it mostly fits in a single window view.


jenniferthibault:
**Looking at the summary budget table:** the cells auto-sizing to the content inside them seems to work better than what it was previously — but do we have control to selectively assign a fixed width in addition to making it fixed position for horizontal scrolling? If so, in order to keep legibility higher, we should keep the left-most column fixed width. Examples 👇 

Table before data gets entered: good size left column! Readable and cozy!
screen shot 2018-11-07 at 7 10 27 pm

Table once data is entered, left column gets narrower and hard to read :(

screen shot 2018-11-07 at 7 09 57 pm


Annnnd in slack you asked if we should style the fixed-position left hand column differently while horizontal scrolling. I agree we should! Probably a slight drop-shadow on the right-hand edge. Glad to mock that up tomorrow

web/src/containers/BudgetSummary.js
```diff
@@ -156,12 +173,58 @@ HeaderRow.propTypes = {
...
+FrozenHeaderCol.propTypes = {
+  title: PropTypes.string.isRequired
+};
+
 const BudgetSummary = ({ activities, data, years }) => (
-  <div className="overflow-auto">
-    <table className="table-cms table-fixed" style={{ minWidth: 1000 }}>
+  <div className="table-frozen-col-wrapper">
+    <div className="table-frozen-col-scroll-window">
+      <table className="table-cms table-frozen-col-pane" aria-hidden="true">
```
hbillings: @mgwalker I'd like your take on this -- basically, I'm creating a secondary table that is just the "header" cells on the left. I don't love this approach, but nothing else I've tried works reliably cross-browser. I think I could abstract it out further than it currently is so that implementing scrolly tables across the app is easier, but I wanted to get your feedback before I go after :all-the-things:

mgwalker: Hmm.  My first reaction is "that's cool!"  But then I wonder how you make sure the header cells are the same height as the content cells in the same row if they're in a different table.  I guess we could explicitly set cell heights, which doesn't seem like it should be a problem?

An alternative could be to set all cell heights and also set the header cell _widths_, and then make the header cells absolutely positioned.  Playing with it in the devtools suggests it works but it might create some quirks with padding.

hbillings: So I'm making a couple of assumptions here: one is that the content in all of these tables is budgetary, and thus will only be one line. This seems to be true so far, but I don't know if it will always be in the future. By setting the width of that first column explicitly on both tables, I can make sure the tallest element (which is the first column's cells) is the same in each.

The other way to do this is to actually create a duplicate of the _entire_ table, then set `visibility: hidden` on all of the cells except the first column. That will size the columns exactly the same in both tables, but will also, y'know, create an entire duplicate table. 

So... there isn't a _great_ way to do this that keeps the table in semantic markup. If you want me to go the clone-the-whole-table route for stability, I'm cool with that. (There are some slicker solutions out there, but they all involve making a "table" out of divs, which is going to be pretty impossible for a screen reader to parse.)

hbillings:
@jenniferthibault Yes! We do have control over the size of that first column. Check out the latest version -- we can increase the width further if you like.

web/src/containers/BudgetSummary.js
mgwalker: Oh, that makes sense re: assuming the rows will be one line.  I think it's all clicking in my head now.

Gosh having a whole duplicate table sounds pretty rough.  I think I'd vote against it, but I _*am*_ curious - would having an invisible table over the top prevent scrolling?  (I assume the top table would not enable scrolling, in order to get the header column to be fixed, but 🤷‍♂️.)  Anyway...

Okay, I gotcha now.  I was just seeing a quirk locally where the content cells weren't as tall as the header cells, but it looks like that's going to get whipped.  I think this is a sensible approach.  Feels a lot less messy than a duplicate table, too.  😬

web/src/styles/app/table.css
```diff
@@ -79,3 +79,33 @@
...
+}
+
+.table-frozen-col-pane td,
+.table-frozen-col-pane th {
+  border-right: 2px solid #fff; /* only half of normal .table-cms spacing applied otherwise */
+}
+
+.table-frozen-col tr > th:first-child,
+.table-frozen-col tr > td:first-child {
+  min-width: 140px; /* same as table-frozen-col-pane; for some reason just width doesn't work */
```
jenniferthibault: This width looks good!

web/src/containers/BudgetSummary.js
jenniferthibault: With having a duplicate table, what happens to that fancy row that expands with the calculation inside it? I noticed that left in the recent changes that got pushed, but not sure if it's on purpose or an accidental casualty 😬 

screen shot 2018-11-08 at 12 40 30 pm

hbillings: The duplicate table approach wasn't one I'd seen before, and it kinda blew my mind. (yes, you can still scroll the underlying table! wat) Here's an example: https://codepen.io/paulobrien/pen/gWoVzN

I'd be interested in seeing that bug you saw locally, just to make sure I'm not wrong about any of my assumptions. 😬 But I'll polish this up, then, and ping you for review when it's done. Thanks for the sanity check!

hbillings: @jenniferthibault Oops, I missed this earlier -- the code to make this work is still in the table, but I couldn't find anywhere it had been used? @mgwalker, do you know?

hbillings: (also, I should note I've never seen said button in action -- I'd love it if someone could give me a demo!

mgwalker: You trigger it by clicking the teensy plus sign on the right side of the cell.  I'm actually not super familiar with how it works, but I notice it seems to be really slow and not amazingly formatted.  😛

It's the `<DataRowDetails>` component that's used inside the `<DataRow>` in the `BudgetSummary.js` file.

jenniferthibault: I missed it in my design audit (fixed now) because the pattern only appears _once you put data into the tables above_.

hbillings: ooooooh. Hrm. Okay. I'm going to have to think about this and play with it a bit.

hbillings:
@mgwalker @jenniferthibault I think I'm ready for review on this one -- there's one failing test I'm trying to debug, and there's one set of tables that, for reasons I can't seem to figure out, has an extra 2px of border or padding on the frozen left pane. (see below). I'm of the opinion that the latter issue shouldn't block merging, but I'm also happy to be convinced otherwise.

screen shot 2018-11-13 at 4 43 40 pm

jenniferthibault:
@hbillings last styling thought — 
Is it possible to add a slight shadow to the right side of the fixed column _only when the user starts scrolling?
In other words, the table would appear flat if it fits in the view width, or otherwise doesn't need to scroll. But once the user scrolls and the table overlaps, the shadow would appear like this?


screen shot 2018-11-14 at 9 48 33 am

I have this set with `box-shadow: 2px 0px 7px 0px #CCCCCC` but there's likely a better way to do it

hbillings:
@jenniferthibault How would you feel about breaking that out into a separate issue? I'm not quite sure how to handle the scroll event in all the disparate React components this touches, and I'm concerned that this PR is already pretty unwieldy. That said, if you think it affects the usability of the tables, I definitely don't want to ship this without it!

web/src/containers/ApdPreviousActivityTable.js
```diff
@@ -53,205 +53,247 @@ const ApdPreviousActivityTable = ({
...
+              <tr key={`${year}-fake-row`}>
+                <th key={`${year}-fake-header`}>
+                  {t('ffy', { year })}
                 </th>
-              </Fragment>
+                <td>
+                  <DollarInput
+                    hideLabel
+                    wrapperClass="m0"
+                    className="fake-spacer-input m0 input input-condensed mono right-align"
```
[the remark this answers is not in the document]
hbillings: Whoops, you're right! I thought I was going to key off of that for tests, but I ended up using the `name` instead. I'll remove from the class list.

web/src/containers/ApdPreviousActivityTable.js
```diff
@@ -53,205 +53,247 @@ const ApdPreviousActivityTable = ({
...
-            <th style={{ width: 90 }} id="prev_act_hit_header_null1" />
-            {Object.values(programs).map((name, i) => (
-              <th
-                key={name}
-                colSpan={4}
-                className={`bg-${colors[i]} center`}
-                id={`prev_act_hit_${name.toLowerCase()}`}
-              >
-                {name}
+    <div className="table-frozen-wrapper table-frozen-narrow-header">
```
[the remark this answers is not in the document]
hbillings: I'm terrified that this means we'll work our way up to galactic CSS at some point, and no one will be able to understand it.

hbillings:
@mgwalker Yeah, stuff like this gets a little hairy 😅. I'd _love_ to abstract this out into its own `Table` component, but since each table is just a little different, that seemed like a pretty heavy lift. :-/ What I'd really like to do is add a `table-frozen` example to Storybook, if possible, so it's easier for future folks to see what styles are applied where, and how to properly fit the pieces together, without having to parse CSS and a complex, React-generated table all at once. (I'm specifically thinking of [this set of card layout examples from CALC](https://calc.gsa.gov/styleguide/#card-layout), and how having them in our component guide made it way easier to reference.)

Atomic CSS, as I understand it, is trying to solve two problems: one is explicitness of selectors, as you've just pointed out. The other is minimizing the effect of the cascade (i.e., it's easy to have parent elements with lots of styles applied to a single class that might affect their children in ways you don't want). As I'm looking at this, I'm realizing that many of these selectors are, in fact, atomic (there's only one CSS rule applied to most of them). They just aren't named that way -- they're named in a loosely [modified BEM](http://getbem.com/naming/) (block-element-modifier) model. 

In this case, it makes sense to me to use BEM because these elements all have to be applied in a specific order to get the desired behavior, and naming them in that way indicates you can't use them willy nilly. It also means you just have to apply a few classes to parent elements that will have cascading effects on their children, rather than having to remember, say, that you need to apply `position-absolute` and `width-140px` classes to the third div in the order. That said, I'm happy to try to rename these classes to match our existing atomic classnames if you'd like! (Or push up a commit where you could see an example of how that would look, if that would help.) 

As an aside, just because it's interesting (to me, anyway), I think this is an instance where using Sass would also make things more readable, because instead of the current flat structure, you could imply the relationship of elements to each other. As an example, this:

```
.table-frozen-wrapper {
  position: relative;
}

.table-frozen-scroller {
  overflow-x: scroll;
  overflow-y: visible;
}

.table-frozen-left-pane {
  position: absolute;
  max-width: 140px; /* no larger than visible element */
}

.table-frozen-left-pane td,
.table-frozen-left-pane th {
  border-right: 4px solid #fff;
  vertical-align: inherit;
}
```
becomes this:
```
.table-frozen-wrapper {
  position: relative;

  .table-frozen-scroller {
    overflow-x: scroll;
    overflow-y: visible;

    .table-frozen-left-pane {
      position: absolute;
      max-width: 140px; /* no larger than visible element */

      td,
      th {
        border-right: 4px solid #fff;
        vertical-align: inherit;
      }
    }
  }
}
```

The downside is that if you do this everywhere, you can really blow out the size of your CSS file. The upside is that some complicated elements like this can benefit from the extra self-documentation. Although it's possible that's only easier to understand for someone whose brain is already warped by Sass. ^_^

PR #1105, opened 2018-11-08
Include expenses text in the app

### LINKED ISSUE #1101: Removing all non-personnel costs throws an error instead of helptext
As seen in the screenshot below

image

 
mgwalker:
Fixes #1101 

The file that contained non-personnel expense text wasn't being included in the app at all.  Other places that use the text are defensive and just don't display if there's no text (that way we can have `<HelpText>` in the code, but if there's no text, it just gets ignored; gives content design some flexibility; probably not a great pattern in hindsight), but this one isn't.  Rather than make it defensive, this PR links in the file so it gets displayed.

- when there are no non-personnel costs listed, shows a message rather than an error


mgwalker:
CC: @jameshupp.  This fix results in description and help text in the activity / in-house costs / non-personnel section.  The text has always been in the YAML, but it wasn't displayed in the app until this PR.  😳

PR #1112, opened 2018-11-09
Update endpoint tests [skip deploy]

### LINKED ISSUE #1085: Technical debt: backend end-to-end tests
Our end-to-end backend tests are pretty thorough and have helped us manage a pretty major refactor in the past.  However, because of the way they're written, they're pretty high burden to maintain.  All of the expected results are written directly into the test code, which means any time we change how we return data, we also have to wade through a bunch of test code to find the right spot to update.

The risk is that when tests are hard to update, they will often not be updated.  That might mean that endpoint tests are turned off "just for now" in order to get a feature out the door, or a specific test is "temporarily" disabled.  That's a slippery slope, and soon you end up just not having end-to-end tests.

One possible approach to making the tests easier to maintain is to switch from the [tap]() testrunner to [Jest]() and make use of Jest's snapshot comparison functionality.  With snapshots, the expected results are created and stored by Jest, separately from the test code itself.  When the actual results don't match the expected results, Jest can show you exactly what changed so you can easily see whether the new results are what you expected, and Jest can also automatically update the snapshot so you don't spend a lot of time manually digging through code.

**Note** that this isn't currently causing any problems and isn't likely to get worse based on the short term roadmap.  It would be nice to get this cleaned up before handing off development to another team, so maybe it would make sense to put a little bit of time into this one when there's a spare hour here or there or as @hbillings suggested, maybe during a "feature freeze" at some point during the ATO process.

### This task is done when...
 - [ ] end-to-end tests are easier to create and maintain

mgwalker:
Closes #1085 

Switches backend tests to use snapshots instead of having to manually build every piece of data we expect to get back.  This will speed up development dramatically by letting tools check the snapshots, show us exactly what changed, and let us quickly decide whether to accept the change or go back and fix the code more.

- no visible changes


api/.eslintrc.json
```diff
@@ -6,5 +6,13 @@
...
-  }
+  },
+  "overrides": [
+    {
+      "files": ["**/*.test.js", "**/*.endpoint.js"],
+      "env": {
+        "jest": true
+      }
+    }
+  ]
```
mgwalker: Sets up eslint so it doesn't complain about Jest's globals in our test files.

api/.snyk
```diff
@@ -0,0 +1,215 @@
...
+    - jest > jest-cli > yargs > os-locale > mem:
+        reason: vulnerable through test library; no data impact
+        expires: '2018-12-27T18:16:49.971Z'
+    - jest > jest-cli > jest-runtime > yargs > os-locale > mem:
+        reason: vulnerable through test library; no data impact
+        expires: '2018-12-27T18:16:49.971Z'
+    - jest > jest-cli > jest-runner > jest-runtime > yargs > os-locale > mem:
+        reason: vulnerable through test library; no data impact
+        expires: '2018-12-27T18:16:49.971Z'
+patch: {}
```
mgwalker: Basically tells Snyk to hush about these vulnerabilities for 30 days.  We can reevaluate then, but since they come via Jest I'm not *too* concerned about them.  They might be an argument for running tests in Docker for a while, though.

api/endpoint-tests/endpoint-coverage.js
```diff
@@ -9,6 +9,17 @@ const readline = require('readline');
...
+  let last = url.replace(/\/:(.+?)(\/|$)/g, '/{$1}$2');
+  let out = last.replace(/\/:(.+?)(\/|$)/g, '/{$1}$2');
+
+  while (out !== last) {
+    last = out;
+    out = last.replace(/\/:(.+?)(\/|$)/g, '/{$1}$2');
+  }
+  return out;
+};
+
```
mgwalker: The previous regex was only catching the first `/:variable:/` in a URL, so something like `/:var1:/:var2:` wasn't getting caught.  The endpoint "coverage" report was messed up, as a result.

api/endpoint-tests/endpoint.sh
```diff
@@ -6,7 +6,7 @@ rm ./endpoint-tests/endpoints-hit.txt
 docker-compose -f docker-compose.endpoint-tests.yml up -d
 docker-compose -f docker-compose.endpoint-tests.yml exec api-for-testing npm run migrate
 docker-compose -f docker-compose.endpoint-tests.yml exec api-for-testing npm run seed
-docker-compose -f docker-compose.endpoint-tests.yml exec api-for-testing npm run test-endpoints
+docker-compose -f docker-compose.endpoint-tests.yml exec api-for-testing npm run test-endpoints $@
```
mgwalker: This passes along arguments from the command line into the actual test.  That way you can update snapshots using the helper script:

```shell
./endpoint-tests/endpoint.sh -- -u
```

api/package.json
```diff
@@ -10,8 +10,9 @@
     "seed": "knex seed:run",
     "start": "node main.js",
     "start-dev": "nodemon ./main.js",
+    "jest": "NODE_ENV=test jest",
     "test": "NODE_ENV=test tap -J --cov --coverage-report=lcov --no-browser --reporter=spec --test-arg=--silent '{,!(node_modules)/**/}*.test.js'",
-    "test-endpoints": "tap --reporter=spec 'endpoint-tests/*/**/*.js'",
+    "test-endpoints": "jest --runInBand '.+\\.endpoint\\.js'",
```
mgwalker: We need to run the tests in serial (or "in band," as Jest calls it) so tests aren't modifying the database at the same time.  We want the database in a known state for each test, so to get that, we need them running individually.  Makes things a little slower, but not a lot.

api/routes/apds/get.endpoint.js
```diff
@@ -0,0 +1,40 @@
...
+  login,
+  request,
+  unauthenticatedTest,
+  unauthorizedTest
+} = require('../../utils.endpoint');
+
+describe('APD endpoint | GET /apds', async () => {
+  const db = getDB();
+  beforeAll(() => db.seed.run());
+  afterAll(() => db.destroy());
```
mgwalker: One big downside to switching to Jest is having to add this to every set of tests that modify the database.  `tap` has an event that fires when tests are finished, and you can tap into that to destroy the database connection - if you don't destroy the connection, the test process hangs and eventually the test is failed due to timeout.  Jest doesn't have that, so we have to hook the `afterAll` teardown method.

api/seeds/test/apds.js
```diff
@@ -21,58 +21,1576 @@ exports.seed = async knex => {
       status: 'not draft'
     }
   ]);
-
-  await knex('activities').insert([
+  await knex('apd_incentive_payments').insert([
```
mgwalker: This seed is basically modified to inject Nick's APD into the test data.  Ideally the test data includes at least one fully-formed APD so we can make assertions about the structure of the data returned by the API.  This should accomplish that.

PR #1132, opened 2018-11-16
Remove "example" document items from the state dashboard

mgwalker:
Just what it says on the tin.  Those things are confusing and shouldn't be there - they were only ever meant to demo what the various status/progress things might look like.  Now that we've moved on from that, they're just noise!

- state dashboard should only show active APDs in the "Current Statuses" section


mgwalker:
Considering that this just removes stuff we've already agreed should be removed, I'm going to go ahead and merge it

Issue #1088, opened 2018-11-02
Instruction text consistency


mgwalker:
Use this structure for instructional text in the yaml and throughout the app:

```yaml
instruction:
  heading:(h3) BLAH BLAH!
  short: (bold) blah blah
  detail:(regular text)  >-
    blah blah blah blah blah
    blah blah blah blah blah
    blah blah blah blah blah    
  helpText: (thing that appears in blue box)
```

We need to preserve the option in the code to switch the order of short/detail to accommodate the standards and conditions section.

### This task is done when...
 - [ ] all the instructional text in the yaml files has been reformatted
 - [ ] the app code uses the new format throughout

jeromeleecms:
What's best practice for documenting this format for future generations? Is this documented in comments within the YAML file, go into Technical documentation, deserve its own page in the future style guide, etc...?

mgwalker:
One cool thing about YAML is that we ***can*** have comments.  I think some kind of standalone thing makes more sense, though.  A place where we can list the textual components that the app is aware of and the structure it expects from the YAML, for example.  🤔

http://storybook.hitech.eapd.cms.gov.s3-website.us-east-2.amazonaws.com/

mgwalker:
I'm curious about the logistics of making this change, because it seems like it could need considerable coordination between content and dev, since changing the keys in the YAML files will necessitate code changes.

I think the biggest two pieces for me are:
- renaming the `reminder` key to `helpText` (which doesn't actually feel all that problematic, except...)
- what to do with everything that's currently stored as `helpText`

My sense is that once we have a good strategy for how to tackle those two, dev can hop into code and make a lot of it happen, with some occasional pairing with design to iron out some weird edge cases.  Then, once the app is prepared to handle all the new content components, it should be easier for @jameshupp to cleanup any booboos we introduced and make everything just right.

mgwalker:
@jameshupp I didn't mean to make any suggestions.  I don't have any reservations about you making structural changes as long as you loop in dev so we can do the associated code changes, especially around this issue since we've already talked about it.  I just wanted to remark that this issue hints at a fairly significant amount of coordination!  😄

jeromeleecms:
Is the current state of the YAML file good enough for the MVP? Or does this revised formatting provide some real solid (aka save time down the road and/or other benefits) for the MVP?

jameshupp:
[...]

PR #1134, opened 2018-11-16
[WIP] Instruction text prototype demo

mgwalker:
Related to #1088.  Demo of how the YAML file might be structured with the proposal from that issue.  Also shows the level of code change involved in addition to the YAML changes.  (Pretty light.)

With this change, the YAML for all of our *subsections* could be switched over to the new `instruction` format - a subsection is inside the collapsible white boxes.  By adding an `instruction` section to the appropriate place in the YAML, the subsection will switch over to using that instead of the old `helpText`/`reminder` - but only for the topmost content.  If there's more than one block of instruction text in the subsection, we'll need to go update the code for those.

The program summary -> program overview subsection YAML was updated to the `instruction` format.  Also, the code was modified to expect the `instruction` format for the other instruction blocks within the subsection.

cc: @jameshupp.  We can pair up sometime to make sure this actually fits with what you intended and makes sense.

- switches some components to use new `instruction` YAML format
- tests intentionally left failing so we don't accidentally merge it

### DIFF
--- a/web/src/components/Instruction.js
+++ b/web/src/components/Instruction.js
@@ -4,12 +4,7 @@ import React from 'react';
 import Md from '../components/Md';
 import { t } from '../i18n';
 
-const HelpText = ({ children }) => (
-  <Md content={children} wrapper="p" className="mb2 text-s alert alert-info" />
-);
-HelpText.propTypes = {
-  children: PropTypes.node.isRequired
-};
+const md = (content, props) => <Md content={content} wrapper="p" {...props} />;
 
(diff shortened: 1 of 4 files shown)


mgwalker:
Closing.  See [the storybook](http://storybook.hitech.eapd.cms.gov.s3-website.us-east-2.amazonaws.com/?selectedKind=Texty%20components&selectedStory=Instruction&full=0&addons=1&stories=1&panelRight=0&addonPanel=storybook%2Factions%2Factions-panel) for the demo instead.

Issue #1117, opened 2018-11-10
Standardize vertical and horizontal spacing increments


jenniferthibault:
Carried forward from #1082 

In order to create more consistency and reduce the number of style decisions our team has to make with each new implementation, we should define a limited set of spacing increments, and apply them to headings or elements by default. This should reduce the number of utility classes in the markup, but will also allow us to override them when needed through the BASCSS styles.

### This task is done when...
 - [ ] Figure out what our spacing increments should be (and what elements they're used in pattern with)
 - [ ] Overwrite BASCSS margin & padding variables with the increments we want
 - [ ] Tweak how & where margin & padding variables are used to make them match design spec
 - [ ] Identify which patterns need to incorporate spacing elements in their default state and build them so!
- [ ] (Probably separate issue) Replace those elements in the code with the patterns that have spacing incorporated

NAretakis:
OBSOLETE DUE TO CMS DESIGN SYSTEM

PR #1138, opened 2018-11-17
Size textareas and inputs according to their expected input

### LINKED ISSUE #1050: Build out the text box sizing updates
Creating this from design issue #1030 

There will be four sizes for the text boxes where state users input content: extra small, small, medium, and large.

All of the text boxes should have a text area max width of 652px and the heights are:

Extra small: 100px
Small: 200px
Medium: 400px
Large: 800px 

Each of the boxes will be scrollable if the amount of content exceeds the space in the box.

For the sections in the app, here are the text box size designations:

**Program summary**
- Program intro - large
- HIT overview - medium
- HIE - medium
- MMIS - medium

**Results of previous activities**
- Previous activities summary - medium

**Program activities**
- Activity overview - extra small
- Activity description - large
- Statement of alternative considerations - medium
- Performance measures and benchmarks - goals - small
- Roadmap - small
- In-house cost - description - small
- Non-personnel - description - small
- Private contractor costs - description - medium
- Cost allocation methodology description - medium
- Other funding description - medium
- Standards and conditions - small

**Assurances and compliance**
- Federal citations - extra small

hbillings:
Closes #1050.

There was one area mentioned in the original issue that I couldn't find: `Roadmap` in `Program activities`.  If I'm missing something, please let me know!

I also think that the extra large size might be too big with the sticky headers. We might not even need a size larger than, er, large?

For beer and skittles, I went ahead and added the `resize: vertical` capability to the rich text editor boxes.

Mmm, beer...

- The sizes of the textareas
- Makes the RTE boxes resizable on drag.


web/src/components/Inputs/RichText.js
```diff
@@ -40,6 +40,7 @@ class RichText extends Component {
         editorState={this.state.content}
         onEditorStateChange={this.onEditorChange}
         onBlur={this.onBlur}
+        editorClassName={this.props.editorClassName || 'rte-textarea-m'}
```
hbillings: `editorClassName` is something that [React Draft WYSIWYG expects](https://jpuri.github.io/react-draft-wysiwyg/#/docs?_k=jjqinp), but we weren't previously using it.

web/src/containers/AssurancesAndCompliance.js
```diff
@@ -96,6 +96,8 @@ class AssurancesAndCompliance extends Component {
                           hideLabel
                           value={explanation}
                           onChange={this.handleExplanationChange(name, index)}
+                          className="m0 textarea textarea-sm"
```
hbillings: I don't entirely understand [higher-order components in React](https://reactjs.org/docs/higher-order-components.html), but we're using them to render textareas. Somehow textareas seem to magically get `m0 textarea` classes, and when I added `textarea-sm`, it overwrote those classes. So I feel like I'm probably duplicating these from somewhere, but I can't figure out where they're coming from.

web/src/containers/activity/Description.js
```diff
@@ -22,8 +22,8 @@ const Description = props => {
             name={`activity-summary-${activity.key}`}
             label="activity summary"
             hideLabel
-            className="m0 textarea"
-            rows="5"
+            className="m0 textarea textarea-sm"
+            rows="3"
```
hbillings: I tried to add the approximate number of rows equivalent to the CSS-specified height, just because it's something the HTML can render on its own so it's extra-bulletproof.

web/src/styles/app/form.css
```diff
@@ -33,3 +33,33 @@
 .select-condensed {
   height: 2rem;
 }
+
+.rdw-editor-main .DraftEditor-root {
+  min-height: unset; /* override of plugin */
```
hbillings: The plugin CSS specified  a `min-height` of `240px`, and our smallest possible textarea is `100px`, so I'm unsetting the `min-height` to prevent weird gaps.

web/src/styles/app/form.css
```diff
@@ -33,3 +33,33 @@
...
+}
+
+/* add resizing capability to divs of react-draft-wysiwyg plugin */
+.public-DraftEditor-content {
+  resize: vertical;
```
hbillings: This gives us [drag-to-resize capability](https://developer.mozilla.org/en-US/docs/Web/CSS/resize) on the divs rendered by our RTE plugin. I didn't even know you could do this with pure CSS!

web/src/styles/app/form.css
```diff
@@ -33,3 +33,33 @@
...
+}
+
+/* rte-* classes are for react-draft-wysiwyg plugin */
```
hbillings: Because the RTE plugin only allows you to specify a class for one of the parent divs that wraps the actual "textarea" div, I had to target the RTE "textareas" slightly more ham-handedly than regular `textarea`s.

web/src/containers/AssurancesAndCompliance.js
mgwalker: The `mo textarea` classes comes from `components/Inputs/index.js`.  It then gets wrapped with an `InputWrapper`, which applies a label around the input.  The wrapper also takes every prop it doesn't know and applies them directly to the input element.  In this case, the wrapper doesn't know about `className`, so it just passes it along, thus overriding the class set initially.

Is it useful to set the `m0 textarea` class on every textarea, or would it be better to target them with CSS?  Then we could ditch the hard-coded `m0 textarea` and using the `className` prop would work as expected?

web/src/styles/app/form.css
mgwalker: Huh, I've never seen this `unset` value before.  Neat!  Do you know if it works in all browsers/is supported by postcss?  (It looks like the generated CSS still uses `unset` but I'm not sure.

We probably ought to adjust our postcss config to always support IE11, anyway.  Right now if IE11 drops below 2%, our config would stop trying to target it.  😬

hbillings: I went down a rabbit hole with this. [MDN](https://developer.mozilla.org/en-US/docs/Web/CSS/min-height#Syntax) lists `unset` as a global value, so I assumed it was kosher. [Caniuse](https://caniuse.com/#feat=css-unset-value), though, specifies that it's not supported in IE11. :-/ So... bummer, and good catch! I'll update it to use `initial` instead.

web/src/containers/AssurancesAndCompliance.js
hbillings: Hm, so, I think we're going to be revisiting what spacing units we use in #1117, so maybe it makes sense to deal with this then? (Applying the spacing class to every element is the correct Atomic Way to do this, I believe, but this might be one of the instances in which we want to use fewer classes with more rules each in order to consistently apply spacing, depending on what we figure out in that issue.)

jenniferthibault:
@hbillings I agree with you —  with the sticky header 600px is way huge, and 400 already feels plenty big to indicate a that more context is expected. @lauraponce are you good with removing the Extra Large size too?

I _think_ the roadmap field  could be this one, because it looks like the goal text should be very short. @lauraponce another question for you :) — Is that the case? And if so, would you be ok with removing the rich text editor from this field and making it size Small?

screen shot 2018-11-19 at 4 28 14 pm

lauraponce:
I'll defer to your recommendations on this, @hbillings and @jenniferthibault! I don't remember talks of an extra large size, and it's all relative anyway so maybe it doesn't matter, but yes getting rid of the XL sounds good 👍. I believe the goals and benchmarks sections were marked to be small without the RTE in #1050, so agreed that that should be the case. (Ignore the pixel sizes in #1050 though because those are clearly not right, but the relative size difference of the sections should still be pretty close to accurate).

jenniferthibault:
Ahhhh sorry this did get kind of convoluted. 

Here's where I thought we landed:
After ruling out an XL in this issue, we should have three sizes left from [where we started in the design issue](https://github.com/18F/cms-hitech-apd/issues/1050#issuecomment-436717961):
- Small: 100px (6.25rem)
- Medium: 200px (12.5rem)
- Large: 400px (25rem)

(And from slack, it looks like the thing we were formerly calling `Roadmap` is the second text box in the Performance Goals and Benchmarks section) 

Does that line up with your understanding @hbillings ?

mgwalker:
➕ to making a discussion about which boxes are RTEs into its own issue.

hbillings:
@jenniferthibault That lines up with what I'm thinking! Thanks for spelling it out. :)

hbillings:
Just opened #1166 for the text editor discussion!

jeromeleecms:
@jameshupp https://en.wikipedia.org/wiki/THX_1138

PR #1144, opened 2018-11-19
[WIP] Consistent date-picker behavior

### LINKED ISSUE #959: Implement Date Picker Functionality Consistently Through Experience
Currently the date picker in the APD creation experience is applied using different settings (modal with blurred background w/ limited keyboard support and as a popup with keyboard support). The preferred implementation is to have the latter (popup with keyboard support).

As implemented under Contract Resources (default setting per http://airbnb.io/react-dates)

image

### This task is done when...
 - [ ] criteria 1
 - [ ] criteria 2
 
mgwalker:
Closes #959

- removes the modal date picker view, which introduces a bug I can't figure out (pinging @hbillings; details below)
- allows typing dates without leading zeroes on months or dates

### @hbillings:

The date picker gets positioned absolutely when it's opened, but for reasons I don't understand, despite being absolutely positioned, it's making its parent taller - but not necessarily tall enough to actually contain the picker, so a bit gets clipped off.  The right side also gets clipped off, because the parent doesn't get wider - only taller.

screen shot 2018-11-19 at 1 44 52 pm

You can see this behavior by opening an activity, going to the Activity Schedule section, and clicking in one of the milestone "planned end date" inputs.  You may have to scroll down to see it. (I don't know why its container has `overflow: overflow-auto` either...)


hbillings:
@mgwalker The `overflow-auto` is what's messing with this! If you remove it [from here](https://github.com/18F/cms-hitech-apd/blob/master/web/src/containers/activity/Schedule.js#L56), I think everything should work as expected. (What I _don't_ understand is why this would have `overflow-auto` on it in the first place. That seem pretty odd.)

mgwalker:
Huh.  No idea.  I traced that line back to when it was created, and it's just...  always been there.  🤷‍♂️  Removed it, and now I guess this one is ready for review.

hbillings:
I'm not sure why these tests aren't passing now, hrm...

Issue #831, opened 2018-07-20
Client Side Validation for MVP


jeromeleecms:
What still needs to be done for Client Side Validation to Complete MVP

@mgwalker to update this or add additional issues where needed, please!

Reference #330 

### This task is done when...
 - [ ] criteria 1
 - [ ] criteria 2
 
 ** don't forget to fill out the LABEL, MILESTONE, and PROJECT sections before submitting (then delete this line) **

mgwalker:
One thing we'd need is to walk through all the fields and figure out which ones need validation and what constitutes valid data.

@hbillings: We have a validating input component at `components/InputsWithValidation`, but it only does local validation (and we're not using it anywhere 🎉).  Should we look into hooking that up to redux somewhere to track which fields are invalid so we can have a rollup view somewhere (like when they try to save it or something)?

@lauraponce: Bren posted a [demo video](https://cl.ly/t8LW) of the validating input field he built.  Do you like the way it behaves?  It looks like it responds to every change after the first blur but I'm not 100% sure.

lauraponce:
@mgwalker For the error message, it is possible to reset it (or just not display the error) when they click back into it and the field is active? In this example, that's when it says Name is required. That way the validation would just happen on blur and display any errors at that point, but if they click into it presumably they are fixing any errors.

Also as far as what we're checking for, maybe just check that it's not blank since character count would be difficult since people can have one letter names? (Though I do like Bren's causal error message 😄 )

hbillings:
Hmm, my first instinct is to just use [HTML5 form validation](https://pageclip.co/blog/2018-02-20-you-should-use-html5-form-validation.html) if we want pure client-side validation, but I also don't have the best grasp of what we're trying to do here. Is there a list of things that need to be validated and when the validation should occur somewhere?

NAretakis:
For @hbillings and @mgwalker 's consideration. I think this can basically be tabled (and closed out) for MVP, because we are going to have federal users doing manual validation (during review) for each field for the MVP. That said, down the road we might want to come back to this, but right now Jerome and I talked it over and think it might be safe to close out. Do y'all see any issues with this approach?

mgwalker:
So far the app doesn't crash when users input weird stuff (as far as we know), so...  should be fine from a technical perspective, I think.

NAretakis:
[...]

PR #1146, opened 2018-11-19
Update number inputs

### LINKED ISSUE #1141: Implement consistent number formatting for fields with numbers
We have lots of fields (tables) that require an input of numbers. We'd like this to be a consistent experience, which formats using commas, dollar signs, and percentage signs. Dollars need to be rounded to the nearest dollar, and we need to be able to accommodate 14 characters ($1,000,000,000). Referencing #1049 for where this work originated.

47188125-94c6bc80-d2eb-11e8-99ce-800cab059602

mgwalker:
Closes #1141

- adds commas to dollar inputs; these are automatic - manual commas are still silently deleted
  screen shot 2018-11-19 at 2 36 20 pm

- allows percent inputs to include 2 digits after the decimal point
  screen shot 2018-11-19 at 2 36 02 pm

### DIFF
--- a/web/src/stories/index.js
+++ b/web/src/stories/index.js
@@ -1,10 +1,10 @@
-import React from 'react';
+import React, { Component } from 'react';
 import { storiesOf } from '@storybook/react'; // eslint-disable-line import/no-extraneous-dependencies
 
 import { DocumentItem, ProgressDot, ProgressLine } from './DashboardStory';
 import InstructionStory from './InstructionStory';
 import Btn from '../components/Btn';
-import { Input, Textarea } from '../components/Inputs';
+import { DollarInput, Input, Textarea } from '../components/Inputs';
 
 import '../styles/index.css';
 
@@ -32,3 +32,30 @@ storiesOf('Buttons', module).add('various', () => (
     <Btn kind="outline" /> <Btn kind="outline" size="big" />
   </div>
 ));
+
+storiesOf('Numeric inputs', module).add('Dollars', () => {
+  class Dollar extends Component {
+    state = { value: 0 };
+
+    change = e => {
+      console.log(e.target.value);
+      this.setState({ value: e.target.value });
+    };
+
+    render() {
+      return (
+        <DollarInput
+          hideLabel
+          wrapperClass="m0"
+          className="fake-spacer-input m0 input input-condensed mono right-align"
+          label="fake-spacer-input"
+          name="fake-spacer-input"
+          value={this.state.value}
+          onChange={this.change}
+        />
+      );
+    }
+  }
+
+  return <Dollar />;
+});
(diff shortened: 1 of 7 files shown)


mgwalker:
I'd probably advocate punting on doing checks on the percentages without doing validation everywhere else.  We had #831 but it got closed (presumably it's layer 2 or higher?)

lauraponce:
That makes sense, it's not a super high priority, more of a nice to have to minimize errors I would say. Should I create a new issue for that and tag it with future enhancement so we don't lose sight of this down the line, and then I could sign off on this PR?

web/src/stories/index.js
```diff
@@ -32,3 +32,30 @@ storiesOf('Buttons', module).add('various', () => (
...
+      console.log(e.target.value);
+      this.setState({ value: e.target.value });
+    };
+
+    render() {
+      return (
+        <DollarInput
+          hideLabel
+          wrapperClass="m0"
+          className="fake-spacer-input m0 input input-condensed mono right-align"
```
hbillings: Oh, shoot, I forgot to remove the `fake-spacer-input` from the classlist when I merged the table PR. There isn't an associated class, so you can drop it from here.

web/src/stories/index.js
```diff
@@ -32,3 +32,30 @@ storiesOf('Buttons', module).add('various', () => (
     <Btn kind="outline" /> <Btn kind="outline" size="big" />
   </div>
 ));
+
+storiesOf('Numeric inputs', module).add('Dollars', () => {
+  class Dollar extends Component {
+    state = { value: 0 };
```
hbillings: I'm a little confused about this pattern. I'm used to seeing component state initialized in a constructor -- can you help me understand why this is different?

NAretakis:
There’s probably a good amount of data validation that we’ll want to have, but it’s not Layer 1. A future enhancement issue for this would be good, and we can figure out what layer to bake it into in the future.

web/src/stories/index.js
mgwalker: Oops, I didn't even mean to include this.  😆 I just threw it in so I could modify and test them faster than reloading the whole app.  But maybe it's worth keeping, in which case I'll clean it up.

But to your question, it's just less code.  AFAIK it works the same, it's just more declarative.  Putting the state declaration in the constructor is one extra hoop you have to jump through to understand the structure of the class.  It's a [stage-3 proposal](https://github.com/tc39/proposal-private-methods) so it should make it into Javascript officially in the near future.

NAretakis:
@mgwalker looking at the build in a few sections, I realized that having two decimal places for percentages seems weird (since everything else is rounded to nearest whole). Can we remove the decimals, and just use whole numbers, i.e. "78%"?

mgwalker:
Yep, we can do that.  Back to the old way!

NAretakis:
Sorry for the rework ☹

PR #1158, opened 2018-11-20
Add key personnel costs

### LINKED ISSUE #1128: "Direct Attributable to the Project" Button doesn't do anything
When entering key personnel, when identifying and individual as attached to the project should add their cost into the total request.  This part of task #884. 

image

For now, should show up as HIT personnel. Project State staff. If the state hires contractors to provide oversight on multiple projects, we should have them include that under the program administration (Activity 1). 

image

### This task is done when...
 - [ ] There is a representation of the Key Personnel through out the budget tables within the app
 - [ ] Design has done a pass of how this will be intergrated through out the app
- [ ] A dev issue has been spun up to support
 
mgwalker:
Closes #1128

When a key personnel has costs attributable to the project, those costs need to be included in the budget tables.  This PR updates the budget reducer to capture those costs in these sections:

- HIT state personnel cost
- HIT total cost
- HIT + HIE state personnel cost
- HIT + HIE total
- federal share by quarter (using the quarterly breakout from the "Program Administration" activity)
  - cost due to state in-house costs per quarter
  - totals of ☝️ those for each FFY
  - total federal share
- total cost


NAretakis:
Neato! It all works, but in order for it to work, there has to be some stuff populated in the Activity 1 (Program administration). I'm good with this as it sits, because Program Administration is always going to be populated with something. However, testing might eventually yield some adjustments.

NAretakis:
I think this is a thing that we need some research and design help with (in a later issue and PR). Because we've consistently struggled with how this is positioned.

web/src/reducers/budget.js
```diff
@@ -497,6 +497,89 @@ const buildBudget = bigState => {
...
+
+        // Compute how much of the key personnel cost is attributable to each
+        // quarter of this FFY
+        const quarterlyCosts = roundedPercents(
+          federalShare,
+          percentPerQuarter[year]
+        );
+
+        const ffyQuarterly = newState.federalShareByFFYQuarter.hitAndHie[year];
+        const ffyTotal = newState.federalShareByFFYQuarter.hitAndHie.total;
```
hbillings: The `newState` / `bigState` terminology is a little opaque to me -- I see `newState` is built from `bigState` in the reducer, but I'm having a bit of a hard time following. (I say this just because while this code seems fine in isolation, I'm not able to make sure it makes sense in context.)

mgwalker: That's a good point, and the distinction is only ever implied.  The `UPDATE_BUDGET` reducer expects the entire redux state passed in with the action.  It then passes that full state object into this method, which calls it `bigState`.  `newState` is the budget component of the redux state that is being constructed and will be returned by the reducer.  Do you think calling it `newBudgetState` or something similar would help clarify it?

web/src/reducers/budget.js
```diff
@@ -497,6 +497,89 @@ const buildBudget = bigState => {
...
+  // We also have to account for the cost attributable to key personnel.  This
+  // isn't added to any activity, but instead gets rolled up into the totals.
+  // It is always a 90% federal/10% state split.  We're just going to make sure
+  // we only look at key personnel for whom the "has costs" flag is set, just
+  // to be safe.
+  keyPersonnel
+    .filter(kp => kp.hasCosts)
+    .forEach(kp => {
+      Object.entries(kp.costs).forEach(([year, cost]) => {
+        // Get the state and federal share.  Always 90/10 (for now).
```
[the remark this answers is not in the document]
mgwalker: This is definitely a question for @NAretakis or @jeromeleecms - I think for HITECH it will always be 90/10, but even if so, we should almost certainly document that this is happening somewhere.  Maybe there should be a whole thing in the documentation that describe how all the budget stuff is calculated - like all the inter-relationships and stuff.

Instead of hard-coding this, we could also have it use the FFP level from the "Program Administration" activity, since that's also where it gets its quarterly distribution.  That one is an HIT activity, so it's always 90/10 too (though users are allowed to change it to 75/25 or 50/50 and maybe they shouldn't be?).

PR #1159, opened 2018-11-20
Add tables to Storybook

hbillings:
Includes a couple of fixes to the table.css to make it a little sturdier.


web/src/styles/app/table.css
```diff
@@ -120,11 +120,14 @@
...
   width: 0;
   border-right: 0px; /* ie fix */
+  padding-left: 0;
+  padding-right: 0;
 }
 
 .table-frozen-data tr > th:first-child,
 .table-frozen-data tr > td:first-child {
-  min-width: 140px; /* same as table-frozen-col-pane; for some reason just width doesn't work */
+  width: 140px; /* same as table-frozen-col-pane; only works on fixed tables */
```
mgwalker: Interesting!  So td `width` is ignored (or somehow not honored) on normal tables?

web/src/stories/FrozenTableStory.js
```diff
@@ -0,0 +1,157 @@
...
+                </tr>
+              </tbody>
+            </table>
+          </div>
+        </div>
+      </RenderViewbox>
+      <h2>The basic structure to create this is as follows:</h2>
+      <pre>
+        <code>
+          <div className="table-frozen-wrapper">
```
mgwalker: If you wrap the contents of this `code` element in ``{` `}`` then it'll render as as string instead of rendering the HTML.
```suggestion
          {`<div className="table-frozen-wrapper">
```

web/src/stories/FrozenTableStory.js
```diff
@@ -0,0 +1,157 @@
...
+                    <td>Total Enhanced FFP</td>
+                    <td className="mono right-align nowrap"></td>
+                  </tr>
+                </tbody>
+              </table>
+              <table className="table-cms table-frozen-data">
+                [your table header and body go here]
+              </table>
+            </div>
+          </div>
```
mgwalker:
```suggestion
          </div>`}
```

mgwalker:
That was supposed to be a "request changes" review.  😬

web/src/stories/FrozenTableStory.js
```diff
@@ -0,0 +1,157 @@
...
+        </code>
+      </pre>
+      <p>
+        Essentially, this renders a second table on top of the first one. The second table is what creates the scrollable sticky pane. In order to get the height of the columns in the second table right, we render the first column of content, also (this is especially helpful when the table contains input elements).
```
mgwalker: This one is a little confusing because "second table" refers to the first one, and "first one" refers to the second table.  How about something like this?

```suggestion
        Essentially, this renders a table of just headers on top of the full table.  The first table is what creates the scrollable sticky pane.  In order to get the height of the columns in the header table right, we render the first column of real content, also (this is especially helpful when the table contains input elements).
```

(This was supposed to be in my review, but I guess I never pushed the button...)

web/src/stories/FrozenTableStory.js
hbillings: Oh! Excellent, thank you!

web/src/styles/app/table.css
hbillings: Yeah, I couldn't find a great explanation for it. My hunch is that because tables normally flex with the width of their cells' contents, the only way to set a firm width is to use the `display: fixed` setting for tables. Other tables seem to regard it as just a suggestion.

Issue #1025, opened 2018-09-20
Baseline Review of Help Text Language within the eAPD


jeromeleecms:
@jameshupp feel free to add more specificity here.

### This task is done when...
 - [ ] criteria 1
 - [ ] criteria 2
 
 ** Don't forget to fill out the LABEL and MILESTONE before submitting. If the issue is for the CURRENT SPRINT, add it to the PROJECT. If the issue is a BUG, FUTURE ENHANCEMENT, or OUTSIDE OF THE CURRENT SPRINT, do not add it to the PROJECT (then delete this line) **

PR #1168, opened 2018-11-23
Baseline help text fixes

jameshupp:
Addresses #1025

### This pull request changes …
- Brings a lot of introductory and instructional language into style compliance 
- reorders the content in certain sections for better flow

### This pull request is ready to merge …
As soon as it’s verified as done. There are three or four other content issues that I’ll open another PR for next week, but this handles the baseline text review.

- [x] Design verifies that the front end still renders properly, especially Program Overview, where the checkboxes for years is now the lead item.


web/src/i18n/locales/en/apd/apd.yaml
```diff
@@ -1,35 +1,30 @@
...
-    Briefly summarize planned HIT funded activities contained within this APD.
-  reminder: >-
+    Briefly summarize planned HIT-funded activities contained within this APD.
 
 hie:
   title: HIE Overview
   helpText: >-
-    Summarize planned HIE funded activities within this APD.
+    Summarize planned HIE-funded activities within this APD. If
+    there are no HIE activities, skip this section. Please include:
```
jeromeleecms: Is it better to ask the user to enter "No HIE activities" or something like that vs just skipping the section?

web/src/i18n/locales/en/assurancesAndCompliance.yaml
```diff
@@ -1,14 +1,13 @@
...
+  Verify that the state complies with all applicable Code of Federal Regulations
+  (CFR) and State Medicaid Manual (SMM) citations. CMS will communicate with the
+  state regarding non‐compliance issues prior to approving requests.
 citations:
   title: Federal Citations and Justification
   helpText: >-
-    Select “yes” to indicate if you are complying with each of
-    these citations. If you select "no", provide additional justifications for each item.
+    For each citation, select “Yes” if the state is complying. Select “No” if
+    the state is not complying. For every “no,” you must also provide a justification.
```
jeromeleecms: "no" -> "No" for consistency?

web/src/i18n/locales/en/proposedBudget.yaml
```diff
@@ -1,23 +1,25 @@
...
     state: In-House Costs
     contractor: Private Contractor Costs
     combined: Total Enhanced FFP
 paymentsByFFYQuarter:
   title: Incentive Payments by FFY Quarter
   helpText: >-
-    This table is for capturing incentive payments for eligible professionals (EPs) and eligible hosptials (EHs).
-  reminder: >-
-    Please include EH and EP payments and counts. This then calculates totals automatically.  
+    Use this table to capture incentive payments for eligible hospitals (EHs) and
```
jeromeleecms: The cms.gov site refers to capital **E**ligible **P**rofessionals (EPs) and **E**ligible **H**ospitals (EHs), referring them as proper nouns. Something consider for consistency: https://www.cms.gov/Regulations-and-Guidance/Legislation/EHRIncentivePrograms/index.html, but won't stop this from going in.

web/src/i18n/locales/en/scheduleSummary.yaml
```diff
@@ -1,11 +1,12 @@
 title: Activity Schedule Summary
 helpText: >-
-  The Activity Schedule Summary shows all activities and milestones for the APD in a singular view.
+  See all activities and milestones for the APD together in one table.
```
jeromeleecms: Example of 2nd vs 3rd person?

web/src/i18n/locales/en/apd/apd.yaml
jameshupp: This is a programatic question, I think. A couple different questions to consider:

- Is it useful for analysts to have that stated affirmatively? (If so, we should ask them to say it.)
- Is it common for people to skip this when they actually should have said something? I'm thinking more often than an edge case, but wouldn't necessarily be most of the time. (If so, asking them to type something can help make sure they don't skip it.)

web/src/i18n/locales/en/proposedBudget.yaml
jameshupp: I updated it. No reason not to keep it consistent with that.

web/src/i18n/locales/en/scheduleSummary.yaml
jameshupp: This in particular is the kind of content I think is helpful to have in second person. The helptext/description at the top of each section is visible before the accordion is ever opened. The content can help someone understand what they would gain by opening the accordion(s) in that section. My own read is that direct address is more effective at that: it pushes more toward using active voice, it gives people an instruction to follow, etc. So I tried to work these descriptions into that form.

(That paragraph was a little convoluted to write because `helpText` is the label we use for a couple different kinds of content, depending on whether it's inside or outside the accordion. We're fixing it for within the accordion, but the shift we'll be making does raise an issue for what to call this text outside of it. This specific example is outside the accordion. No action for right now but cc @hbillings for visibility; I think this issue might be up her alley.)

That said, if there's something incongruous about this choice with the way state staff are used to reading CMS content, it's not set in stone. I think it's important to keep these elements consistent _with each other_ to give people recognizable patterns and make it easy to read the same kind of content the same way in different places.

web/src/i18n/locales/en/apd/apd.yaml
jameshupp: Per your comment at the end of the PR, I'm going to backlog this question and not treat it as a show-stopper for now.

web/src/i18n/locales/en/scheduleSummary.yaml
jameshupp: Per your comment at the end of the PR, I'm going to backlog this question and not treat it as a show-stopper for now.

web/src/i18n/locales/en/apd/apd.yaml
jeromeleecms: @jameshupp In the APD's template, HIE content is focused in Appendix D (HIE) which most states would make a note that they do not have any HIE requests. Since we've mainstreamed that content, we could mirror that behavior here.

Also, thinking in the context of reading the APD, how would the analyst know that the state either 1) decided not to fill it out or 2)  accidentally skipped it?

web/src/i18n/locales/en/scheduleSummary.yaml
jeromeleecms: Great to understand context/use. Thank you for the detailed response!

web/src/i18n/locales/en/apd/apd.yaml
jameshupp: Love this line of thinking. Okay, let's do that. I'll add it in.

jameshupp:
Okay so it looks like this is ready to merge? Going once...

PR #1169, opened 2018-11-23
Layer 1 state dashboard

### LINKED ISSUE #1161: Build Layer 1 state dashboard
Carried forward from design in #1110 

Changes to what's currently on production include: 
- Adding introductory text to the page
- Removing the `My tasks`, `Recent activity`, and `Upcoming events` sections of the existing dashboard
- Changing the style and features of `Current statuses` to only show what APD documents exist, and allow functionality to start a new one. (Matching mockups below)
- Re-labeling the title of the former `Current statuses` section to `[State name] APDs` in the main content area and the sidebar
- Moving the "Create new" button from the sidebar and into the main content area
- Created a null state view for when there aren't any documents yet

### This task is done when...
-  [x] Design provides dev with production-ready assets (content & icons) for placement
 - [ ] Production site matches mockups for null state
 - [ ] Production site matches mockups for populated state
 
dashboard-layer1-null

dashboard-layer1

mgwalker:
Closes #1161

Some notes:
- the text colors aren't exactly right, but I wasn't sure if that was something we'd fix as part of getting everything else normalized
- I used the `h1`, `h2`, etc. classes to size the text because I wasn't sure if there was a better way 😬 (cc: @hbillings)

- implements the state dashboard from #1161


web/src/containers/StateDashboard.js
```diff
@@ -0,0 +1,101 @@
...
+
+import Icon, { faPlusCircle, faSpinner } from '../components/Icons';
+import Sidebar from './StateDashboardSidebar';
+import TopBtns from './TopBtns';
+import { createApd, selectApd } from '../actions/apd';
+import { Section } from '../components/Section';
+import { t } from '../i18n';
+
+const Loading = () => (
+  <div className="h2 p0 pb3 center">
```
mgwalker: I know this `h2` class for text size is 😬 but I wasn't sure if we had another approach ready to go yet.

web/src/containers/StateDashboard.js
```diff
@@ -0,0 +1,101 @@
...
+            <div className="px3 py2 border-bottom border-bottom-darken-1 blue">
+              <button
+                className="btn bg-blue white rounded p1 right"
+                size="small"
+                onClick={create}
+              >
+                Create new&nbsp;&nbsp;
+                <Icon icon={faPlusCircle} />
+              </button>
+              <span className="h2">{state.name} APDs</span>
```
mgwalker: Another `h2` class

web/src/containers/StateDashboard.js
```diff
@@ -0,0 +1,101 @@
...
+              {apds.map(apd => (
+                <div key={apd.id} className="p2 mb2 bg-gray-lightest">
+                  <div className="inline-block p2 mr2 bg-white blue">
+                    <img
+                      src="/static/img/icon-document.svg"
+                      alt=""
+                      className="align-middle"
+                    />
+                  </div>
+                  <a href="#!" className="h3 bold" onClick={open(apd.id)}>
```
mgwalker: And an `h3`

web/src/containers/StateDashboard.js
hbillings: Nope, this is fine for now (and might be fine indefinitely; since classes don't have semantic meaning, it's weird but maybe will end up being the best way to solve the problem anyway)!

web/src/containers/StateDashboard.js
hbillings: Any chance we could swap the order of the span and the button? The DOM order here means a screen reader will read "Create New" and then "<state> APDs," and it might make more sense the other way around. Also, can we make the span an `h2` element so it's officially the section header here? (Might have to tell it to `display: inline` or `display: inline-block` to get it to not break the button on its own line.)

web/src/containers/StateDashboard.js
hbillings: Hm, could we make it an `h3` element that wraps an `a` tag?

jenniferthibault:
SO FAST LIKE LIGHTNING!! ⚡️ ⚡️ ⚡️ !!

A couple style changes and questions:
- Since there are already APDs populated in my view, I'm not sure how to check the null state — @mgwalker for peace of mind would you mind adding a screenshot of that to the thread here for review?

- Please round the corners and square the shape of the container around the icon
  - screen shot 2018-11-26 at 10 24 02 am

- I know we're not prioritizing mobile/small screen designs at this point, but there's one small layout change that might be worth fixing — how about we make sure that if there's a really long document title (or it's on a medium/small screen), the title doesn't wrap under the image?

  - screen shot 2018-11-26 at 10 37 40 am

mgwalker:
@hbillings Changes per your request!

@jenniferthibault Changes as requested!  The rounded corners are 5px instead of 4px because that's what the `rounded` class does.  I'm happy to add a `rounded-4` class or set the radius directly if you'd rather, though.  Also, here's the screenshot:

screen shot 2018-11-26 at 9 58 19 am

hbillings:
@mgwalker woo! I betcha we can override the default `rounded` value with 4px pretty easily somehow, if @jenniferthibault wants to go with that number for all our rounded corners.

jenniferthibault:
So great! I didn't know we had a `rounded` class, so I was mostly eyeballing it in the design files. Let's keep it as is and I'll update the design files to 5px. I'm waiting for the code bot comment with the updated build link to see it in action, and then I'll review & approve. 👀 + ✅

jenniferthibault:
Oh wait — I just needed to hard refresh 😂 I'm not seeing the square size of the space behind the icon, but I understand why that is based on looking more closely at how it's built. Since this component will likely change in a future iteration, I think this works and looks totally fine for now. 👍 👍

PR #1176, opened 2018-11-26
Update dependencies - one vulnerability, one mismanaged

mgwalker:
- Through a several-layers-deep dependency (`nodemon` ➡️ `pstree.remy` ➡️ `ps-tree`), we were bringing in a [vulnerability in the `event-stream` library](https://github.com/dominictarr/event-stream/issues/116).  The things we actually depend on have been updated and `event-stream` no longer appears in our dependencies at all.

- Via the [`request`]() library, we have a dependency on `har-validator`.  The author of `har-validator` introduced a breaking change in version 5.1.2 that only affected some users - not us.  However, when the author was notified, rather than simply issue a version 5.1.3 that reverted the breaking change, they [removed version 5.1.2 altogether](https://github.com/ahmadnassri/node-har-validator/issues/112).  Unfortunately, while 5.1.2 was still active, our `package-lock.json` was updated to pin us to that version.  Thus, our builds would fail.  (They haven't failed yet because npm uses local cacheing, so our development machines and CircleCI still have 5.1.2 available; but the caches expire eventually.)
https://github.com/ahmadnassri/node-har-validator/issues/112

- updates our lockfile to bump us up a dependency version on `pstree-remy`, which removes `ps-tree` and thus `event-stream`
- updates our lockfile to pin us to `har-validator`@5.1.3

### DIFF
--- a/api/package-lock.json
+++ b/api/package-lock.json
@@ -4,6 +4,138 @@
   "lockfileVersion": 1,
   "requires": true,
   "dependencies": {
+    "@babel/code-frame": {
+      "version": "7.0.0",
+      "resolved": "https://registry.npmjs.org/@babel/code-frame/-/code-frame-7.0.0.tgz",
+      "integrity": "sha512-OfC2uemaknXr87bdLUkWog7nYuliM9Ij5HUcajsVcMCpQrcLmtxRbVFTIqmcSkSeYRBFBRxs2FiUqFJDLdiebA==",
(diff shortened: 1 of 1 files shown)


mgwalker:
A point of clarity:  `nodemon`, which is the top-level dependency that introduces the vulnerable `event-stream` is only used in our development environment.  Our staging/production configuration does not rely on `event-stream` at all.

PR #1184, opened 2018-11-27
Remove deploys we don't use

mgwalker:
Removes the UX testing and storybook deploys, since we're not using those.

(We should add a proper storybook deploy back in later.)

- no visible changes
- builds don't appear paused forever waiting on a UX testing approval


mgwalker:
That's a good idea.  I think we were deploying it to S3 now but I don't remember because it's been a few months 😬  The good news I guess is that we can do whatever we want since we're starting clean with it?

PR #1188, opened 2018-11-27
Dev enhancement: hot reload React components in-place

mgwalker:
This PR updates the dev environment to support having React components automatically reload themselves when the code is changed, without reloading the whole page.  This also maintains the application state, so you shouldn't have to re-navigate all over the place.  Unfortunately, it doesn't reload styles (because we're using the `ExtractTextPlugin` in webpack; given a little more time I might be able to take that out and get style reloading working but it's the end of the workday and 💰).

- no deployed changes, only in dev

To get your Docker environment fixed up:

```shell
docker-compose stop web
docker-compose rm web
docker-compose build web
docker-compose up -d
```


web/Dockerfile
```diff
@@ -10,4 +10,4 @@ ADD ./package-lock.json .
 
 RUN npm install
 
-CMD npm run start
+CMD npm run start -- --public=0.0.0.0:8080
```
mgwalker: Tell webpack-dev-server what address to write into the client JS it builds, so it knows how to connect back.  By default it goes to `localhost:[webpack port]`, but since we're inside a Docker container, we need to go to that container's port instead.

web/package.json
```diff
@@ -10,7 +10,7 @@
     "scaffold": "mkdir -p dist && cp src/index.html dist/ && cp ../Staticfile dist/ && cp -r src/static dist/",
     "build": "npm run scaffold && NODE_ENV=production webpack -p",
     "lint": "eslint 'src/**/*.js'",
-    "start": "webpack-dev-server --host 0.0.0.0 --port 8001 --history-api-fallback --content-base src/",
+    "start": "webpack-dev-server --config webpack.config.dev.js",
```
mgwalker: Moved a bunch of this config stuff into the config file, because having it spread around doesn't make a ton of sense.

web/src/app.dev.js
```diff
@@ -0,0 +1,44 @@
...
+    document.getElementById('app')
+  );
+};
+
+module.hot.accept('./components/Root.js', () => {
+  const HotRoot = require('./components/Root').default;
+  render(HotRoot, { history, store });
+});
+
+render(Root, { history, store });
```
mgwalker: Create a separate entrypoint for dev just because that's less likely to go awry than using conditional logic in the main entrypoint.

web/webpack.config.js
```diff
@@ -44,14 +36,9 @@ const config = {
...
+    new UglifyJSPlugin()
   ]
 };
 
-if (isProd) {
-  config.plugins.push(new UglifyJSPlugin());
-} else {
-  config.devtool = 'cheap-module-eval-source-map';
-}
-
```
mgwalker: Since this config is only for production now, we can get rid of all these conditionals.

web/src/app.dev.js
hbillings: This is a super smart idea!

web/webpack.config.js
```diff
@@ -4,8 +4,6 @@ const ExtractTextPlugin = require('extract-text-webpack-plugin');
 const UglifyJSPlugin = require('uglifyjs-webpack-plugin');
 const webpack = require('webpack');
 
-const isProd = process.env.NODE_ENV === 'production';
```
hbillings: Is this no longer necessary because we now have an explicit dev configuration?

web/webpack.config.js
hbillings: ha, welp, consider my question above answered.

Issue #1226, opened 2019-01-10
Saving the APD is not working (though system reports a successful save)


jeromeleecms:
Updating an APD and saving the changes is not working. Updated project staff, activities, and some additional fields, but after receiving a successful save notification, backing out and coming back into the APD shows none of the data has been saved.

### This task is done when...
 - [ ] Identify why a successful save notification is being reported
 - [ ] Notification information is accurately reported and/or saves are correctly being done in the database

mgwalker:
Once everything is back up and running, can we try this one again?  I'm not able to reproduce it right now, but this is a Big Deal™ so let's check it a lot.  😄

jeromeleecms:
@mgwalker Let me know when you want me to retest. Now that the app is back and running, I created a new APD, filled in a single field, and selected save. System reported "Save Successful," but when I looped back to the dashboard, back into the APD, the field was blank again.

mgwalker:
Oh!  I see.  I can reproduce that.

If you reload the page in the browser and then open your APD, does the data come back?  If so, I'm pretty sure I know where the problem is.

mgwalker:
Here's how the data flow currently goes:

1. When you load the site, we send a `GET /apds` request to the API.  This returns all of the APDs for the user's state.  This is stored in the application state at the `apds.byId` location.
2. When you select an APD on the site, we *copy* the APD data for the selected APD from `apds.byId` to `apds.data`.
3. As you edit the APD, we update `apds.data`.
4. You press save and the data from `apds.data` is uploaded to the API and saved in the database.
5. When you click back to the dashboard and select your APD again, we again copy from `apds.byId`, which was never updated.  This creates the appearance that all your saved changes were lost.  If you save the APD now, they really will be lost.

---

Proposed fix:

1. When you load the site, we send a `GET /apds` request to the API.  This returns basic metadata about all of the APD's for the user's state, but not ALL of the APD data.  This is stored in the application state at the `apds.byId` location.
2. When you select an APD on the site, we send a `GET /apds/:id` request to the API.  This returns the full data for the selected APD.  We store that at the `apds.data` location.
3. As you edit the APD, we update `apds.data`.
4. You press save and the data from `apds.data` is uploaded to the API and saved in the database.
5. When you click back to the dashboard and select your APD again, we fetch it from the API again.

This way, every time you select an APD, you are getting the most recent version from the database.

---

We need to make sure we understand everywhere that `apds.byId` is being used and what data needs to be there.

[...]

PR #1238, opened 2019-01-28
Fixes a bug where APD data appears not to be saved

mgwalker:
Addresses #1226.  Keeping that issue open until after deploy, so product owners can verify it out then.

- modifies the `GET /apds` endpoint to return only the IDs and years of a user's APDs
- adds a `GET /apds/:id` endpoint to return a single complete APD
- modifies the behavior when selecting or creating an APD in the web app
  - since the whole APD is not in app state, fetch it
  - future enhancement: when creating an APD, there's no reason to go back to the API to fetch it again
- modified the APD reducer so that it doesn't do any deserialization - pushed that logic into the action creator, instead of having it split between the two in no logical fashion

- when you edit an APD, save it, navigate back to the dashboard using the button at the top, then navigate back to your APD, your changes will still be there

- ~~Product has approved the experience~~ Unfortunately, because there are breaking backend changes, product won't be able to review this one until it's deployed.


api/routes/apds/get.js
```diff
@@ -29,4 +29,34 @@ module.exports = (app, ApdModel = defaultApdModel) => {
       return res.status(500).end();
     }
   });
+
+  app.get('/apds/:id(\\d+)', can('view-document'), async (req, res) => {
```
mgwalker: Makes this path match more precise with a regex.  Otherwise it collides with `/apds/submitted`, and the winner is whichever one just happens to be called last.   If this one is called, then it thinks the APD ID is `submitted`, which blows up the database.  But hey, we can avoid that by just making this more precise.

web/src/actions/apd.js
```diff
@@ -56,29 +56,34 @@ export const updateApd = updates => dispatch => {
...
-export const selectApd = (id, { global = {}, pushRoute = push } = {}) => (
-  dispatch,
-  getState
-) => {
-  dispatch({ type: SELECT_APD, apd: getState().apd.byId[id] });
-  dispatch(updateBudget());
-  dispatch(pushRoute('/apd'));
+export const selectApd = (
+  id,
+  { deserialize = fromAPI, global = {}, pushRoute = push } = {}
```
mgwalker: Incidentally, because `global` defaults to an empty object instead of defaulting to `window`, the `global.localStorage.setItem` on line 69 never does anything.  So that's why going straight to an APD on reload doesn't work right now.  😅

web/src/actions/apd.js
```diff
@@ -223,7 +235,9 @@ export const submitAPD = (save = saveApd) => (dispatch, getState) =>
 
       const {
         activities: { byKey: activities },
-        apd: { data: { id: apdID } },
+        apd: {
+          data: { id: apdID }
+        },
```
mgwalker: Prettier seems to just change things at random sometimes.  🤷‍♂️

web/src/actions/apd.test.js
```diff
@@ -244,36 +263,34 @@ describe('apd actions', () => {
         });
       });
 
-      it('selects an APD if local storage contains a last APD ID', () => {
-        const apd = {};
+      it('selects an APD if local storage contains a last APD ID', async () => {
```
mgwalker: This one's funny because it tests this one unit of a whole flow.  The unit works, the flow does not.

web/src/reducers/apd.js
```diff
@@ -64,7 +52,7 @@ const reducer = (state = initialState, action) => {
         {
           byId: {
             [action.data.id]: {
-              ...fromAPI(action.data),
```
mgwalker: It was really weird to me that the reducer was responsible for deserializing this data, but only this data - in other places, the data was already deserialized.  So...  I made it consistent.

The Bobs

web/src/actions/apd.js
hbillings: Should there be an issue for fixing this?

web/src/actions/apd.test.js
hbillings: 😬

PR #1240, opened 2019-01-31
Update some budget tables to use Medicaid totals, not grand totals

### LINKED ISSUE #1217: Totals in Summary Budget Table should total for the row (they are using pre-cost allocation totals)
In the Summary Budget Table, the totals should reflect _Federal_ + _State_ amounts. Instead, what is happening is that the totals are showing the pre-cost allocation total amounts for each project activity type. 

There might be a clever way to show the pre-cost allocation total, the cost allocated amount, and then the Federal and State shares, but if there is not, then just totaling _Federal_ and _State_ will do.

image

### LINKED ISSUE #1218: Program Budget Table needs to total Federal + State, instead of total pre-cost allocation dollars.
Related to #1217 the program budget table total amount should be a sum of Federal + State. And the same for MMIS (Federal + State).
image

mgwalker:
The summary budget table, in the Proposed Budget section, had a column for "total."  This column was the total cost, but it should be using the Medicaid share (ie, `total - other funding`).  The same bug was in the program budget tables in the Executive Summary section.  The result is that the columns did not sum up within a row (ie, if there was other funding, `state + federal` would not sum to `total`).

This PR adds Medicaid costs to the budget state and updates those two tables to use the Medicaid share instead of total costs.  Also tweaks the column headers in those tables to read `Medicaid total` instead of just `Total`, and tweaks the description of the summary budget table.

- Fixes #1217
- Fixes #1218


web/package.json
```diff
@@ -122,7 +122,7 @@
     "postcss-import": "^11.1.0",
     "postcss-loader": "^2.1.6",
     "postcss-url": "^7.3.2",
-    "prettier": "^1.15.1",
+    "prettier": "^1.16.3",
```
mgwalker: It seems all the prettier editor plugins were updated to point to a particular parser or something or other, and now they all require prettier ^1.16?  Anyway, this made prettier work in my editor again, hopefully anyone else on our team using a prettier plugin won't be hurt by this change...  If so, let's chat and find a solution!

web/src/containers/BudgetSummary.js
```diff
@@ -95,14 +95,9 @@ class DataRow extends Component {
...
-          detailsOpen && (
-            <DataRowDetails
-              category={category}
-              entries={entries}
-              years={years}
-            />
-          )}
+        {hasData && detailsOpen && (
+          <DataRowDetails category={category} entries={entries} years={years} />
+        )}
```
mgwalker: Speaking of prettier...  🤷‍♂️

web/src/containers/BudgetSummary.js
```diff
@@ -177,22 +172,13 @@ const BudgetSummary = ({ activities, data, years }) => (
           <th id="summary-budget-null2" />
           {[...years, 'total'].map(y => (
             <Fragment key={y}>
-              <th
-                className="right-align"
-                id={`summary-budget-fy-${y}-total`}
-              >
-                Total
+              <th className="right-align" id={`summary-budget-fy-${y}-total`}>
+                Medicaid total
```
mgwalker: @jameshupp I lowercased "total" - it seemed like we were doing that in other places, where we only uppercase the first word.  Is that right?

web/src/i18n/locales/en/executiveSummary.yaml
```diff
@@ -21,4 +21,4 @@ budgetTable:
   mmisTotal: MMIS (Total)
   fedShare: Federal
   stateShare: State
-  grandTotal: Total
+  grandTotal: Medicaid total
```
mgwalker: @jameshupp Same here with the capitalization.

mgwalker:
"Medicaid total" isn't new, but I think this is the first time we're putting it in one of the tables - it was already shown in the cost allocation section.  It makes more sense this way - previously, the columns in the table didn't appear to add up because the "other funding" amount wasn't in the table, but it was accounted for in the total.  This way, the "other funding" is both not shown and not included in the displayed sum.

PR #1264, opened 2019-02-06
Make authentication replay-resistant [skip deploy]

### LINKED ISSUE #1249: Add replay resistance to authentication
NIST 800-53 IA2(8) requires a replay-resistant authentication mechanism.  Our current mechanism is trivially susceptible to replay.  We can make authentication two steps and introduce a single-use server-created and -verified nonce.

(Should we have a milestone related to login?)

### This task is done when...
 - [ ] authentication has two steps - one to get the nonce and a second to get the user's authentication token
 - [ ] frontend is updated to use two-step authentication

mgwalker:
This PR modifies our authentication to turn it into a two-step process.  This is to mitigate replay attacks.  A replay attack is one in which a malicious actor records a network request from a valid user and replays that request later to gain access to a system.  Our existing authentication is susceptible to this kind of attack.  NIST 800-53 IA2(8) requires replay attack mitigation.

In the new approach, an authentication request must first obtain a nonce from the API.  This is done by sending the username to the `/auth/login/nonce` endpoint.  The resulting nonce encodes the username and a current timestamp, is cryptographically signed by the API, and expires after 3 seconds.  Next, a request for an authenticated token is sent to `/auth/login`.  This request sends the nonce along with the user's password.

Several conditions must pass in the second request in order for a user to be authenticated:
- the nonce must be structurally valid
- the nonce signature must validate, indicating that it was not tampered with since the API created it
- the nonce must not be expired
- the username in the nonce must match a real user
- the password in the request must match the password for the username in the nonce

Because the nonce includes a timestamp and has an expiration, it cannot be used directly in replay attacks.  Because the nonce is cryptographically signed and validated, a malicious actor cannot modify the nonce without invalidating it.  Replay attack: mitigated!

This is all invisible to the user.  They will still enter their username and password into the form.  The web app will handle making the two separate API calls automagically.

Closes #1249

- no visible changes


api/auth/index.js
```diff
@@ -20,9 +20,12 @@ const defaultStrategies = [new LocalStrategy(authenticate)];
...
   app,
-  passport = Passport,
-  strategies = defaultStrategies,
-  session = sessionFunction
+  {
+    auth = authenticate,
+    passport = Passport,
+    session = sessionFunction,
+    strategies = defaultStrategies
+  } = {}
```
mgwalker: Changed this set of injection parameters into a single object. Makes them more coherent, and also means the order doesn't matter because they're destructured by name. 😄

api/auth/index.test.js
```diff
@@ -24,23 +28,33 @@ tap.test('authentication setup', async authTest => {
...
+    authSetup(app, { passport, strategies, session });
 
     setupTest.equal(
       app.use.callCount,
       3,
       'three middleware functions added to app'
     );
     setupTest.ok(
-      app.use.calledWith(sinon.match.func),
+      app.use.calledWith(session),
```
mgwalker: Not sure this wasn't injected in the tests in the first place.

api/auth/index.test.js
```diff
@@ -24,23 +28,33 @@ tap.test('authentication setup', async authTest => {
...
+
   const strategies = ['strategy1', 'strategy2'];
 
+  const res = {
+    send: sandbox.stub(),
+    status: sandbox.stub(),
+    end: sandbox.spy()
+  };
+  res.send.returns(res);
+  res.status.returns(res);
```
mgwalker: Moved `res` higher up in the scope, since it was being declared in multiple places.  I needed to add another place, so it made sense to just bump it up here.

api/utils.endpoint.js
```diff
@@ -21,9 +21,15 @@ const login = async (
 ) => {
   const cookies = request.jar();
 
+  const {
+    body: { nonce }
+  } = await execRequest('post', getFullPath('/auth/login/nonce'), {
+    json: { username }
+  });
```
mgwalker: This is in the `login` helper method.  Added this step to get a nonce.

web/package.json
```diff
@@ -86,8 +86,7 @@
     "redux": "^3.7.2",
     "redux-thunk": "^2.3.0",
     "tippy.js": "^2.6.0",
-    "updeep": "^1.0.0",
-    "yup": "^0.25.1"
```
mgwalker: This library has a funky dependency on an old version of the Babel runtime and it was causing build problems.  We only use it in our validating input components, which we aren't using except in one place and we shouldn't be.  So I removed those too...

web/src/actions/auth.js
```diff
@@ -2,8 +2,6 @@ import axios from '../util/api';
 
 import { fetchApd } from './apd';
 
-const API_URL = process.env.API_URL || 'http://localhost:8000';
```
mgwalker: We can rely on the base URL when we first setup the API library, so no need to put it here.

api/auth/authenticate.js
```diff
@@ -1,34 +1,62 @@
...
   userModel = defaultUserModel,
   bcrypt = defaultBcrypt
-) => async (username, password, done) => {
+) => async (nonce, password, done) => {
   try {
-    logger.verbose(`got authentication request for [${username}]`);
+    logger.verbose(`got authentication request. decoding jwt...`);
+    let verified = false;
+    try {
+      verified = jwt.verify(nonce, NONCE_SECRET, { algoithms: ['HS256'] });
```
hbillings: typo: `algoithms`

api/auth/index.test.js
```diff
@@ -169,17 +183,54 @@ tap.test('authentication setup', async authTest => {
...
-      send: sinon.stub(),
-      status: sinon.stub(),
-      end: sinon.spy()
-    };
-    res.send.returns(res);
-    res.status.returns(res);
+      post({}, res);
+
+      test.ok(res.status.calledOnce, 'an HTTP status is set once');
+      test.ok(res.status.calledWith(400), 'sets a 200 HTTP status');
```
hbillings: wait, the status  returns 200 for a 400?

api/auth/authenticate.js
mgwalker: Oooh, good catch.  Need to add a test on this one - setting the `algorithms` property when verifying the tokens is strictly necessary, for security reasons.  Otherwise an attacker could set the token header to indicate a `null` algorithm and it'd happily verify it, regardless of the signature.  😬

api/auth/index.test.js
mgwalker: Doh.  Will fix the typo.

PR #1265, opened 2019-02-06
Fix session/cookie [skip deploy]

### LINKED ISSUE #1251: Make authentication token expiration configurable
When you log in, your authentication token currently has a hard-coded lifetime and sliding expiration.  This needs to be configurable.  The sliding expiration needs to be a separate configuration because it might not slide at all, depending on CMS policy.

(Login milestone?)

### This task is done when...
 - [ ] token expiration is configurable
 - [ ] token lifetime extension is configurable

### LINKED ISSUE #1252: Fix cookie/token permanence bug
The library we use to set and extend authentication tokens currently uses the expiration information stored in the original cookie to calculate the token's expiration, regardless of any configuration changes.  It also does not unset the cookie when a user logs out, it just empties the decoded contents, so the original cookie expiration info is persisted indefinitely.

The upshot is that once someone logs in, they will permanently inherit the token expiration policy in place at that moment and there is no way to enforce a new policy on them.  That's...  obviously wrong.

(login milestone?)

### This task is done when...
 - [ ] token expiration is calculated based on token creation time and ***current*** policy
 - [ ] auth cookie is cleared on logout

mgwalker:
# 🚨 Hold for #1264

Closes #1252 

The library we were using ([client-sessions](https://npm.im/client-sessions)) set a session cookie with an encoded lifetime value (not an expiration date - a number of seconds) and provides no way to unset the cookie.  Further, when the session cookie already exists (even if it's invalid) and the user logs in, the cookie is updated but the lifetime value is not modified.  The upshot is that once a user has a session cookie, it is impossible to change the lifetime of the cookie - that user is immune to system configuration changes.

This PR implements our own cookie handling.  The new behavior is to include an expiration date rather than a lifetime, and to update it from the system configuration.  Incoming cookies are only used to reconstruct the session.  Outgoing cookies are built from scratch using server-side data.

Closes #1251

Session expiration times are controlled by the new `SESSION_LIFETIME_MINUTES` environment variable, which defaults to 2 days.  Sessions expire after this long ***without activity***.  There is no absolute expiration - as long as users keep using the system with less than a 2-day inactivity gap, they will stay logged in.

- no visible changes


api/auth/session.js
```diff
@@ -1,29 +1,121 @@
...
+            process.env.SESSION_SECRET,
+            {
+              algorithm: 'HS256',
+              expiresIn: `${process.env.SESSION_LIFETIME_MINUTES}m`
+            }
+          ),
+          { httpOnly: true }
+        );
+      } else {
+        // Else, write a cookie with an immediate expiration
```
hbillings: what does writing a cookie with an immediate expiration mitigate?

mgwalker: It just gets rid of it.  There's no mechanism in the HTTP spec for telling a browser to delete a cookie.  The closest thing is having it expire in the past or right now.

Issue #1244, opened 2019-02-05
Set up webpack to compile Sass


hbillings:
If we're going to use the CMS design system, we're going to need to get Webpack to compile Sass.

### This task is done when...
 - [ ] we have a Sass folder and webpack checks it for things it needs to compile.

mgwalker:
Closed by #1266 I think?

PR #1266, opened 2019-02-06
Add Sass compilation to webpack configuration [skip deploy]

mgwalker:
Addresses #1244 

Our CSS relies on `ace-css` and includes it via a CSS `@import`.  Webpack can traverse that import and find the right thing.  The Sass compiler, however, cannot.  To manage this, we now have two style outputs.  The current CSS has been renamed `legacy.css` and outputs a `legacy.css` file.  A new `index.scss` has been added that generates `app.css`.  The `index.html` file was updated to load them both.

I also updated the webpack configs to load the style files directly, instead of relying on an `import` statement in the Javascript somewhere.

In the dev configuration, webpack now writes the styles directly into the Javascript via `style-loader`, which will write `<script>` tags as soon as the script loads.  Doing it this way, hot reloading works with styles.  😀

- no visible changes


web/webpack.config.js
```diff
@@ -36,7 +66,6 @@ const config = {
     new webpack.EnvironmentPlugin({
       API_URL: null
     }),
-    new ExtractTextPlugin('app.css'),
```
hbillings: I'm glad we got to rip that out!

PR #1293, opened 2019-02-13
Adds API endpoint to modify users

### LINKED ISSUE #1292: Create backend endpoint to modify users
Need a PUT endpoint at `/users/:id` to update an existing user's name, email address, phone number, position, associated state, and authorization role in order to be able to do #1254.

### This task is done when...
 - [ ] users can be updated with the API

mgwalker:
Adds a `PUT /users/:id` endpoint that allows an admin to modify other users.  For security reasons, this endpoint cannot be used to update the current user and cannot be used to modify passwords.  (We should have a password reset for that.)

Closes #1292

- no visible changes


api/routes/users/put.test.js
```diff
@@ -0,0 +1,411 @@
...
+        test.ok(res.send.notCalled, 'body not sent');
+        test.ok(res.end.calledOnce, 'response is terminated');
+        test.ok(
+          res.end.calledAfter(res.status),
+          'response is terminated after status is set'
+        );
+      });
+    });
+  });
+});
```
mgwalker: I fear I've gone down a bad path with these unit tests, from the beginning. They are so complex. I think it implies the code under test is also too complex, or has too many dependencies.  Mocking all the database stuff is difficult, but necessary since we don't want to hit an actual database.  I don't know...  But writing unit tests takes longer than anything else, and that feels like a smell.  😕

api/docker-compose.endpoint-tests.yml
```diff
@@ -1,7 +1,7 @@
 version: "3"
 services:
   db:
-    image: postgres:latest
+    image: postgres:10
```
hbillings: :jinx:

api/routes/users/put.endpoint.js
```diff
@@ -0,0 +1,115 @@
...
+
+      expect(response.statusCode).toEqual(403);
+      expect(body).toMatchSnapshot();
+    });
+
+    describe('when updating a valid user ID', async () => {
+      it('...with an invalid state ID', async () => {
+        const { response, body } = await put(2001, { state: 'xx' });
+
+        expect(response.statusCode).toEqual(400);
```
hbillings: Why would we want a valid user ID to return a 400?

api/routes/users/put.js
```diff
@@ -0,0 +1,118 @@
...
+        );
+        return res.status(404).end();
+      }
+      logger.verbose(
+        req,
+        `request to modify user [${targetUser.get('email')}]`
+      );
+
+      // These fields aren't related to other tables, so we can update them
+      // all willy-nilly.
```
hbillings: my favorite way to update things, personally

api/routes/users/put.endpoint.js
mgwalker: The user is valid, but the state is not. AFAIK, 400 is a generic "the client sent something wrong" response, but if that's not right, happy to change it to something more appropriate

Issue #1038, opened 2018-09-26
Fix the max width of the left nav


lauraponce:
When the page width increases, the left nav width also continues to increase with it. Can we limit the left nav at a certain point (it looks like 300px would work) so that it doesn't continue to grow and take up real estate of the page? 

![screen shot 2018-09-26 at 10 39 33 am](https://user-images.githubusercontent.com/21222704/46098327-fecec480-c178-11e8-9747-8583e2ecad9a.png)

hbillings:
Okay, it looks like we're using a homegrown grid pattern, which means:
a) doing responsive design of any kind is going to suck
b) controlling max-widths becomes a fragile thing because suddenly you're doing it on a case-by-case basis rather than using general rules.

We _really_ should be using a standard grid for this, so I'm going to look into what Basscss provides and see if that's helpful. Implementing a one-off grid system is just going to cause us (and future devs) pain down the road.

Issue #1267, opened 2019-02-07
Restyle side nav in eAPD prototype to use design system vertical nav component


jenniferthibault:
_Part of #1250_ 

In order to start understanding how to use the CMS design system, we should start with a component that occurs infrequently and is modular enough to edit alone. The side nav in the app prototype fits that bill!

### This task is done when...
 - [x] The side nav has been replaced with the vertical nav component in the sketch files
 - [x] We know what we expect it to look like in place (through mockups)
 - [ ] The side nav has been replaced and customized in the code
 - [ ] The side nav's sticky functionality has been restored on long pages

 **Current side nav:**
![screen shot 2019-02-06 at 10 55 31 pm](https://user-images.githubusercontent.com/11636908/52389627-5acc2280-2a62-11e9-99a3-4af01c4b6fed.png)

**[Vertical navigation component](https://design.cms.gov/components/vertical-nav/) from CMS Design System**
![screen shot 2019-02-06 at 10 56 58 pm](https://user-images.githubusercontent.com/11636908/52389664-8d761b00-2a62-11e9-9719-4bf1ae8b57ca.png)

jenniferthibault:
Here's what the vertical nav could look like in our context:
![screen shot 2019-02-12 at 11 04 21 am](https://user-images.githubusercontent.com/11636908/52649281-0ab3ec80-2eb6-11e9-87bc-b6aceaac2afc.png)

**Additional notes:**
- I dropped the "HITECH" text in this mockup to reflect [Jerome's comment](https://github.com/18F/cms-hitech-apd/issues/1126#issuecomment-460375167)
- I started looking at the button styles in addition to the vertical nav component, since they're layout was grouped, but we could totally reserve that for #1280 

A question for @lauraponce: 
- Do you remember if there was a decision to truncate the menu text with ellipses `...` on purpose? In the mockup above, I wrapped text to a new line instead of hiding it, but glad to revisit if you remember that being tested & intentional

A question for @hbillings:
- This mockup aligns the navigation to the grid, but the right-hand side of the nav container expands into the grid's gutter (probably thru padding?). There aren't any layout files for me to compare this with, so I'm not sure what the most logical treatment is, but from a layout perspective, it will always look best when we line up text on the grid columns within containers. Is this a feasible way to go? Or is there a better approach?

[...]

PR #1295, opened 2019-02-14
Add CMS design system grid

### LINKED ISSUE #1246: Add CMS design system grid to our project
The CMS design system separated their grid into an NPM package that's installable separately from the main system. I think it makes sense to install the grid first, since we already know from #1054 that our current grid isn't fulfilling all of our needs. It will also buy us a little time to figure out how the more visible elements of the design system affect what we've currently got going on.

### This task is done when...
 - [x] The grid package is installed
 - [ ] The pages have been updated to use the grid styles

hbillings:
Closes #1246, #1038.

Out with the old grid, in with the new grid!

- the grid system
- left nav should now be a fixed width thanks to the new grid


docker-compose.yml
```diff
@@ -1,7 +1,7 @@
 version: "3"
 services:
   db:
-    image: postgres:latest
+    image: postgres:10
```
mgwalker: Thank you!

hbillings:
@mgwalker I realized pairing this morning with Jen that the site's original grid was 10-column based, while the design system grid is 12-column based. So you're right -- proportionally, the main content was missing two columns! Good eye!

The loss of the sticky sidebar is definitely unintentional, and I'm not sure what I did to cause that. I'll poke around a bit more in this PR. If it seems like it's a big lift, I'll roll that work into the nav restyling issue I'm picking up next.

hbillings:
Okay, the sticky sidebar was previously created by using `position: fixed` on the sidebar container. This worked because we were specifying the sidebar's `max-width` columns in the CSS. With the new grid system, the column width is specified as a class on the parent container, and setting `position: fixed` causes the children to overflow the container. I'm not entirely sure what the right fix for this is, but since we're going to need to restructure the nav to implement the new styles, I think it makes sense to bundle this with that. Are you good with that, @mgwalker & @jenniferthibault?

screen shot 2019-02-14 at 3 47 03 pm

jenniferthibault:
Taking that on as part of #1267 works for me!

mgwalker:
To make sure I understand, the plan is to merge it as-is and fix the sticky piece of the nav in #1267?  If so, that makes a ton of sense to me.  I'd ping in @jeromeleecms and @NAretakis for their awareness and feedback, though, if any.

hbillings:
Thanks, both! I added a checkbox to the description of #1267 to make sure we don't lose that acceptance criterion. I'll ping the POs just to make sure.

Issue #1178, opened 2018-11-27
Clarify and execute the handling of section descriptions in YAML


jameshupp:
This is a dependency for #1088.

Section descriptions—the introductory text for a section that appears _before_ the subsection accordions—are currently `helpText`. But per #1088, we are revising the handling of `helpText` and `reminder` to be `instruction` and `helpText`, respectively. Along with that we are going to more strictly manage what kind of text goes in each place.

Given this, it's probably inappropriate to handle section descriptions as `helpText`. So we need to get clarity on that.

### This task is done when...
 - [x] Front end, visual design, and content have a shared understanding of what this text is and how it aligns with the rest of the text in the app
 - [x] The appropriate label has been applied to these content elements in YAML.

jameshupp:
Even though the PR that executes this is still open, I'm moving this to done because the instruction blocks themselves look right. Just have to make sure the rendering works in #1088.

PR #1299, opened 2019-02-15
Implement instruction block style

jameshupp:
This replaces the disparate way of structuring our microcopy with a single structure. It applies to the majority of copy, but there are a few exceptions, which are [described in 1088](https://github.com/18F/cms-hitech-apd/issues/1088#issuecomment-463317107). As it stands today, it addresses #1178, but **it is not ready to merge until #1088 is addressed**.

- [x] The orphaned yaml element issues in 1178 are resolved, probably pairing between me and @mgwalker
- [x] 1088 is addressed so that the changes of 1178 render properly in the app
- [x] Associated OpenAPI documentation has been updated (if relevant)


mgwalker:
- in `activities/overview/detail`, it looks like the `hit` and `mmis` sub-properties are never used, only `standard` and `hie`.  The content under `standard` is pretty minimal.  Should we be using the `hit` and `mmis` properties (which have identical content), or should the `standard` property have the content from the other two?  (Not meant to suggest those are the only two options, but they're the two that seem likely to me at a glance.)
- in `previousActivities`, there's an `outline.instruction.label` property.  I modified the code to use it, but it left me wondering if maybe we should take that out and remove it from the code too.

The code changes are done, just need to talk to James real quick and then I'll push them up.

mgwalker:
@hbillings Would you mind reviewing the code changes when you get a chance?

jameshupp:
There are a couple issues with how this is rendering.

- Per #1088, the `short` part of an `instruction` block should be rendered as **strong text**. Right now it's rendering the same as `detail`, which is to say, as normal body text.

- In costAllocate.yaml there is a numbered list. The list is rendering without any line breaks between items. For lists of short text this makes sense; for lists of long text like this one not so much. I'm fine with leaving this as is for right now, since it's unique and the CMS design system may alter how it renders. No need to come up with a whole new list class. But if there's a way to enter numbers at the beginning of paragraphs in YAML without turning them into lists I would be very into it. (This may not be the only one but it is a good example of where this is weird. I'm not sure why this is a numbered list, but it seems technical in nature so I'm leaving it.)

That second bullet has some very front end-y things going on, so tagging @hbillings for visibility if nothing else.

mgwalker:
Oh, oops!  The short wasn't bolded because I was setting its class to `strong` instead of `bold`.  😬 Fixed in e8d2d93

On the list in cost allocation, if you add an extra line break between each paragraph, you'll get some separation between them.  It's a result of the way the markdown library we use parses - instead of putting each block into a `<p>` tag, it replaces line breaks with `<br>` tags.  That feels like a bug, and maybe we should look into switching markdown libraries...

jameshupp:
The rendering of the font weight looks good to me! As far as I'm concerned this is good to push, just need code review.

It's raised a couple new issues for me, though, but they shouldn't stand in the way so I'm filing them separately.

web/src/components/Instruction.js
```diff
@@ -11,29 +11,31 @@ HelpText.propTypes = {
...
+  const detail = t([source, 'detail'], { defaultValue: false, ...args });
+  const helpText = t([source, 'helpText'], { defaultValue: false, ...args });
 
   return (
     <div>
       {heading && <h3>{heading}</h3>}
-      {reverse && detail && <p>{detail}</p>}
-      {short && <p className="strong">{short}</p>}
-      {!reverse && detail && <p>{detail}</p>}
+      {reverse && detail && <Md content={detail} wrapper="p" />}
```
hbillings: I'm curious about what this change gets us.

mgwalker: Instead of making the `detail` just a `<p>` with no internal formatting, it renders it as Markdown wrapped in a `<p>`.

PR #1302, opened 2019-02-20
Add frontend for editing user accounts

### LINKED ISSUE #1254: Create UX to manage users
Need a UX to assign users to roles, which is how they get permission to do things in the system.

(login milestone?)

### This task is done when...
 - [ ] the form exists and submits new users to the API
 - [ ] users without permission are redirected to the dashboard

mgwalker:
Closes #1254

- adds a web UX at `/admin/edit-account` for modifying account details

   screen shot 2019-02-20 at 10 14 30 am

   - the form is hidden if no user account is selected
   - the "Authorization role" field is disabled if the user selects their own account

- updates the `PUT /users/:id` endpoint so users can update themselves with this endpoint, but just not their authorization role (it felt like hostile design to limit it that way)
- updates the `PUT /users/:id` endpoint to provide a way to unset the account's associated state


web/src/actions/auth.js
```diff
@@ -56,7 +69,7 @@ export const checkAuth = () => dispatch => {
     .get('/me')
     .then(req => {
       dispatch(completeAuthCheck(req.data));
-      dispatch(fetchApd());
+      dispatch(loadData(req.data.activities));
```
mgwalker: If the user is logged in, instead of just loading their APDs, load _all_ relevant data.

web/src/actions/auth.js
```diff
@@ -25,6 +26,18 @@ export const failLogin = error => ({ type: LOGIN_FAILURE, error });
...
+  if (activities.includes('view-document')) {
+    dispatch(fetchApd());
+  }
+  if (activities.includes('view-users')) {
+    dispatch(getUsers());
+  }
+  if (activities.includes('view-roles')) {
+    dispatch(getRoles());
+  }
+};
```
mgwalker: Check the user's permissions to see which data to load, and dispatch those loads.

web/src/actions/auth.js
```diff
@@ -38,7 +51,7 @@ export const login = (username, password) => dispatch => {
     )
     .then(req => {
       dispatch(completeLogin(req.data));
-      dispatch(fetchApd());
+      dispatch(loadData(req.data.activities));
```
mgwalker: Also do this data loading dance when the user finishes logging in, in case they weren't already.

web/src/containers/admin/EditAccount.js
```diff
@@ -0,0 +1,201 @@
...
+      );
+    }
+    return null;
+  };
+
+  handlePickAccount = e => {
+    const { users } = this.props;
+    const { value } = e.target;
+
+    const user = users.filter(u => u.id === +value).reduce((_, u) => u, null);
```
mgwalker: IE doesn't support [Array.find](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/find), so this seems like a reasonable workaround - if there's not a matching user, returns `null`; else returns the single matching user.

web/src/containers/admin/EditAccount.test.js
```diff
@@ -0,0 +1,188 @@
+import { shallow } from 'enzyme';
```
mgwalker: There aren't any snapshot tests in here.  I think I'm pleased with that, but I'm curious what you think @hbillings?

web/src/reducers/auth.js
```diff
@@ -11,19 +11,30 @@ const initialState = {
   initialCheck: false,
   fetching: false,
   authenticated: false,
-  error: ''
+  error: '',
+  user: null
```
mgwalker: Interestingly, we've never stored the logged-in user's information anywhere.  Surprising!

web/src/containers/admin/EditAccount.test.js
hbillings: I've never _really_ understood what we get from snapshot tests, so I think I'm fine leaving them out.

Issue #1164, opened 2018-11-21
Clarify the language related to the Incentive Payments table 


lauraponce:
### Background

In the 8th round of usability testing, there was some confusion about the EH and EP counts. One user said:

> Is the EH or EP count totals or unique values? I found it difficult to determine unique counts for EP in a given fiscal year. Our EPs come in on different years. I’m assuming this here is not a unique count, this is the EH payments made.

According to Nick, this is a HITECH program nuance. Payments is dollar amounts and the count is optional but it’s confusing to people. Are they unique physicians that participated in the program or are they the count of payments made to the physicians? This needs to be better defined.

### This task is done when...
 - [ ] There are language updates to the Incentive Payments section so that it's easier to understand.

Currently, this section looks like:

<img width="708" alt="screen shot 2018-11-21 at 10 22 06 am" src="https://user-images.githubusercontent.com/21222704/48860994-52bfe900-ed77-11e8-8e0a-bb67fa986dc4.png">

jameshupp:
Recording this question for when I'm able to pick up this issue.

My read of the issue as written is:
- It's actually not clear _to HITECH_ whether `Count` refers to the number of physicians or the number of payments made.
- On the other hand, it is clear that `Payments` refers to the total amount of money, not the amount per payment or per physician.

Is that accurate? In which case, we're really just clarifying what `Count` means?

lauraponce:
@jameshupp your read of the issue is the same as mine. Most of the uncertainty for this section is related to `Count` and what that means.

When I wrote this issue, I tried to capture what @NAretakis was explaining when we were going over the synthesis, but hopefully he can add more about how we are defining `Count` in this context.

NAretakis:
Here's the scoop:

* `Count` refers to the number of physicians or eligible hospitals (EHs) that were paid. This might look something like, an EP `Count` of 10 (10 providers paid in Q1), and a corresponding total payment of $85,000 (paid in Q1). This equates to $8,500 paid per provider.

* `Payments` refers to the total amount of money, not the amount per payment or per physician.


Does this make sense?

jameshupp:
So for the number of physicians/hospitals, we need to say that `Count` refers to number of providers in a way that's clear for both states and HITECH, while also not stomping all over the layout of the table.

Okay! I'll put something together and see if we can close this out!

PR #1305, opened 2019-02-20
Clarify what Payment and Count are

jameshupp:
Addresses #1164.

I went back and forth on this for a bit, trying to figure out how to say `EH Payments` differently in the table, but I think it works way better to change the instruction copy. Now it explains those words more clearly; the words are unchanged. (This has the bonus benefit of not requiring any code, which changing table headers would have required. I don't really like that setup fwiw.) I also got a bit more specific about "calculate"; the previous wording made it seem like `payments` may mean amount per payment.

No need for an accessibility audit, as this only changes text that was confusing in the first place.

### DIFF
--- a/web/src/i18n/locales/en/proposedBudget.yaml
+++ b/web/src/i18n/locales/en/proposedBudget.yaml
@@ -21,5 +21,6 @@ paymentsByFFYQuarter:
   title: Incentive Payments by FFY Quarter
   helpText: >-
     Use this table to capture incentive payments for Eligible Hospitals (EHs) and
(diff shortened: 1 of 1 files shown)


mgwalker:
Is it `number of payments` or `number of entities paid` (poor wording, but...)?  E.g., if a hospital is paid 10 times, is `count` 1 or 10?

lauraponce:
For `Count` we're defining as **number of payments (Count)**, but could it also apply to the amount of physicians or hospitals that were paid and not just the payments?

lauraponce:
Realized I might have asked the same question as @mgwalker

PR #1309, opened 2019-02-21
Hooks up export button to browser print dialog

### LINKED ISSUE #1281: Remove print preview icon and hook up export button
_Builds off of #1280_

The Save / Export buttons will be updated in #1267, so this issue is for hooking up the **functionality** of the Export button (currently, the button label is Save as PDF) and removing the printer button.

Greg's note about the dev work on #1280:

> From a code perspective, we can programmatically trigger the browser's print dialog, so if we want an `Export` button that opens the print dialog, we can do that.
> 
> One thing to note is that we need to run some code before the app is ready to print. I think most of the changes are handled with a print media stylesheet (for which we don't have to do anything 🎉), but some will need to be done with Javascript (e.g., changing `$1.2M` to `$1,203,582`). We can reliably run our code if the user triggers a print using our button, but it might be a little trickier if they use the browser's native `File -> Print` method.
> 
> There are browser events for [before the print dialog is displayed](https://developer.mozilla.org/en-US/docs/Web/API/WindowEventHandlers/onbeforeprint) and [after the job has been sent to the printer](https://developer.mozilla.org/en-US/docs/Web/API/WindowEventHandlers/onafterprint), so we should be able to hook into those to run our code before it gets to the print dialog. My understanding is that support isn't completely consistent across browsers, so we might need to check it in a few to be sure it works as expected. We could prototype that separately, though.

### This task is done when...
 - [ ] Print preview button on the bottom right of the page has been removed
 - [ ] The export button works and triggers the print dialog

mgwalker:
Closes #1281

- gets rid of the floating "print preview" button
- hooks up the "Save as PDF" button to trigger app state changes and then open the browser's print dialog
- uses `@media print` instead of a big ol' class on top of the regular media

For testing the print behavior, we can use Chrome's rendering dev tool to emulate the print media.

- [ ] ~~The experience passes a basic manual accessibility audit (keyboard nav, screenreader, text scaling) OR an exemption is documented~~ these changes don't affect a11y
- [ ] ~~The change has been documented~~
  - [ ] ~~Associated OpenAPI documentation has been updated~~


web/src/actions/print.js
```diff
@@ -0,0 +1,6 @@
+export const PRINT_APD = Symbol('print : apd');
+
+export const printApd = ({ global = window } = {}) => dispatch => {
+  dispatch({ type: PRINT_APD });
```
mgwalker: Later, we can have a reducer that responds to this action to update some piece of state, and we can use that piece of state to trigger UI changes elsewhere?  I was thinking, for example, that our `Input` components could check that state and switch from inputs to just text blocks.  That'd be pretty neat.

web/src/styles/scss/print.scss
```diff
@@ -0,0 +1,54 @@
+@media print {
```
mgwalker: Copied these from `print.css`, which is deleted.

hbillings: Aha! I wondered what was going on there. Thanks for moving this over!

PR #1310, opened 2019-02-22
Add design system sidenav styles + component

hbillings:
Swaps the existing sidenav for the standard CMS design system one.

- Adds the design system core to our codebase
- The React component used to render the sidenav
- Obsoletes some of our code


web/src/containers/StateDashboardSidebar.js
```diff
@@ -4,6 +4,7 @@ import { connect } from 'react-redux';
 
 import { t } from '../i18n';
 import { selectApd } from '../actions/apd';
+import VerticalNav from '@cmsgov/design-system-core/dist/components/VerticalNav/VerticalNav';
```
hbillings: Did it this way so we're only importing one component 😬

web/src/containers/StateDashboardSidebar.js
```diff
@@ -29,20 +30,23 @@ class Sidebar extends Component {
...
+              items={[
+                {
+                  label: `${place.name} APDs`,
+                  id: "state-apds",
+                  items: apds.map(apd => (
+                    {
+                      id: `${apd.id}`,
+                      label: `${apd.years.join(', ')} ${place.id.toUpperCase()} APD`,
+                      url: 'javascript:void(0);',
+                      onClick: this.pickApd(apd.id)
```
hbillings: This isn't documented in their stuff but seems to work...

jenniferthibault:
I know this is still in progress, but here's the only thing that might be worth flagging early:

On the dashboard, it looks like the menu is butting up to the edge of the page, like so:
screen shot 2019-02-22 at 4 48 58 pm

I'd expect to see some space between the edge of the nav and the edge of the page, as seen on the design system site:
screen shot 2019-02-22 at 4 48 45 pm

I suspect that padding on both sides should also make the two have identical widths. Helpful?

hbillings:
I think this is almost entirely done except for a few easy styling fixes and figuring out how to show which item is selected. I'm having a devil of a time figuring out how to pass the `selectedId` prop up from `VerticalNavItem` to `VerticalNav`. It _seems_ like this should be something that happens automatically, but it isn't, at least as far as I can tell. The [docs](https://design.cms.gov/components/vertical-nav/#usage) merely show a hardcoded use case. I'll come back to this Monday.

web/src/styles/legacy.css
```diff
@@ -32,8 +32,11 @@
     'avenir next', avenir, 'helvetica neue', helvetica, roboto, noto, 'segoe ui',
     arial, sans-serif;
 
-  --font-size-xl: 1.25rem; /* 20px */
-  --line-height-xl: 1.875rem; /* 30px */
+  --font-size-xl: 1.5rem; /* these need to be properly assigned by Viz Dee */
+  --line-height-xl: 2rem;
```
hbillings: @jenniferthibault These are just stubbed in values to get the state title in the sidebar closer to the size reflected in your Sketch screenshot. Let me know if there are real values you'd like me to plug in!

web/src/containers/Sidebar.js
```diff
@@ -1,219 +1,215 @@
...
+                width="40"
+                height="40"
+              />
+              <h1 className="text-xl">
+                {place.name} <br />
+                {t('title', { year: '2018' })}
+              </h1>
+            </div>
+            <VerticalNav
+              selectedId={this.props.selectedId}
```
hbillings: @mgwalker I think I need a pair to try to figure out what I'm doing wrong here 😞

web/src/containers/StateDashboardSidebar.js
```diff
@@ -13,36 +13,39 @@ class Sidebar extends Component {
...
+            <div className="flex items-center">
               <img
                 src={`/static/img/states/${place.id}.svg`}
                 alt={place.name}
                 className="align-middle mr2"
                 width="40"
                 height="40"
               />
-              <h1 className="m0 blue h3 light caps line-height-2">
+              <h1 className="text-xl">
```
jenniferthibault: Is this a class from the new system or old?

web/src/styles/legacy.css
```diff
@@ -32,8 +32,11 @@
     'avenir next', avenir, 'helvetica neue', helvetica, roboto, noto, 'segoe ui',
     arial, sans-serif;
 
-  --font-size-xl: 1.25rem; /* 20px */
-  --line-height-xl: 1.875rem; /* 30px */
+  --font-size-xl: 1.5rem; /* these need to be properly assigned by Viz Dee */
```
jenniferthibault:
```suggestion
  --font-size-xl: 1.3125rem; 
```

web/src/styles/legacy.css
jenniferthibault:
```suggestion
  --line-height-xl: 1.7rem;
```

jenniferthibault: Added what I'm using in the design files (which are the settings for an [H3](https://design.cms.gov/style/typography/) in the new CMS design system). I'm not sure if these will be future-proof if the `text-xl` class is also from that system, but should work ok for now?

web/src/containers/StateDashboardSidebar.js
hbillings: This is a class we made up when we were trying to come up with a logical way to break text sizes away from their heading levels (so we could do away with classes named `h2`, for instance).

web/src/containers/StateDashboardSidebar.js
```diff
@@ -13,36 +13,39 @@ class Sidebar extends Component {
     const { apds, place } = this.props;
 
     return (
-      <div className="ds-l-col--3">
-        <aside className="site-sidebar bg-teal">
+      <div className="ds-l-col--3 bg-white">
+        <aside className="site-sidebar">
           <div className="xs-hide sm-hide">
-            <div className="px2 py3 lg-px3 lg-py4 bg-white flex items-center">
+            <div className="flex items-center">
```
jenniferthibault:
```suggestion
            <div className="flex items-center ds-u-border-bottom--1 ds-u-padding-y--2 ds-u-margin-bottom--4">
```

web/src/containers/Sidebar.js
```diff
@@ -1,219 +1,236 @@
...
-            <Btn kind="outline" extraCss="bg-white blue">
-              {t('sidebar.savePdfButtonText')}
-            </Btn>
+    ];
+
+    return (
+      <div className="ds-l-col--3 bg-white">
+        <aside className="site-sidebar">
+          <div className="xs-hide sm-hide">
+            <div className="flex items-center">
```
jenniferthibault:
```suggestion
            <div className="flex items-center ds-u-border-bottom--1 ds-u-padding-y--2 ds-u-margin-bottom--4">
```

jenniferthibault: Sorry @hbillings, I didn't realize there were two!

web/src/containers/StateDashboardSidebar.test.js
```diff
@@ -33,8 +33,8 @@ describe('State dashboard sidebar component', () => {
   });
 
   test('selects an apd', () => {
-    const component = shallow(<Sidebar {...props} />);
-    component.find('SidebarLink[hash="1"]').simulate('click');
+    const component = mount(<Sidebar {...props} />);
+    component.find('VerticalNavItem[id="1"]').simulate('click');
     expect(props.selectApd.calledWith(1)).toBeTruthy();
```
hbillings: This is failing with a return value of `false`. I'm not entirely sure we need to be testing this, now that it's not a direct child of `StateDashboardSidebar`, but it does seem like we should be testing `selectApd` somehow? @mgwalker, I'd love your opinions on this.

web/src/containers/Sidebar.js
```diff
@@ -1,219 +1,227 @@
...
+              />
+              <h1 className="text-xl">
+                {place.name} <br />
+                {t('title', { year: '2018' })}
+              </h1>
+            </div>
+            <VerticalNav
+              selectedId={this.state.selectedId}
+              items={links}
+              onLinkClick={this.handleSelectClick}
```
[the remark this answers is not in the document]
hbillings: It _should_ get called when top-level items (with no children) are rendered, but for clarity, I ripped this out in favor of adding `onClick` events to each item explicitly.

web/src/containers/Sidebar.js
```diff
@@ -1,219 +1,227 @@
...
 
-const linkGroup1 = [
-  {
-    id: 'apd-state-profile',
-    name: t('apd.stateProfile.title'),
-    sub: [
+class Sidebar extends Component {
+  state = { selectedId: 'apd-state-profile-office' };
+
+  handleSelectClick = (id) => this.setState({ selectedId: id.id });
```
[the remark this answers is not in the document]
hbillings: @mgwalker I'm not entirely sure I understand what you're asking -- are you suggesting passing `id.id` into the function so that the state setter reads `setState({ selectedId: id })`? Or am I missing the point?

web/src/containers/Sidebar.js
```diff
@@ -1,219 +1,227 @@
...
+    } = this.props;
+
+    let activityItems = activities.map((a, i) => ({
+      id: a.key,
+      url: `#${a.anchor}`,
+      label: t(`sidebar.titles.activity-${a.name ? 'set' : 'unset'}`, {
+        number: i + 1,
+        name: a.name
+      }),
+      onClick: (evt, id) => this.handleSelectClick({id})
```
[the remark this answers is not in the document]
hbillings: oh, nvm my previous comment 😅. yes! that seems to work -- I'm not sure why I did it this way. 😬

web/src/containers/StateDashboardSidebar.js
```diff
@@ -13,36 +13,39 @@ class Sidebar extends Component {
...
+              selectedId="state-apds"
+              items={[
+                {
+                  label: `${place.name} APDs`,
+                  id: "state-apds",
+                  items: apds.map(apd => (
+                    {
+                      id: `${apd.id}`,
+                      label: `${apd.years.join(', ')} ${place.id.toUpperCase()} APD`,
+                      url: 'javascript:void(0);',
```
mgwalker:
```suggestion
                      url: 'javascript:void(0);', // eslint-disable-line no-script-url
```

This `javascript:void(0);` seems like the cleanest way to get the nav to render an `<a>` without the link sabotaging our intended behavior.

jeromeleecms:
@hbillings This is good to go + excited to see this one go in. Thanks for explaining that the stickiness of the sidebar is a separate issue!

PR #1322, opened 2019-02-26
Alert the user when navigating away could cause them to lose changes

### LINKED ISSUE #1237: Issue warning to user if APD has been edited, prior to leaving the APD creation page
From the conversation in #1226, the app currently tracks whether or not changes have been made to the  app content and a Boolean `dirty` which indicates if the content of the APD has changed. 

If a user tries to leave the app before saving & `dirty` reports that the APD has changed, than the system should prompt the user that leaving the page will result in changes not being saved.

This functionality should only be on if the APD is in draft/editable. It should not be available to users once the APD has been submitted or a decision on the APD has been made.

### This task is done when...
 - [ ] Any action that navigates away from the APD creation page should issue a warning that the user will save any unsaved changes
 - [ ] Warning is consistent with design system

mgwalker:
Navigation events that could cause the user to lose unsaved changes fall into two broad types:

## Browser events

These are initiated by the user interacting directly with the browser's UI, such as reloading the page or typing something new into the navigation bar.  This can also happen if they click a link on our site that goes to an external site, but I don't think we have any of those.  Anyway, browsers provide the [beforeunload](https://developer.mozilla.org/en-US/docs/Web/Events/beforeunload) event that allows us to catch when that happens, interrupt the navigation, and prompt the user if they're sure.  If they click ok, the navigation proceeds.

screen shot 2019-02-25 at 9 04 46 pm

Be aware that the specification for this event allows browsers to ***ignore*** it, so there's no guarantee that we can actually prompt the user in this case.  Furthermore, the spec allows us to define custom text to go in the dialog prompt but does not require browsers to use it - Chrome does not, for example (in the screenshot above, that text is from Chrome and we can't change it).  That said, this PR does add a handler for that event to allow us to handle it in the cases where we can.

## App events

These are when the user navigates using the app itself, such as clicking the dashboard button or inside the navigation sidebar.  For these events, we have total control.  In this case, if they are looking at an APD and it is dirty, we will prompt them when they navigate away from the APD (e.g., the dashboard button), but not when they navigate to a different section of the APD (e.g., the sidebar nav).  We can set the prompt text, but we can't change the button text.

screen shot 2019-02-25 at 9 00 51 pm

- navigating away from an APD while the APD has unsaved changes will prompt the user
- closes #1237


web/src/containers/ApdApplication.js
```diff
@@ -16,50 +16,101 @@ import { selectApdOnLoad } from '../actions/apd';
...
-  place,
-  selectApdOnLoad: dispatchSelectApdOnLoad
-}) => {
-  if (!apdSelected) {
-    dispatchSelectApdOnLoad('/apd');
-    return <Redirect to="/" />;
+const unsavedPrompt =
+  'You have unsaved changes to your APD. Are you sure you want to change pages?';
+
+class ApdApplication extends Component {
```
mgwalker: We have to turn it into a class-based component to hook the lifecycle events. 🙁

jenniferthibault:
What types of things do we have control over in the experience of App events @mgwalker ? That would help me get a sense of what to review for.

mgwalker:
As far as navigation events, we can control whether they happen at all, whether we display some content to the user before doing the navigation (the easiest way is with the prompt shown in the second screenshot above - it's built into the routing/navigation library we use), and where it goes.  We could even intervene and change the destination.  There's not much we can't do.

I think the only events this story is concerned with are the ones that would cause the APD screen to go away, and I ***think*** there are only 5 ways that can happen:
- user clicks the "logout" button (app event - and I just realized that one is unhandled 😬)
- user clicks the "state dashboard" button (app event)
- user clicks the browser back button (app event - surprise! - because of how we do internal app routing)
- user types in a new address in the address bar (browser event)
- user closes the tab/browser window (browser event)

jeromeleecms:
@mgwalker Do you want to us to hold off on reviewing this PR until you've had a chance to fix the user clicks the "logout" button app event? Looks like the recent discussion on #1237 indicates the team is onboard with this as the right direction for the pilot. The other 4 events were tested successfully, but could only check in Firefox.

mgwalker:
That probably makes sense @jeromeleecms.  I'm going to work on the logout piece now.  🤞 it won't be too hard...

mgwalker:
Now works on the logout button too (once it's deployed, anyway)

web/src/containers/Logout.js
```diff
@@ -0,0 +1,22 @@
+import PropTypes from 'prop-types';
```
mgwalker: An empty page that simply fires the logout action.  This lets us create a route for logging out, and we can intercept routing behavior to alert about unsaved changes.  This is much easier than trying to modify the logout behavior to be internally conditional.

web/src/containers/TopBtns.js
```diff
@@ -5,13 +5,12 @@ import { connect } from 'react-redux';
...
 
 import { t } from '../i18n';
-import { logout } from '../actions/auth';
 import Btn from '../components/Btn';
 
 class TopBtns extends Component {
   handleLogout = e => {
     e.preventDefault();
-    this.props.logout();
+    this.props.pushRoute('/logout');
```
mgwalker: Update the top nav button to route to `/logout` instead of calling the logout action directly

web/src/routes.js
```diff
@@ -2,13 +2,15 @@ import NoMatch from './components/NoMatch';
...
 import Login from './containers/Login';
+import Logout from './containers/Logout';
 
 import StateDash from './containers/StateDashboard';
 
 const routes = [
   { path: '/', component: StateDash, exact: true, nonPrivate: false },
   { path: '/apd', component: ApdApplication, exact: true, nonPrivate: false },
   { path: '/login', component: Login, nonPrivate: true },
+  { path: '/logout', component: Logout, nonPrivate: false },
```
mgwalker: Make the `/logout` path private, so when you logout, you land on that page, get logged out, and immediately get routed back to `/login`

web/src/containers/Login.js
```diff
@@ -27,7 +27,10 @@ class Login extends Component {
     const { username, password } = this.state;
 
     if (authenticated) {
-      return <Redirect to={from} />;
+      if (from.pathname !== '/logout') {
```
mgwalker: If we landed on the login page from the logout page, we don't want to redirect back to the logout page.  That'd be a nasty loop!

jameshupp:
I did a little searching for good in-app movement text. Here's a bit of the background thought I put into it. I'll propose the actual text change in-line.

Jerome concept: _You have unsaved changes to your APD. Are you sure you want to change pages?_

How Facebook does it (I assume they research the heck out of this): _You haven’t finished your post yet. Do you want to leave without finishing?_

[N/N recommendations](https://www.nngroup.com/articles/confirmation-dialog/): Be specific about what the action is and why it matters; don’t ask if they’re sure; and use very descriptive buttons.

But Greg’s writeup says we can’t change the button text.

web/src/containers/ApdApplication.js
```diff
@@ -16,50 +16,101 @@ import { selectApdOnLoad } from '../actions/apd';
...
-const ApdApplication = ({
-  apdSelected,
-  place,
-  selectApdOnLoad: dispatchSelectApdOnLoad
-}) => {
-  if (!apdSelected) {
-    dispatchSelectApdOnLoad('/apd');
-    return <Redirect to="/" />;
+const unsavedPrompt =
+  'You have unsaved changes to your APD. Are you sure you want to change pages?';
```
jameshupp: You have unsaved changes to your APD. Do you want to leave this page without saving?

"Are you sure" prompts have a glaze-over effect. People see them too much and burying the action behind them sometimes leads people to not think it through.

mgwalker:
@jameshupp To clarify, we can set the button text for "app events" as described in the PR description.  For "browser events," we can _suggest_ button text to the browser, but we can't guarantee what the browser will actually display.

jeromeleecms:
If you are curious how it appears in Firefox: 

image

PR #1326, opened 2019-02-26
Add an API route for the current user to edit their own account

### LINKED ISSUE #1319: Create an API endpoint for users to modify their own account
Need a `PUT /me` to allow users to modify their own accounts, including (and especially!) their password.

### This task is done when...
 - [ ] there's an API endpoint where users can change their name, email, position/title, phone number, and password

mgwalker:
Adds the `PUT /me` route to compliment the `GET /me` route.  This allows the currently-logged-in user to modify parts of their own account:

- email
- name
- password
- phone
- position

- closes #1319 


api/auth/serialization.js
```diff
@@ -25,6 +25,9 @@ module.exports.deserializeUser = async (
       done(null, {
         username: user.get('email'),
         id: user.get('id'),
+        name: user.get('name'),
+        phone: user.get('phone'),
+        position: user.get('position'),
```
mgwalker: Return these as part of the user object, so when we add a UI for users to modify their own accounts, they'll be able to see what values are here already.  Because...  I don't know why these weren't already returned.  🤷‍♂️

api/db/user.js
```diff
@@ -14,6 +14,14 @@ module.exports = (zxcvbn = defaultZxcvbn, bcrypt = defaultBcrypt) => ({
       return this.hasOne('state', 'id', 'state_id');
     },
 
+    format(attr) {
+      if (this.hasChanged('password')) {
+        // eslint-disable-next-line no-param-reassign
+        attr.password = bcrypt.hashSync(attr.password);
```
mgwalker: Hash the password as it goes into the database, if if has changed.  If it hasn't changed, then it's already hashed, and hashing it again would be problematic.  So don't do that.

api/db/user.js
```diff
@@ -70,7 +78,6 @@ module.exports = (zxcvbn = defaultZxcvbn, bcrypt = defaultBcrypt) => ({
         }
 
         logger.silly('password is sufficiently complex; hashing it');
-        this.set({ password: bcrypt.hashSync(this.attributes.password) });
```
mgwalker: This was hashing the password in the `validate()` function, which seems to make sense, except `save()` also calls `validate()` - so if you validate the model independent of saving it, you end up double-hashing the password.  And then you can never log in again.  So...  That's no good.

api/routes/me/openAPI.js
```diff
@@ -1,8 +1,56 @@
...
+      type: 'string',
+      description: 'Names of system activities this user can perform'
+    }),
+    id: {
+      type: 'number',
+      description: `User's unique ID, used internally and for identifying the user when interacting with the API`
+    },
+    name: {
+      type: 'string',
+      description: `The user's name, if defined`
```
hbillings: This is real name, not username, right?

mgwalker: Correct

PR #1336, opened 2019-02-28
Feature to allow APDs to be archived

### LINKED ISSUE #1021: Add in a way to delete an APD draft
If an APD is started by accident or no longer needed, there should be a way to remove that from the state user's queue by deleting it.

mgwalker:
Provides a way for users to archive a draft APD, and hides archived APDs from view.  This PR calls the action "delete" even though the data in the database is ***not*** deleted.

- adds `DELETE /apd/:id` endpoint that changes the APD's status to `archived`
- archived APDs are filtered out of the application state so they don't show up on the dashboard
- adds a delete button on the dashboard that prompts the user to confirm their action, then calls the API endpoint
   - uses the CMS button style (🎉)

screen capture

Closes #1021


mgwalker:
The original intent was to _delete_ the APDs but I was concerned about a future team not removing the delete functionality when/if eAPD becomes a system of record, so we compromised on archival.  Calling it "delete" serves the original purpose and doesn't create functionality that could be problematic in the future.  Win-win?

I think eventually yes, users may want to recall an archived APD, but for now, since we're calling it "delete," probably not.  But that's kind of some wild guessing about what users will want.  😛

PR #1342, opened 2019-03-01
Update dependencies

mgwalker:
We have several out-of-date dependencies.  Two of them have vulnerabilities listed as high-severity.

- updates API dependencies
   - updates code as necessary (some lint rules got stricter)
- removes unused API dependencies
   - removes `lodash.pick` in favor of a little piece of custom code
- updates web app dependencies
   - updates code as necessary (Babel 7 has a different scoped package for its polyfill library; fortawesome renamed packages and got rid of default exports)

### To be done in separate bodies of work, all frontend:
- update `webpack` 😱
   - will also update the plugins and related/reliant stuff (`css-loader`, `cssnano`, `postcss`, `postcss-cli`, `postcss-custom-properties`, `postcss-import`, `postcss-loader`, `postcss-url` [good grief with all the postcss stuff], `uglifyjs-webpack-plugin`, and `webpack-dev-server)
- update `eslint` (tried this and there were a couple hundred lint issues, so I backed it out for now)
   - will also update `babel-eslint`, `eslint-config-airbnb`, and `eslint-config-prettier`
- switch from deprecated `react-router-redux` to `connected-react-router`
   - will also update `redux` and `react-redux`

- [x] All tests are passing (if tests are updated, that should be a smell that something is wrong)
- [x] The change has been documented - package.json


api/db/base.js
```diff
@@ -62,7 +61,9 @@ const defaultSyncChildren = async (
 // These properties are added to model instances.
 const instanceExtension = (orm, models) => ({
   pickUpdateable(raw) {
-    return pick(raw, this.static.updateableFields);
+    return Object.entries(raw)
+      .filter(([key]) => this.static.updateableFields.includes(key))
+      .reduce((obj, [key, value]) => ({ ...obj, [key]: value }), {});
```
hbillings: I don't really understand what this change does/why this change exists, but I'm assuming it has something to do with updating something...

api/package.json
```diff
@@ -66,17 +63,17 @@
...
-    "eslint": "^4.19.1",
-    "eslint-config-airbnb-base": "^12.1.0",
-    "eslint-config-prettier": "^2.10.0",
-    "eslint-plugin-import": "^2.14.0",
-    "jest": "^23.6.0",
-    "nodemon": "^1.18.6",
+    "eslint": "^5.14.1",
+    "eslint-config-airbnb-base": "^13.1.0",
+    "eslint-config-prettier": "^4.1.0",
+    "eslint-plugin-import": "^2.16.0",
```
hbillings: Flagging because `eslint` was one of the things you mentioned would be done in another issue.

web/src/components/Icons.js
```diff
@@ -1,23 +1,42 @@
...
+// import faClock from '@fortawesome/free-regular-svg-icons/faClock';
+// import faCode from '@fortawesome/free-solid-svg-icons/faCode';
+// import faCog from '@fortawesome/free-solid-svg-icons/faCog';
+// import faEdit from '@fortawesome/free-solid-svg-icons/faEdit';
+// import faExclamationTriangle from '@fortawesome/free-solid-svg-icons/faExclamationTriangle';
+// import faPlusCircle from '@fortawesome/free-solid-svg-icons/faPlusCircle';
+// import faPrint from '@fortawesome/free-solid-svg-icons/faPrint';
+// import faSpinner from '@fortawesome/free-solid-svg-icons/faSpinner';
+// import faHelpSolid from '@fortawesome/free-solid-svg-icons/faQuestionCircle';
+// import faSignOut from '@fortawesome/free-solid-svg-icons/faSignOutAlt';
```
hbillings: What's goin' on with these suckers?

api/db/base.js
mgwalker: Just getting rid of `lodash.pick` as a dependency.  It drags in a bunch of code (it handles a lot more use cases than we need) and the lodash family of packages are pretty routinely coming up in Snyk vulnerability reports.  So...  I figured just get rid of it.  😛

api/package.json
mgwalker: Good catch!  I should have been clearer - updating eslint on the frontend.  On the backend there were maybe a half dozen things it wanted fixed, so I did it.  On the frontend, though...  woof.

web/src/components/Icons.js
mgwalker: Oops.  I commented them out as I replaced them, then forgot to delete them afterwards.  Will fix that...  😬

PR #1352, opened 2019-03-05
Update webpack

mgwalker:
Updates webpack to version 4.x, which turned out to be a lot easier than I expected - in large part because of updates we'd made in order to support Sass previously.  Yay, happy accidents!

- updates to webpack 4, webpack-dev-server 3
- fixes vulnerabilities introduced via webpack-dev-server


web/package.json
```diff
@@ -4,7 +4,7 @@
   "description": "APD app for CMS HITECH program",
   "main": "app.js",
   "scripts": {
-    "storybook": "start-storybook -p 9001 -c .storybook",
+    "storybook": "start-storybook --ci -p 9001 -c .storybook",
```
mgwalker: We went from storybook 3 to 4.  They changed it so the `start-storybook` command launches your default browser unless you use this `--ci` option.  So to keep things basically the same for us, I added the flag.  But also I hate when CLI tools open my browser. 🤷‍♂️

web/package.json
```diff
@@ -100,13 +98,13 @@
     "@babel/plugin-transform-runtime": "^7.3.4",
     "@babel/preset-env": "^7.0.0",
     "@babel/preset-react": "^7.0.0",
-    "@storybook/react": "^3.4.11",
+    "@storybook/react": "^4.1.13",
```
mgwalker: 5.0.0 dropped today (what a coinkydink!).  I tried to go ahead and update to that, but it had some weird issues with core.js dependency version mismatches on _something_ downstream.  People have had luck getting it to work by adding a direct dependency on core.js, but...  I dunno, I opted not to use an hours-old version that didn't work out of the box.  That seems like what patch versions are for.  😛

web/package.json
```diff
@@ -117,29 +115,31 @@
...
     "react-test-renderer": "^16.8.3",
     "redux-logger": "^3.0.6",
     "redux-mock-store": "^1.5.3",
+    "resolve-url-loader": "^3.0.1",
     "sass-loader": "^7.1.0",
-    "sinon": "^7.2.5",
+    "sinon": "^7.2.7",
     "style-loader": "^0.23.1",
     "topojson": "^3.0.2",
-    "uglifyjs-webpack-plugin": "^1.3.0",
```
mgwalker: Tree-shaking and minification are built into webpack now.  🎉

web/src/actions/apd.test.js
```diff
@@ -167,7 +167,7 @@ describe('apd actions', () => {
       const store = mockStore();
       fetchMock.onDelete('/apds/apd-id').reply(400);
 
-      store.dispatch(actions.deleteApd('apd-id')).then(() => {
+      return store.dispatch(actions.deleteApd('apd-id')).then(() => {
```
mgwalker: These tests were buggy because if the promise rejected - which it would sometimes due - Jest didn't know about it because the promise hadn't been returned.  So you'd see log output that it failed but it'd still exit successfully.  By returning the promise, we're letting Jest track its outcome.  Whoops on my part.  😬

web/src/components/Icons.js
```diff
@@ -11,14 +11,18 @@ import {
   faCog,
   faEdit,
   faExclamationTriangle,
-  faHelpSolid,
+  faQuestionCircle as faHelpSolid,
   faPlusCircle,
   faPrint,
   faSpinner,
-  faSignOut
+  faSignOutAlt as faSignOut
```
mgwalker: More whoops on my part - there aren't `faHelpSolid`, `faSignOut`, or `faHelp` components in these libraries.  We were importing and renaming them previously, but I didn't notice that when I switched to the newer version and import format.

mgwalker:
You think it's not a huge pain, but you get to rebuild your docker container!  :yay:

PR #1353, opened 2019-03-05
Update eslint and base rules

mgwalker:
Update eslint along with the base rules we start with.  This required a few small code changes for newly-introduced rules.

- updated eslint


web/.eslintrc.json
```diff
@@ -17,6 +17,8 @@
         }
       }
     ],
+    "jsx-a11y/label-has-associated-control": ["error", { "assert": "either" }],
```
mgwalker: This `label-has-associated-control` rule is new, and it replaces an older deprecated rule called `jsx-a11y/label-has-for`.  The default behavior is to require `<label>` elements to both wrap the inputs they apply to ***AND*** have a `for` attribute.  We felt like that was overkill (and also difficult), so we modified the older rule.  This is just updating our config to similarly modify the new rule.  The old rule is deprecated but still in force, so I couldn't remove our modification for it yet.

web/src/components/Collapsible.js
```diff
@@ -23,11 +23,14 @@ class Collapsible extends Component {
       children,
       id,
       nested,
+      open: openProp,
       sticky,
       title,
       onChange
     } = this.props;
-    const isOpen = onChange ? this.props.open : this.state.open;
+    const { open: openState } = this.state;
```
mgwalker: This is the main theme of the code changes.  Our eslint rules now require all React `state` and `prop` variables to be destructured - no more `this.props.blahblah`.

web/src/containers/admin/CreateAccount.js
```diff
@@ -23,15 +23,16 @@ class CreateUser extends Component {
   };
 
   toggleShowPassword = () => {
-    this.setState({ showPassword: !this.state.showPassword });
+    this.setState(prev => ({ showPassword: !prev.showPassword }));
```
mgwalker: By happenstance, this new rule also caught this bad practice: you shouldn't access `this.state` in a call to `setState` because those calls are queued and run asynchronously.  Eg, if `showPassword` is false, then you had this:

```js
toggleShowPassword();
toggleShowPassword();
```

...you'd then expect `showPassword` to still be false (toggled to true then back to false).  But in reality, both calls to `toggleShowPassword()` capture `this.state` before it has been updated, so they both start with `showPassword` false and change it to `true`, so the end result is that `showPassword` is true.

If you queue up functions instead of new objects, they are always called in the order they were triggered, and they are called with the most recent version of state after all previous updates in the cycle were applied.

web/.eslintrc.json
hbillings: Thanks for the clarification!

web/src/components/Collapsible.js
hbillings: Thanks for trudging through this and updating it!

web/src/containers/ApdSummary.js
```diff
@@ -21,18 +21,21 @@ class ApdSummary extends Component {
...
-    } = this.props.apd;
+      apd: {
+        years,
+        yearOptions,
+        programOverview,
+        narrativeHIT,
+        narrativeHIE,
+        narrativeMMIS
+      }
+    } = this.props;
```
hbillings: Ooh, that's the first time I've seen a nested object destructured. Interesting!

web/src/containers/Notification.js
```diff
@@ -7,12 +7,16 @@ import { processQueue, closeNotification } from '../actions/notification';
 
 class Notification extends Component {
   handleClose = (_, reason) => {
-    if (reason === 'clickaway') return;
-    this.props.closeNotification();
+    if (reason === 'clickaway') {
+      return;
+    }
+    const { closeNotification: action } = this.props;
+    action();
```
hbillings: I don't quite understand this one. Why is it `action()` instead of `closeNotification()`?

web/src/containers/StateDashboardSidebar.test.js
```diff
@@ -32,6 +36,20 @@ describe('State dashboard sidebar component', () => {
     expect(component).toMatchSnapshot();
   });
 
+  test('selects apds', () => {
+    const component = shallow(<Sidebar {...props} />);
+    const topItems = component.find('VerticalNav').prop('items');
```
hbillings: Thanks for adding this back in! (And also for teaching me that that you can target props passed to a component!)

web/src/containers/admin/CreateAccount.js
hbillings: > If you queue up functions instead of new objects, they are always called in the order they were triggered, and they are called with the most recent version of state after all previous updates in the cycle were applied.

Suddenly, mystery bugs from previous projects make sense...

web/src/containers/Notification.js
mgwalker: A variable named `closeNotification` exists in an upper scope and eslint will complain about that, so I renamed it when I destructured it out of props.  I went with `action` because it represents a redux action and I hoped that'd make it clear what it is, but I reckon it did not.  😬

web/src/containers/StateDashboardSidebar.test.js
mgwalker: Honestly I only added it because I introduced a bug that broke clicking on APDs.  😬

Issue #1306, opened 2019-02-21
Apply CMS design system styles to user management flow


jenniferthibault:
So that users of the eAPD system have a consistent visual and interaction experience from start to finish in the app (and maintain a high level of confidence in its authenticity), we need to use the CMS design system styles in the user management flow created in #1253, #1254, and #1276.

### This task is done when...
 - [x] The experience for Logging in is designed using the CMS design system
 - [x] The experience for Logging out is designed using the CMS design system
 - [ ] A new issue has been created passing designs to implementation

lauraponce:
Here is a first pass at the main **Log in** page for how it could look with the CMS design system:

![log in page new](https://user-images.githubusercontent.com/21222704/54007797-7c2f3580-4118-11e9-9202-30f938985e68.png)

### Incorrect email or password:
![log in page new copy](https://user-images.githubusercontent.com/21222704/54007801-82251680-4118-11e9-9a81-12461e8d2c2e.png)

### Tried to log in with an empty field:
![log in page new copy 2](https://user-images.githubusercontent.com/21222704/54007817-8f420580-4118-11e9-86b4-641a402985ed.png)

@jenniferthibault 👀& feedback for this layout?
@jameshupp do you have thoughts on `log in` vs `sign in` (or feedback on any of this)?

jeromeleecms:
@jameshupp Feel free to updated #1308 if you provide @lauraponce feedback on the error messages here!

lauraponce:
@jenniferthibault and I talked through the login experience and here are the desigs that we landed on:

### Log in page:
![Log in page new](https://user-images.githubusercontent.com/21222704/54058997-623f3280-41ac-11e9-8fc8-ad43e5253939.png)

### Log in page with incorrect email/password error state:
![Log in page new - error](https://user-images.githubusercontent.com/21222704/54058995-61a69c00-41ac-11e9-92a4-187f019ed8fb.png)

### Log in page with empty field error state:
![Log in page new error 2](https://user-images.githubusercontent.com/21222704/54058996-61a69c00-41ac-11e9-8ea7-346f8fc5e64e.png)

Questions for @jeromeleecms & @NAretakis -
- To limit the scope for the pilot, @jenniferthibault and I talked about the experience for logging out and felt like it would make sense to just take them back to the main Log in page when they log out so they can get back in if they need to. Does that feel like the right call for now?
[...]

Issue #1274, opened 2019-02-08
Create a UX for users to change their passwords


mgwalker:
Users need to be able to change their passwords.  F'real.

There is no API support for this, either.

### This task is done when...
 - [ ] users can change their own passwords

jenniferthibault:
@mgwalker got this started in #1357. I had a lot of design notes, so pulling that discussion into here in case any of it needs more discussion:

> - Adds a top button to view your own account - in the preview link, clicking that button will be the only way to get to the page for editing your account

>  <img width="955" alt="screen shot 2019-03-06 at 6 38 46 pm" src="https://user-images.githubusercontent.com/1775733/53923846-55e99700-403f-11e9-8f8c-0463dd36f949.png">

This makes sense for now since our available functionality is limited. Could we call it `Manage account` to mirror [Login.gov's IA](https://login.gov/)? I'd wager that theirs is informed by some data, which is more than we can say for ours right now.

____

> - which takes you to this fancy-pants account editor (the pants aren't really that fancy) - using the *actually* fancy-pants CMS design system buttons and alert box

>  <img width="529" alt="screen shot 2019-03-06 at 6 41 54 pm" src="https://user-images.githubusercontent.com/1775733/53923923-921cf780-403f-11e9-84bc-ea8d4985d76f.png">

The [Form label with hint text](https://design.cms.gov/components/form-label/) feels like a better fit here instead of the alert because it's specific to that field, and lots smaller.

This components in the mockup I've proposed below also include the:
- [Text field component](https://design.cms.gov/components/text-field/) ,
- [Masked field component](https://design.cms.gov/components/masked-field/) for the Phone number field, 
- [Checkbox component](https://design.cms.gov/components/choice/) at the small size variant. 
-   `ds-c-button--transparent` [modifier](https://design.cms.gov/components/button/) for the "Cancel" button (which replaces  "Back")

![account-manage](https://user-images.githubusercontent.com/11636908/53928249-673e9d80-4057-11e9-834e-78bb9819012d.png)

______
> - and adds a new password meter

>  ![passwords](https://user-images.githubusercontent.com/1775733/53924072-2edf9500-4040-11e9-84e1-d74da8947dde.gif) 

In order to avoid the jump in the experience when the password meter hides/displays, we could instead have it always displayed (but in "empty" state) until the user starts typing in the field.

[...]

Issue #1268, opened 2019-02-07
Make dashboard page reflect eAPD Builder concept using CMS Design System styles


jenniferthibault:
_Part of #1250_

For completion of Layer 1, the site will be positioned as an eAPD builder tool. In order to get the dashboard page to reflect this functionality accurately, we could make slight temporary tweaks to reign back some of the status & interaction features we were thinking about previously. 

While doing this, we also have the opportunity to update the page's styles and layouts to use the [CMS Design System](https://design.cms.gov/) styles.

### This task is done when...
 - [x] We agree on a new mockup of the dashboard page
 - [ ] A new issue has been created for development changes for the concept.

 **Current dashboard page:**
![screen shot 2019-02-06 at 11 08 34 pm](https://user-images.githubusercontent.com/11636908/52389985-2c4f4700-2a64-11e9-9d90-885d8067a34b.png)

jenniferthibault:
A few ideas:
- Given that we won't have a large backlog of APDs for the pilot & layer 1, I think we can safely drop the side nav here, and only use one top navigation.
- If we add a little more structure to the text, the instructions are easier to spot!
- We can use the [Review pattern](https://design.cms.gov/patterns/review/) from the CMS design system for the list of documents here (woo!) I'm not sure if it would let us customize it to use the document icon with it – if that adds complexity I'd love to discuss other tactics @hbillings 
- The design system doesn't stipulate a standard footer, so I grabbed the basics from a quick scan of other HHS/CMS sites. We can edit this in the future once we have more clear needs.
- The gray rectangle is a placeholder for either a site title or branding, when we know what that should be


![layer 1 dashboard - design system](https://user-images.githubusercontent.com/11636908/53917882-9511eb00-4033-11e9-9e87-0c4d6e2d9ea3.png)

Thoughts @NAretakis @jeromeleecms @lauraponce @hbillings ?

mgwalker:
Whoa, that is lovely.

jeromeleecms:
@jenniferthibault Nice and clean!

1) Header - do we need to have a consistent header after a user logs in to the app (I saw you indicated a one layer nav)? Right now, we have "Manage My Account", "My Dashboard", and "Logout" but if/when we add new apps how would the header scale? I couldn't find any design pattern in the CMS design system, but the USDS 2.0 beta has this guidance: https://v2.designsystem.digital.gov/components/headers/

[...]

Issue #1063, opened 2018-10-23
Inconsistent Browser Experience with CMS Network on AWS


jeromeleecms:
Using various browsers on the CMS AWS instance results in different behaviors with the App:

1) Chrome: Can log in, but can not view existing APDs or create a new one. James mentioned on standup that he was able to get the app to work on his computer by using an incognito window,
2) Firefox: Appears to work as intended
3) IE: Can login, but a refresh kicks the user back out to the login screen. There is also no ability to view existing APDs or create new ones.

### This task is done when...
 - [ ] criteria 1
 - [ ] criteria 2
 
 ** Don't forget to fill out the LABEL and MILESTONE before submitting. If the issue is for the CURRENT SPRINT, add it to the PROJECT. If the issue is a BUG, FUTURE ENHANCEMENT, or OUTSIDE OF THE CURRENT SPRINT, do not add it to the PROJECT (then delete this line) **

jeromeleecms:
Hmm.. not sure what step I missed for triage...? 

I saw you had a similar issue during standup on Monday, so I would consider this is a P1 bug since it affects you and I (at minimum) and we can not complete the experience. While this bug exists, we can not review or approve PRs that are deployed to the new environment (which affects this sprint). 

We could kick this down the road, but I prefer to deal with this now because of its implications to the project. Once we make the cut over to the CMS AWS, that also means our CMS stakeholders won't be able to use the app.

Thoughts?

jeromeleecms:
Good chat with @NAretakis (always!). 

For now, will keep on the board since I'm meeting with @mgwalker about this in the afternoon. If this turns into a heavy lift to resolve, will kick this into the backlog and will advise users who are experiencing this bug to use Firefox for the time being.

jeromeleecms:
Third party cookies are blocked, so authentication does not persist after the initial login. APDs do not show up because there is no cookie to pass back with each request to confirm the person is a valid, logged in user.

Per @mgwalker this will resolve itself when we have our own domain name since, currently, the amazonaws.com adddress is being recognized as a third party. No additional action can be taken. 

Mystery solved... for now.

PR #1357, opened 2019-03-07
Allow users to change their own passwords

mgwalker:
- adds a top button to view your own account - in the preview link, clicking that button will be the only way to get to the page for editing your account

  state dashboard with new manage account button

- which takes you to this fancy-pants account editor (the pants aren't really that fancy) - using the *actually* fancy-pants CMS design system buttons and alert box

  screen shot of new account management screen

- and adds a new password meter

  gif of new password meter


jenniferthibault:
@mgwalker I have a lot of notes here, so pulling them into the original issue #1274 in order to keep the PR review comments cleaner.  (I don't know if that's the right move, so please tell me what you prefer for the future)

NAretakis:
This is awesome, and all the fields work as advertised, including fancy pants password strength thingy. However, I get this message when I try to save. Is this something to worry about?
image

mgwalker:
Hmm, that's something to check into before we deploy it for sure.  If you're using your CMS laptop and Chrome or IE, it could be related to #1063.  If you're not, we should pair up next week and take a look at it.  If there's any wonkiness in there, it'd be nice to get it out before we deploy!

But back to #1063 for a second!  You'd be able to get a "success" message from the API when you log in so the app would let you see the dashboard and navigate to the "manage account" page, but it wouldn't save your authentication cookie so you wouldn't actually be able to do anything.  That popped into my head when I pulled up the logs from last night:

screen shot 2019-03-08 at 8 15 29 am

The red box at the top is a snip of the log message from the cloud.gov router in front of the API showing the endpoint that the browser requested - `PUT /me`, which is the endpoint for updating your own information.  The yellow box near the bottom is a snip of a log from the API itself.  The `jwt must be provided` bit is very telling - it means the authentication cookie is missing (our authentication cookie is [JWT token](https://jwt.io/)).  The next line is also from the API, from our authentication middleware, just saying the request is not authenticated.  And the timestamp lines up to about 10:30 MT last night, which I assume was you.

Back on the app side, it's looking for errors with messages.  The unauthenticated and unauthorized errors intentionally don't send messages (don't want to reveal anything that would be useful to an attacker, so we only send error messages to authenticated users), so the app doesn't find one and then it'd show the "unknown error" as a fallback.

jeromeleecms:
@mgwalker I was thinking the same thing, so I tested in Firefox this morning...

image

@NAretakis Were you trying to save in Chrome? Did your test APDs show up in your account (that's how we discovered #1063 was an issue)...

NAretakis:
That's it! Tested in Firefox and I was able to update all the fields, save them, log out, and log back in. I even changed my email address, and it worked logging in with the new email address. 🎉

mgwalker:
I talked to @kimberbat and she agrees we shouldn't be letting users change their own email addresses.  Instead, she suggested having a policy for how users can request their email address be changed, and once CMS approves, someone from CMS logs into an admin account and changes it.

I'll make the change on the frontend in this PR and I've opened an issue to make the backend fix (#1360)

jeromeleecms:
Seems like a separate issue, but once the password change is complete, the user only has two "escape" routes: 1) Click on the top left "CMS HITECH APD" or 2) cancel. The "Manage My Account" "State Dashboard", "Logout" buttons are not visible - but the scope of this PR is only for the change password UX, so that's good... I'm going to drop something in #1268 about the headers/buttons in a moment.

jenniferthibault:
Seeing this and #1268 I went one further and split the nav issue into #1366 

@mgwalker logistically for GitHub, what path forward do you recommend for this PR? Merge as is, then immediately figure out #1360 ? Or are you planning to stop progress toward #1274 and bring #1360 into this? Trying to figure out where/how I can help most.

mgwalker:
My thought is get this to satisfy #1274 only.  Once it's in, then we can fix any problems that arise with it.  I've already removed the email field from this form since that was trivial (#1360 is about updating the backend to not accept email changes).

I think once it's in line with your intentions from #1274, we can go ahead and merge it, then fix stuff we find later.

I know of 3 style issues that need to be fixed:
- the padding within the form is too small
- the margin above the header increases dramatically when an alert is present
- the "show password" checkbox floats too high

#1306 and #1367 can benefit from figuring those issues out, because however we solve it in one account management form, we should be able to just do the same thing again in the other account management forms.  I'm okay with either fixing this form now, or merging now and fixing all of the forms at once, later.

web/src/components/PasswordWithMeter.js
```diff
@@ -0,0 +1,103 @@
...
+
+  toggleShowPassword = () => {
+    this.setState(prev => ({ showPassword: !prev.showPassword }));
+  };
+
+  render() {
+    const { title, value, ...rest } = this.props;
+    const { showPassword, strength } = this.state;
+
+    let passwordQuality = 'Password is great';
```
jenniferthibault:
```suggestion
    let passwordQuality = 'Password strength: Great';
```

web/src/components/PasswordWithMeter.js
```diff
@@ -0,0 +1,103 @@
...
+    if (value.length === 0) {
+      passwordQuality = <span>&nbsp;</span>;
+    } else if (strength < 3) {
+      passwordQuality = 'Password is weak';
```
jenniferthibault:
```suggestion
      passwordQuality = 'Password strength: Weak';
```

web/src/components/PasswordWithMeter.js
```diff
@@ -0,0 +1,103 @@
...
+    } else if (strength === 3) {
+      passwordQuality = 'Password is good';
```
jenniferthibault:
```suggestion
      passwordQuality = 'Password strength: Good';
```

mgwalker:
I [added the class to the phone number field](https://github.com/18F/cms-hitech-apd/pull/1357/files#diff-99b50425a315510539aea057a99628daR91).  It's a little weird in React because it turns the `className` property into the HTML `class` property.  I'm not entirely clear why they had to rename it, but here we are.  🙂

Also fixed the password strength copy.  I'll schedule time with Heather when she's back to figure out those other styling issues.

jeromeleecms:
@mgwalker I did learn something from my self-paced React training marathon (slow and steady!) ... but since _class_ is a reserved word in JavaScript, they had to use _className_ with JSX so it would transpile correctly... also feel free to ignore this comment if that's not what you were referring to :)

Randomly signing in and now randomly signing off!

web/src/components/PasswordWithMeter.js
```diff
@@ -0,0 +1,105 @@
...
+
+          <TextField
+            hint="A strong password is at least 9 characters, not a commonly-used word or phrase, and not too similar to the person’s name or email address."
+            label={title || 'Password'}
+            name="password"
+            type={showPassword ? 'text' : 'password'}
+            value={value}
+            onChange={this.changePassword}
+          />
+          <div className={strengthClass}>
```
hbillings: I think this meter portion might need an ARIA label to tell screenreaders to ignore it. (Or to read out what the quality level is.) We should test.

web/src/components/PasswordWithMeter.js
```diff
@@ -0,0 +1,105 @@
...
+            <div
+              className="strength-meter-fill"
+              data-strength={value.length ? strength : 'empty'}
+            />
+          </div>
+          <div className={qualityClass}>{passwordQuality}</div>
```
hbillings: Let's pop this in `p` tags instead of a `div`.

web/src/containers/admin/MyAccount.js
```diff
@@ -0,0 +1,165 @@
...
+  editAccount = e => {
+    e.preventDefault();
+    const { editAccount } = this.props;
+    const { user } = this.state;
+    editAccount(user);
+  };
+
+  goBack = () => {
+    const {
+      history: { goBack }
```
hbillings: niiiiice

web/src/containers/admin/MyAccount.js
```diff
@@ -0,0 +1,165 @@
...
+      <Fragment>
+        <header className="clearfix mb2 px2 py1 bg-white">
+          <div className="left">
+            <Link to="/" className="btn px0 bold caps">
+              {t('titleBasic')}
+            </Link>
+          </div>
+        </header>
+
+        <div className="ds-l-col--5 ds-u-fill--white ds-u-margin-x--auto ds-u-margin-y--3 ds-u-padding--6 ds-u-radius">
```
hbillings: I got mocks from Jen that made me realize this needs to be `ds-l-col--8`. We also want to remove the padding class and make the next div down `ds-l-col--6`. This will give it one column of padding on either side.

Here's a screenshot of the Sketch file so you can see what I'm talking about:
Screen Shot 2019-03-15 at 6 14 34 PM

web/src/containers/admin/MyAccount.js
```diff
@@ -0,0 +1,165 @@
...
+          <div className="ds-u-margin-x--auto" style={{ maxWidth: '460px' }}>
```
hbillings: This max width can go away once we put the `ds-l-col--6` class on.

web/src/containers/admin/MyAccount.js
```diff
@@ -0,0 +1,165 @@
...
+            {!!success && <Alert variation="success">Changes saved</Alert>}
+
+            <h1 className="ds-h1">Manage account</h1>
```
hbillings: Did you figure out the padding issue going on here, or do you still want a pair?

web/src/styles/scss/password-meter.scss
```diff
@@ -0,0 +1,80 @@
+@import '@cmsgov/design-system-support/src/settings/index';
+
+.password-with-meter {
+  // inputs in the design system have a max width, so cap our container at that
+  // width too, to prevent floats and 100% widths from going too big
+  max-width: $input-max-width;
+
+  .password-with-meter--show-password label.ds-c-label {
+    margin: 24px 0 0 0;
```
hbillings: So it looks like the original issue of the checkbox being too high has to do with the fact that `TextField`s look to inherit a clearfix. If we change this rule to just `margin-top: 0;` and then add a `className` to `TextField` that overrides the clearfix utility, these will stay in sync if a user increases the text syze. (I'll bust up an suggest commit right after this.)

web/src/styles/scss/password-meter.scss
```diff
@@ -0,0 +1,80 @@
...
+
+.strength-meter {
+  position: relative;
+  height: 3px;
+  background: #ddd;
+  margin: 7px 0;
+  border-radius: 2px;
+
+  &:before,
+  &:after {
```
hbillings: okay this bit kind of blows my mind 0_o

web/src/containers/admin/MyAccount.js
```diff
@@ -0,0 +1,165 @@
...
+
+            <form onSubmit={this.editAccount}>
+              <TextField
```
hbillings: It looks like `TextField` supports an `ariaLabel` property that we could use to write something more descriptive for screen readers (I'm thinking, "Enter your full name" or "enter your user name," depending on what the name is supposed to be). We could do this with all of them, so it reads something like:
"Enter your full name.
"Please provide a phone number.
"Enter your position [within CMS?]"

web/src/components/PasswordWithMeter.js
```diff
@@ -0,0 +1,105 @@
...
+      'strength-meter-quality',
+      value.length > 0 ? 'visible' : 'hidden'
+    ]
+      .join(' ')
+      .trim();
+
+    return (
+      <div {...rest}>
+        <div className="password-with-meter">
+          <Choice
```
hbillings: I'm a little concerned at this coming before the associated `TextField` in the source. This means that screen readers are going to read the "Show Password?" checkbox prior to announcing that the user is about to enter a password. Maybe we don't use the `TextField` for this one so we can put things in a more sensible order?

web/src/containers/admin/MyAccount.js
```diff
@@ -0,0 +1,165 @@
...
+          </div>
+        </header>
+
+        <div className="ds-l-col--5 ds-u-fill--white ds-u-margin-x--auto ds-u-margin-y--3 ds-u-padding--6 ds-u-radius">
+          <div className="ds-u-margin-x--auto" style={{ maxWidth: '460px' }}>
+            {!!success && <Alert variation="success">Changes saved</Alert>}
+
+            <h1 className="ds-h1">Manage account</h1>
+
+            <form onSubmit={this.editAccount}>
```
hbillings: Let's throw all of these in a `fieldset` with a "manage account details" `legend` (maybe the legend gets the screen-reader-only class). That'll tell screen readers that all these fields are a group.

web/src/actions/admin.test.js
```diff
@@ -474,4 +474,135 @@ describe('admin actions', () => {
...
+    });
+
+    it('handles an invalid phone number', async () => {
+      const store = mockStore({ notification: { open: false, queue: [] } });
+
+      fetchMock.onPut('/me').reply(400, { error: 'update-self-invalid-phone' });
+
+      try {
+        await store.dispatch(actions.editSelf(user));
+      } catch (_) {} // eslint-disable-line no-empty
```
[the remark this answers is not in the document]
mgwalker: Since it's testing an error condition and the action returns a promise, it'll reject, so we want to catch that and then check the expectations.  We could do that inside the `catch` but that feels weird to me.

The `_` is commonly used in some languages to indicate a variable that you're not going to use.  Like if you have a function that takes two arguments but you only care about the first, you might declare it as `function (_, arg) { ... }`.  eslint also won't complain if you have an unused variable named `_`.  Here, Javascript requires you to catch *something* but since we're ignoring it, I made it `_`, which satisfies basically everything.  🙂

web/src/components/PasswordWithMeter.js
```diff
@@ -0,0 +1,105 @@
+import { Choice, TextField } from '@cmsgov/design-system-core';
```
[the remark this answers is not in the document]
mgwalker: Webpack 4 does tree-shaking by default!  🎉

web/src/components/PasswordWithMeter.js
```diff
@@ -0,0 +1,105 @@
...
+      passwordQuality = <span>&nbsp;</span>;
+    } else if (strength === 3) {
+      passwordQuality = 'Password strength: Good';
+    } else if (strength > 3) {
+      passwordQuality = 'Password strength: Great';
+    }
+
+    const strengthClass = [
+      'strength-meter',
+      value.length > 0 ? 'visible' : 'hidden'
```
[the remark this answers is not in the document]
mgwalker: Oh, good catch, that needs to be cleaned up.  The meter originally disappeared while the password field was empty, but Jen pointed out that was a jarring user experience, to have stuff appear and move content down just because you typed something.

web/src/components/PasswordWithMeter.js
mgwalker: I think it was originally a `p` but I switched it to a `div` because of spacing issues.  Definitely something we can look at next week because I'm curious if we have a convenient way of saying "this is a `p` but it shouldn't have vertical margins".

web/src/containers/admin/MyAccount.js
mgwalker: Is that because the max width on the inputs are also based on columns?  Or at least based on column widths, anyway.  That's what drove this `maxWidth`, but I love the direction this is going.

web/src/containers/admin/MyAccount.js
mgwalker: I think you've solved it with the 8/6 columns above!

web/src/styles/scss/password-meter.scss
mgwalker: Oooooooh.

web/src/styles/scss/password-meter.scss
mgwalker: Me too, I copied it from the internet.  I need to add a comment here citing where it came from.

web/src/containers/admin/MyAccount.js
mgwalker: Nice!  Is this something we should ping James on?  (I'm also not sure what "position" means, it was just in the data model so I made a field for it.  Maybe we don't need it?  @jeromeleecms @NAretakis?)

web/src/components/PasswordWithMeter.js
mgwalker: Ah, great point.  I hadn't even thought about that.  Can we plan on pairing on this one too?  I'm not sure what'd make a proper fix.

web/src/containers/admin/MyAccount.js
jeromeleecms: @mgwalker In some of our earlier discussions, I thought the position field was meant to be pulled into this...

image

If it doesn't get used, we can cut it from layer 1 and bring it back if we find a need for it.

web/src/components/PasswordWithMeter.js
mgwalker: Added `role="alert" aria-live="polite"`

web/src/containers/admin/MyAccount.js
mgwalker: I ***think*** we got this one fixed?

web/src/styles/scss/password-meter.scss
mgwalker: Added a comment with the Github repo this came from, for future reference

web/src/styles/scss/password-meter.scss
mgwalker: Added a `no-clearfix` class to the `TextField` and made it essentially undo what the `TextField`'s clearfix does.  Then changed the margin to 0, and all is well!

web/src/containers/admin/MyAccount.js
mgwalker: Added `ariaLabel` props.  Still not sure what text should be for the position/role input, but I renamed the visual label from "position" to "role".

web/src/containers/admin/MyAccount.js
mgwalker: Put in a fieldset and an sr-only legend

web/src/containers/admin/MyAccount.js
hbillings: The inputs should be inheriting a max width from the design system!

web/src/components/PasswordWithMeter.js
hbillings: Yes! Or I can noodle it a bit and submit a suggested fix?

mgwalker:
What do we need to do to get this PR pushed through?  Is https://github.com/18F/cms-hitech-apd/pull/1357#discussion_r266181307 the only thing holding us up?

hbillings:
@mgwalker I think so. I'm inclined to say let's punt on it. (We could also hide the checkbox from screen readers, because if you're using a screen reader you probably can't see your password, so the option to show or hide it seems irrelevant?)

PR #1365, opened 2019-03-08
Build and return APD names

### LINKED ISSUE #1323: When an APD is returned from the API, include its creation timestamp
Step 1 of #1143 

The creation timestamp is already captured in the database in the `created_at` column, we just need to grab that and send it back along with everything else.

### This task is done when...
 - [ ] API calls to `GET /apds` and `GET /apds/:id` include the APD creation timestamp
   - in both so lists of APDs can also show the "formal" identifier
- [ ] OpenAPI documentation is updated

mgwalker:
APD names are built from the `created_at` and `state_id` fields, and become:

```
[state_id uppercase]-[created_at as YYYY-MM-DD]-HITECH-APD
```

The last part of the name - the document type - is currently coded as `APD` since that's all eAPD will be supporting for layer 1.

Closes #1323

- adds a synthetic `name` property to the data model for APD names ("synthetic" meaning it doesn't actually exist as a property in the data model, but it's built from other parts of the data model)
- returns the `name` property on `GET /apds`, `GET /apds/:id`, and `POST /apds` API requests

- [ ] Product has approved the experience - mainly just need verification that the naming scheme is right


api/seeds/test/apds.js
```diff
@@ -7,16 +7,24 @@ exports.seed = async knex => {
   await knex('apds').insert([
     {
       id: 4000,
+      // William Howard Taft becomes the only person to serve as both
+      // President and Chief Justice of the Supreme Court
```
hbillings: These are delightful.

Issue #1330, opened 2019-02-27
Implement CMS Design System in Key Personnel child section


jenniferthibault:
_Broken up for implementation from #1269_
____
In order to build out the transition to the CMS design system, add the pieces needed to implement the system’s new components and styling for the Key Personnel and Project Management Planning section: https://hitech-apd.app.cloud.gov/apd#apd-state-profile-key-personnel

### This task is done when...
- [x] The Primary APD point of contact is available by default when the section loads  
  - [x] The primary APD point of contact cannot be removed from the list
- [x] The [Review component](https://design.cms.gov/patterns/review/) has been added and used for each contact
- [x] The “costs attributable” question uses the [Choice component](https://design.cms.gov/components/choice/) for checkboxes with children
   - [x] The component has been added to the codebase
   - [x] “No” option has been selected by default
   - [ ] A user can “Cancel” adding costs for an individual, and the choice will select “No”
 - [ ] Overall layout and interaction matches the [InVision prototype](https://gsa.invisionapp.com/share/BJQNX2KXQVW#/348460956_State_Profile_Contact)
- [x] Users can review a list of the contacts they created, edit an existing contact they created, and remove any contact who isn’t the Primary APD point of contact

PR #1375, opened 2019-03-13
Make individual buttons screen-only as appropriate

### LINKED ISSUE #1348: Remove buttons in print view
For print media, buttons should be hidden/removed because paper is not interactive.

From #1160

### This task is done when...
 - [ ] in print mode, there aren't any buttons

mgwalker:
- adds the `visibility--screen` CSS class to buttons that should be hidden in print view
- closes #1348


web/src/components/CollapsibleList.js
```diff
@@ -44,6 +44,17 @@ class ListItem extends Component {
...
+          <button
+            type="button"
+            className="btn btn-no-focus p1 flex-auto left-align bg-blue-light rounded"
+            onClick={this.toggle}
+          >
+            <header className="flex items-center justify-between">
+              {header}
+            </header>
+          </button>
+        </div>
```
mgwalker: This is for the little `x` button on the far right side of the mini-collapsible list headers, which will delete an item:

Screen Shot 2019-03-13 at 10 29 29 AM

Whether to display the `x` is conditioned on whether or not there's a delete handler - no handler, no `x` button.  Because the `x` button provides the rounding on the right side, just hiding the `x` button messes up the rounding.  Instead of doing lots of extra logic work here, I just split this into screen and print versions, where the print version doesn't do any of the logic because the `x` is always omitted.

web/src/components/DeleteConfirm.js
```diff
@@ -57,7 +57,7 @@ class DeleteButton extends Component {
       <div>
         <Btn
           size="small"
-          extraCss="bg-black white h6"
+          extraCss="bg-black white h6 visibility--screen"
```
mgwalker: Not really wild about this `extraCss` stuff here, but when we switch everything over to the CMS design system buttons, at least this can serve as documentation of which buttons should be hidden in print?

web/src/components/CollapsibleList.js
hbillings: Wild idea: what if we blew away the button altogether and just displayed the text??

web/src/components/DeleteConfirm.js
hbillings: 👍

web/src/components/CollapsibleList.js
mgwalker: You mean in print?  That...  makes too much sense.  I was hesitant to change too much other than just making the clickable buttons go away, but replacing the button with just a header with text in it makes sense.  Happy to do that if you think it's a better approach!

Alternatively, these headers are useless in print because they get truncated (as in the screenshot) and we need a totally different view for them in print?  I think that would be a separate PR, but maybe worth asking @jenniferthibault what she thinks while we still have the chance?

jenniferthibault: I think that when the designs from https://github.com/18F/cms-hitech-apd/issues/1330 are implemented, we won't have this problem because we'll be using the Review component, which neither truncates text nor has little x buttons. So do whatever you gotta do to get there :)

Issue #1145, opened 2018-11-19
Update content inconsistencies with terminology and refine some language (from round 8) 


lauraponce:
### Background

From the [8th round of testing](https://docs.google.com/document/d/1F8deUxl2zuNiXoKHv05cu4o3oueqP7Te9girOcsyEw8/edit), state users brought up some inconsistencies with the language and pointed out some areas of confusion.

### This task is done when...
 - [ ] When should we use the term "project" vs "program"? The app seems to use them interchangeably in some places. 
 - [ ] In the In-House Cost Categories section, there is a Salary Amount box that counts benefits as part of the salary. This wasn't clear to users because there is nothing telling them that. How can this be clarified?
- [ ] People used the phrases "save as" and "print out" when talking about turning the APD into an exportable doc. What is the clearest way to refer to saving and printing so that it makes sense to the most people?

jameshupp:
I'm just gonna leave comments here about each of the checkboxes.

1. The project vs program distinction really comes down to one section, `Key Personnel...`. That section uses `project` repeatedly, but it's clearly about people involved in activities covered by the APD. Those activities are _Program Activities_ so I think we can switch to `program` without any trouble. The one exception: there is a definition of `key personnel` that reads as somewhat technical. Is this first sentence set in stone, or can we change `project` to `program` here @NAretakis?

> Key personnel are individuals involved with all activities within this APD, or are involved with governance, coordination, stakeholder decision-making, oversight, and/or **project** performance. 

2. Two solutions for how to clarify that salary includes benefits. Either call it `total compensation` or write a helpText note to specify that benefits are included in salary. The first option is more direct so it's my preference, but it could still leave people confused on whether things like bonuses and leave time are included (as opposed to health care and retirement contributions, which certainly are). Better to try one of these options rather than nothing, but open to input about which. Also looking for input from Nick/Jerome here, if they know the answer.

3. Currently says `Save as PDF`. This feels right to me. That it is printable will be immediately apparent when they click the button, but I'd rather not make anyone feel like they're committing to that act without seeing how it's going to render. This does raise a question for me: does the app force the print dialogue to select PDF as the target for the print operation? Or is that just because my computer is not connected to a printer and that's my default? @mgwalker

mgwalker:
3 - Because you don't have a printer.  We can only make the dialog pop up, we can't control anything else about it.

jeromeleecms:
[...]

PR #1381, opened 2019-03-15
corrects minor content inconsistencies

jameshupp:
Addresses #1145.

- clarifies what salary amount includes (by, uh, rewording it entirely) and that it's per FTE
- redresses conflicting use of `program` and `project`.


jeromeleecms:
@jameshupp The content updates looks fine. From a readability standpoint as I was reading the help text, is there anything we can do that calls out that "Cost with Benefits" and "Number of FTE's" refer to specific fields in the section below? 

If you think this is outside the scope of the PR, I can go ahead and approve - this seems(?) like this question may have larger implications on all the help text within the app.

jameshupp:
@jeromeleecms I just moved a lot of that content out of the helpText into the instruction. I think it makes more sense there anyway. The helpText is now just the example. It's a little long this way, but I think it makes more sense and is clearer.

Issue #1378, opened 2019-03-14
Implement new styles and layout for Layer 1 state dashboard/landing page


jenniferthibault:
Carried forward from design in  #1268.
#1346 and #1366 will continue to build on this, but should not block it. 

### This task is done when...
 - [ ] The code implemented matches the design mockups, using the CMS Design System and its components
 
Desktop screenshot:
![Layer 1: Dashboard - design system Copy](https://user-images.githubusercontent.com/11636908/54382689-dffcb580-4666-11e9-8a0f-0991db9fadb8.png)

jenniferthibault:
**Implementation styling notes:**

- `[State name APD builder]` uses H1 visual styling
- `[State name APDs]` uses H2 visual styling
- `To submit an APD` uses H6 visual styling (matching the [list component styling](https://design.cms.gov/components/list/) in the design system)
- `[State name APDs]` or APD name list items uses H4 visual styling + link styling
- The list of APD's is styled like the [Review pattern](https://design.cms.gov/patterns/review/)
- This uses an 8 columns centered on the grid (or 8 columns off-set by 2 on each side)
- Vertical spacing is still eyeballed — as we've discussed previously, use the defaults and let's see what it looks like before adjusting
![Uploading Screen Shot 2019-03-14 at 2.42.38 PM.png…]()

Issue #1351, opened 2019-03-05
Design admin landing page


mgwalker:
There are currently admin pages for creating and editing user accounts, but the only way to get to them is to navigate to them by URL.  My thinking was this landing page might just be links to those two pages, but maybe those two items become widgets on a combined admin page?

### This task is done when...
 - [x] we know what a single admin point-of-entry (?) looks like and how it behaves
 - [x] handed off to dev (see #1273)

jenniferthibault:
One place we could start that tried to keep things simple:

![Layer 1: Admin Landing_2](https://user-images.githubusercontent.com/11636908/54328368-0594ab00-45e4-11e9-9384-d45d47ca95be.png)




Implementation considerations:
- This uses the same grid foundation as the state dashboard page from #1268 
- Each card is four grid columns wide with one gutter in-between
- The icons are [user-cog](https://fontawesome.com/icons/user-cog?style=solid) and [user-plus](https://fontawesome.com/icons/user-plus?style=solid) from Font Awesome under a Creative Commons Attribution 4.0 International license. That means we need to abide by the following:
> You must give appropriate credit, provide a link to the [license](https://fontawesome.com/license), and indicate if changes were made. You may do so in any reasonable manner, but not in any way that suggests the licensor endorses you or your use.

In the past, I've handled this by listing the icons we use and providing the attribution and license link in the repo's LICENSE file. If we go for it here, I'd make the PR for that from this issue separate from the dev issue. (And I could do it!)

I did consider adding both widgets to this landing page directly, as Greg suggested above, but that quickly lost the clarity in exactly what tasks were possible at this stage, which the above mockup provides. I also considered making larger solid background card-type components for this (see below), but those would be totally custom elements not from the design system. Since the mockup above uses just typography styles, it allows us to avoid customization and code used in a one-off way.
![Screen Shot 2019-03-13 at 11 01 06 PM](https://user-images.githubusercontent.com/11636908/54328339-e9910980-45e3-11e9-820e-18815810ec0a.png)

NAretakis:
From my perspective, this look great. I like the typography styles version as it keeps things simple but still accomplishes what this page needs to do.

jenniferthibault:
Moved to implementation in #1273

Issue #1035, opened 2018-09-21
Add visual styling documentation to wiki


jeromeleecms:
So that future folks understand how we've made decisions so far, we should include information about what/how we're working.

This task will be complete when the wiki includes: 
-[ ] Tell people we use the CMS Design system, and where to find it
-[ ] Point out ways in which we've differed or customized (list examples of custom components)
-[ ] Point to the sketch file and how to find the latest version

jenniferthibault:
@jeromeleecms since we're now using the CMS design system, I'm thinking we don't need this issue exactly as it was initially intended since there's no separate style guide. What it might make sense to do instead is something like:

- Tell people we use the CMS Design system
- Point out ways in which we've differed or customized (list examples of custom components)
- Point to where Sketch files live on Google Drive

At a later point, the team may want to decide if it's worth creating a more full visual library of patterns or styles for this site, but if feels like our implentation of the CMS Design system is still kinda early to tell.

jeromeleecms:
I'm good with this as long as it has no impact on ATO / satisfies any ATO documentation requirements (@mgwalker or @NAretakis?).

Question/Clarification; If we point to Google Drive, will be the general public be able to access it? If they can't, can we post the latest version on GitHub?

mgwalker:
AFAIK, the ATO process doesn't take style guides or documentation into account.  Generally speaking, ATO is more interested in "what" over "how"; e.g., it will want to see that we satisfy the requirements of section 508 (accessibility), but will largely not care about _how_ we do that.

jeromeleecms:
Excellent! As long as the sketch file is accessible by visitors, let's close this out.

jenniferthibault:
Oh the sketch file — I'd highly recommend prioritizing the team's access to the Sketch file over external visitors. The Sketch file changes daily, so if we upload it to GitHub, that's probably not a reasonable expectation for keeping the most updated version available. The same is true of uploading it to Google Drive, except it's not disappointing anyone from the public. 

If we're just uploading it as a backup for a working resource, and not for the SOURCE OF TRUTH:tm: (which should always be the code) we could do either, as long as we're setting clear expectations in the framing content about who on the team is responsible for the most up-too-date files.

jenniferthibault:
And if that feels good, then there's still work to do do complete this issue! So plz don't close

jeromeleecms:
Sounds good - I'll defer to you @jenniferthibault about when you are ready to close!

[...]

PR #1382, opened 2019-03-15
Add icon license info

jenniferthibault:
Adds required license and attribution info to the repository for the icons that are part of #1351 and #1378 to implement the new admin & state user dashboards. I'll also include instructions for this in the documentation for #1035 

I modeled this after what we used for the [18F brand site](https://github.com/18F/brand/blob/master/LICENSE.md)


mgwalker:
@jenniferthibault We're using some other FA icons as well.  Would you mind if I added them to this list rather than create a new PR?

jenniferthibault:
Go for it @mgwalker !

PR #1386, opened 2019-03-21
Un-collapse subsection accordions

### LINKED ISSUE #1344: Modify <Subsection> component to use CMS design system styles with new layout
_Continuation of #1269_
____
To start implementing the style and layout of the new design system, update our `<Section>` component

### This task is done when...
 - [x] The `Subsection` component is updated to stop providing an accordion
  - [x] The styling matches the mockups below:
	- Subsection headline is semantically and stylistically h3: `ds-h3`
	- Accordion functionality has been removed
 - [x] Overall layout and interaction matches the [InVision prototype](https://gsa.invisionapp.com/share/BJQNX2KXQVW#/348460956_State_Profile_Contact)
screen shot 2019-02-27 at 12 01 24 am

hbillings:
Closes #1344. No more accordions in sections, yay! I believe the only place the `Collapsible` element is used now is in the Activity section, to collapse activities down to a readable list. 

There is a little weirdness in those Activity sections: the `Collapsible` component passes its children to `Subsection` for reasons I don't entirely understand (I feel like it probably doesn't actually need to, but I'd like to look at this with @mgwalker). Because of this, the headings in those subsections need to be `h4` instead of `h3`. There's a little bit of repetitive code to make this work (by checking whether the subsection has a `nested` property.

Screen Shot 2019-03-21 at 6 33 50 PM

- Removes the `Collapsible` component from `Subsection`.
- Applies the correct heading to `Subsection` titles (including taking into account whether they are nested subsections).

### DIFF
--- a/web/src/components/Collapsible.js
+++ b/web/src/components/Collapsible.js
@@ -33,7 +33,7 @@ class Collapsible extends Component {
     const replaceSpaces = / /g;
     const removeColon = /:/g;
     const removeCarat = /›/g;
-    const slug = `${title.replace(replaceSpaces, '-').replace(removeColon, '').replace(removeCarat, '').toLowerCase()}`;
+    const slug = title.replace(replaceSpaces, '-').replace(removeColon, '').replace(removeCarat, '').toLowerCase();
     const contentId = `collapsible-${slug}`;
 
     let btnClass = deline`
--- a/web/src/components/Collapsible.js
+++ b/web/src/components/Collapsible.js
@@ -1,5 +1,4 @@
 import deline from 'deline';
-import kebabCase from 'lodash.kebabcase';
 import PropTypes from 'prop-types';
 import React, { Component } from 'react';
 
@@ -31,7 +30,11 @@ class Collapsible extends Component {
     const { open: openState } = this.state;
 
     const isOpen = onChange ? openProp : openState;
-    const contentId = `collapsible-${kebabCase(title)}`;
+    const replaceSpaces = / /g;
+    const removeColon = /:/g;
+    const removeCarat = /›/g;
+    const slug = `${title.replace(replaceSpaces, '-').replace(removeColon, '').replace(removeCarat, '').toLowerCase()}`;
+    const contentId = `collapsible-${slug}`;
 
     let btnClass = deline`
       btn btn-collapse block col-12 py2 sm-px3 h3 sm-h2 line-height-1 left-align
--- a/web/src/components/Collapsible.js
+++ b/web/src/components/Collapsible.js
@@ -50,7 +50,7 @@ class Collapsible extends Component {
     `;
 
     return (
-      <div id={id} className={`mb3 bg-${bgColor} rounded shadow accordian`}>
+      <div id={id} className={`mt3 mb3 bg-${bgColor} rounded shadow accordian`}>
         <button
           type="button"
           className={btnClass}
--- a/web/src/components/Section.js
+++ b/web/src/components/Section.js
@@ -1,7 +1,6 @@
 import PropTypes from 'prop-types';
-import React from 'react';
+import React, { Fragment } from 'react';
 
-import Collapsible from './Collapsible';
 import Instruction from './Instruction';
 import { t } from '../i18n';
 
@@ -52,10 +51,16 @@ const Subsection = ({ children, id, nested, open, resource }) => {
   const title = t([resource, 'title'], { defaultValue: '' });
 
   return (
-    <Collapsible id={id} title={title} open={open} nested={nested}>
+    <Fragment>
+      <h3
+        id={id}
+        className='subsection--title'
+      >
+        {title}
+      </h3>
       <Instruction source={`${resource}.instruction`} />
       {children}
-    </Collapsible>
+    </Fragment>
   );
 };
 
--- a/web/src/components/Section.js
+++ b/web/src/components/Section.js
@@ -47,17 +47,27 @@ Section.defaultProps = {
   isNumbered: false
 };
 
-const Subsection = ({ children, id, nested, open, resource }) => {
+const Subsection = ({ children, id, nested, resource }) => {
   const title = t([resource, 'title'], { defaultValue: '' });
 
   return (
     <Fragment>
-      <h3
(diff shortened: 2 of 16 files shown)


web/src/components/Section.js
```diff
@@ -48,30 +47,44 @@ Section.defaultProps = {
...
+          {title}
+        </h3>
+      }
+      {nested &&
+        <h4
+          id={id}
+          className='ds-h3'
+        >
+          {title}
+        </h4>
```
hbillings: I'm not crazy about this way of doing this -- as far as I can tell, the activity details are the only thing that need the `h4` version. I'd rather just have them not call the subsection, since this seems to just be rendering the children now with very little other benefit.

web/src/styles/scss/section.scss
```diff
@@ -28,8 +28,13 @@ body {
     }
   }
 
-  margin-top: $spacer-2;
+  margin-top: $spacer-6;
```
hbillings: I added a little extra padding to the regular `Section` component here, because it felt like it needed more breathing space once the subsection was uncollapsed.

web/src/components/Collapsible.js
```diff
@@ -1,5 +1,4 @@
 import deline from 'deline';
-import kebabCase from 'lodash.kebabcase';
 import PropTypes from 'prop-types';
```
hbillings: While I was tracing the route of the Activity components, I ran across this. All it seems to do is replace spaces/special chars with dashes and lowercase things, both of which are easy to do with regex in vanilla JS. Seemed lighter weight than importing something that's only being used once in the file. 🤷‍♀️

web/src/components/Collapsible.js
```diff
@@ -1,5 +1,4 @@
 import deline from 'deline';
-import kebabCase from 'lodash.kebabcase';
```
hbillings: While I was tracing the route of the Activity components, I ran across this. All it seems to do is replace spaces/special chars with dashes and lowercase things, both of which are easy to do with regex in vanilla JS. Seemed lighter weight than importing something that's only being used once in the file (especially since it only needs to care about two special chars. 🤷‍♀️

web/src/components/Collapsible.js
```diff
@@ -31,7 +30,11 @@ class Collapsible extends Component {
     const { open: openState } = this.state;
 
     const isOpen = onChange ? openProp : openState;
-    const contentId = `collapsible-${kebabCase(title)}`;
+    const replaceSpaces = / /g;
+    const removeColon = /:/g;
+    const removeCarat = /›/g;
```
hbillings: I could group these together, but I sort of liked naming them explicitly because regexes are awful.

web/src/components/Section.js
hbillings: 👍 This sounds like something to address when we start tackling converting individual sections to the design system.

hbillings:
@mgwalker  So weird; I don't know why either of those visual regressions happened! They should be fixed now, though. Thank you!

lauraponce:
Should the bot have created a new link to view the changes after the most recent updates, @hbillings? When I use the initial link (http://pr1386.hitech.eapd.cms.gov.s3-website-us-east-1.amazonaws.com) and clear my cache, I'm still seeing the version with the issues in the Activity section; so I guess I'm not sure where to go to see the latest?

mgwalker:
@lauraponce It updates its previous comment rather than create a new one, so the link stays the same. It adds a commit ID when it updates, so we can track which commit it has deployed.  That said, I also still see a blue background in the activity section.  The rich text boxes are white, though! 😄

hbillings:
So I _can_ put the white box back, but it will mean adding back extra padding also. I'm wondering if maybe we could just get rid of the blue background altogether:
Screen Shot 2019-03-25 at 11 32 04 AM

For reference, here's how the current version goes:
Screen Shot 2019-03-25 at 11 32 39 AM

The current version with the white and blue background makes more sense to me when those sections are accordions. When they're not, it feels a little extraneous? Or maybe it's just me:
Screen Shot 2019-03-25 at 11 37 31 AM

jameshupp:
In the accordion format, activities are at a different level of hierarchy than the other accordions. Because they were within a different accordion? My question is: without accordions, are they still at a different level of hierarchy? If so, is there something that visually distinguishes them? If they're not at a different level _or_ if they still visually play as different, I agree the background color is extraneous. If it's a different level we need some way to make sure that's clear.

hbillings:
@jameshupp These are great questions! It seems like now that the accordions are gone, these fields are all just subheadings, so I'm not sure if there's value in creating further subdivision between them.

lauraponce:
Great points, @jameshupp. You're right that the blue was just added in as a visual hierarchy as a way to connect all the subsections within a single activity to show what's part of one activity and to minimize confusion when entering multiple activities.

In my opinion, the blue can go away. And hopefully the organization of activities and their corresponding activity subsections will be clear by the default design system headings and font sizes?

NAretakis:
So, its still a little hard to navigate an activity without the sticky sidebar telling me where I am inside of that activity. The sticky header helps, telling me if I am inside of Activity 1, 2, or 3, however where I am inside of Activity 1 is definitely tricky without the sidebar nav.

That said, to @jameshupp 's point, each subsection within an activity was really meant to just give some sense of navigation within an activity (because an activity can be very large on the page). The ability to use the accordion within the activity (using the accordion section headers) was just a way to navigate, and declutter the screen as one filled out each subsection for an activity.

- Activity 1
  - Activity 1 Overview
  - Activity 1 Goals
  - Activity 1 Schedule
  - Activity 1....
- Activity 2
  - Activity 2 Overview
  - Activity 2 Goals
  - Activity 2 Schedule
  - Activity 2....

NAretakis:
Also if we removed the set heights on all of the fields to reduce white space, I think that would be a huge win. They grow in size as needed, but look to have a default height that looks very odd when there is just a little bit of text. Might not be related this PR though..

image

jameshupp:
We probably need a check at some point later on all the fields' default sizing. They're a hint about what we're expecting, right? If we can even distinguish "short" and "longish" we'll be doing people a big favor.

Issue #1337, opened 2019-02-28
Re-stickify sidebar


hbillings:
The sidenav lost its sticky with the redesign, partly because it seems like the nav no longer fits on some screens. We should re-stick it and address the design concerns as needed.

### This task is done when...
 - [x] We have a decision on how the nav should look and function. (For example, we've adjusted the padding to make things fit on a single screen, or we've made the nav wide enough to accommodate longer text on a single line.)
 - [x] The sidebar sticks.
 - [x] The stickiness works across all browsers.

jenniferthibault:
@jameshupp before I change the scaffolding, is there anything we could do to shorten the parent and child-level titles through better content design? I'm not sure how recently this was visited.

jameshupp:
I could use some constraints.

* How much shorter is useful? All on one line, fewer lines wherever they may come from, just shorter please who cares about lines, depends on which level of title it is how much shorter it should be?
* What is the goal or direction you're trying to go in here? Or if this could go multiple directions, what are the options, even loosely?

There probably are some changes we could make, but I want to get a sense there to see if what feels realistic in a short-ish timeframe is going to be helpful.

jenniferthibault:
FAIR! Yes good point.

So. I don't have the rules, but I can tell you what's driving this need and maybe that helps us get to the rules together.

The primary in-page navigation is that side nav. We want it to be sticky so that it travels with the user down the page. Right now, the nav is taller in height than many browser windows, so you can't actually access any of the nav items at the bottom of the list.

When any accordion panel is expanded, you can't get to nav items at the bottom of the list until other panels above it are closed.

There are so many sections that this will never be perfect, and we'll always have this problem as long as the app is one long page instead of multiple pages. But it seems like the best thing we can do to make it as short as we can is to keep section titles (as they appear in the nav) as few lines as possible. One is GREAT! Two, less great. Three: really not great.

Glad to pair up if that helps!

jenniferthibault:
Here are some additional tactics that could help:
1. Use just the state name and icon in the top left corner. Drop `HITECH APD`  to make more space.
Hypothesis: By now the user will know where they are and what they're doing.
2. We banish "Export" back into the nav as a final step (where users can get further instructions for what's next) instead of as a button available from everywhere: https://github.com/18F/cms-hitech-apd/issues/1321 
[...]

Issue #1321, opened 2019-02-26
In order to feel confident that APD creators have reached the end of the form experience, provide a closing cue


jenniferthibault:
_Carried forward from a comment in https://github.com/18F/cms-hitech-apd/pull/1320 and issue #1318_

Without the final "submit" action, removed for the pilot from issue #1318, we lose a bit of clarity that someone is "done" with the form after the Executive summary, and obvious next steps.

Through design and content, we should create the feeling of done-ness and clear steps for finishing. There are likely multiple ways to do this. 

### This task is done when...
 - [x] The UX for finishing the form is designed
 - [x] The user has clear instructions to follow to submit their APD (addressing in #1256)
 - [x] The team agrees
 - [x] Issues to implement the change have been created if needed

NAretakis:
Move buttons from #1348 to the bottom and this issue can be resolved.

jenniferthibault:
Hold on there speedy! 

Moving buttons from the side to the bottom works if users complete the form starting at the top and ending at the bottom, but I'm not confident that's the ONLY case. I thiiiiink we know that folks jump around a bit, and want to save at multiple points, not just when they're "done".

Here's another idea!
We could Replace the section formerly known as `Certify & Submit` with a more simple section called `Export` (check it out @lauraponce I think I found a scenario for the BIG BUTTON 😄 ) Since the side nav is going to be sticky, folks will be able to get to the export functionality from any other point in the page by going through the side nav. 

![App - 9: Export](https://user-images.githubusercontent.com/11636908/54065506-d8b15400-41ef-11e9-8965-072eeb938b35.png)

It needs some content design here to make sure the instructions are right for this closing stage (ping! @jameshupp )

This would provide the "end cue" as we've defined, and I'd plan on resolving the rest of the "Save" button location problem in #1337 

Thoughts?

NAretakis:
I like this a lot. Clean and consistent approach, that also works with the sticky side nav.

jameshupp:
We just heard in testing that in interstitial step or two between deciding to export and getting a PDF would be nice. We probably can't provide that right away, but this separate Export section gives us enough room to describe the situation effectively. Assigning myself so I can knock out copy for this. This looks fantastic!

jameshupp:
This pattern emerged in yesterday's conversation about implementing the design system across the sections of the app. Everyone seemed onboard with creating this and we have Nick's comment above, so I'm going to check off "team agrees" so I'm going to check this off.

[...]

PR #1395, opened 2019-03-26
Highlight navbar items as the user scrolls

mgwalker:
Adds waypoints before all `<Section>` and `<Subsection>` elements and updates the active navigation section as they come into view, where "in view" is defined as 90% from the bottom of the screen.  The 90% is kind of arbitrary, but simply "on the screen" ends up with an aggressive battle over what thing is currently selected when multiple things are on the screen.  E.g., should the nav bar change as soon as another section scrolls into view from the bottom, even though 90% of the screen is occupied by the current section?

I think 👆 that is a tough design question.  What are the conditions to set a section as the active one?

Into the dev weeds, because this generates state changes as the user scrolls, I noticed a lot of performance problems.  It turns out some of our `mapStateToProps` functions are doing data calculation and mutation, which means they get executed every time the state changes.  That was contributing to our performance problems because it meant some of the budget data was being recomputed on every keypress (😬), but then it got worse because it was also happening on scroll (😬😬😬😬😬).  This PR also adds the [reselect](https://npm.im/reselect) library to create memoized state selectors for the `mapStateToProps` functions that do calculation/mutation.

Reselect creates a function that remembers its last arguments and return value.  If the arguments don't change, then it just returns the previous return value.  By narrowing which piece of the app state the function relies on, reselect can memoize based on just that piece of state, rather than the whole state.  It can also memoize on props, so you can create a function that is only executed once until its relevant pieces of data are changed.

- highlights the navbar as the user scrolls
- memoizes mutating state-to-props mapping functions for performance gains

### DIFF
--- a/web/src/actions/navigation.js
+++ b/web/src/actions/navigation.js
@@ -1,0 +1,24 @@
+export const NAVIGATION_SCROLL_TO_WAYPOINT = Symbol(
+  'navigation : scroll to waypoint'
+);
+
+let jumping = false;
+
+export const jumpTo = waypoint => dispatch => {
+  if (waypoint) {
+    jumping = true;
+    setTimeout(() => {
+      jumping = false;
+    }, 100);
+    dispatch({ type: NAVIGATION_SCROLL_TO_WAYPOINT, data: waypoint });
+  }
+};
+
+export const scrollTo = waypoint => dispatch => {
+  if (!jumping && waypoint) {
+    dispatch({
+      type: NAVIGATION_SCROLL_TO_WAYPOINT,
+      data: waypoint
+    });
+  }
+};
--- a/web/src/components/Section.js
+++ b/web/src/components/Section.js
@@ -1,13 +1,14 @@
 import PropTypes from 'prop-types';
-import React from 'react';
+import React, { Component, Fragment } from 'react';
+import { connect } from 'react-redux';
+import { Waypoint } from 'react-waypoint';
 
 import Collapsible from './Collapsible';
 import Instruction from './Instruction';
+import { scrollTo } from '../actions/navigation';
 import { t } from '../i18n';
 
-const SectionTitle = ({ children }) => (
-  <h2>{children}</h2>
-);
+const SectionTitle = ({ children }) => <h2>{children}</h2>;
 
 SectionTitle.propTypes = {
   children: PropTypes.node.isRequired
@@ -19,60 +20,96 @@ SectionDesc.propTypes = {
   children: PropTypes.node.isRequired
 };
 
-const Section = ({ children, id, isNumbered, resource }) => {
-  const title = t([resource, 'title'], { defaultValue: false });
-  const helptext = t([resource, 'helpText'], { defaultValue: false });
+class SectionPlain extends Component {
+  hitWaypoint = id => () => {
+    const { hasOverview, scrollTo: action } = this.props;
+    action(hasOverview ? `${id}-overview` : id);
+  };
 
-  return (
-    <section
-      id={id}
-      className={isNumbered && 'numbered-section'}
-    >
-      <h2 className="ds-h2">{title}</h2>
-      <span className="ds-text--lead">{helptext}</span>
-      {children}
-    </section>
-  );
-};
+  render() {
+    const { children, id, isNumbered, resource } = this.props;
+    const title = t([resource, 'title'], { defaultValue: false });
+    const helptext = t([resource, 'helpText'], { defaultValue: false });
 
-Section.propTypes = {
+    return (
+      <section id={id} className={isNumbered && 'numbered-section'}>
+        <Waypoint onEnter={this.hitWaypoint(id)} bottomOffset="90%" />
+        <h2 className="ds-h2">{title}</h2>
+        <span className="ds-text--lead">{helptext}</span>
+        {children}
+      </section>
+    );
+  }
+}
+
+SectionPlain.propTypes = {
   children: PropTypes.node.isRequired,
+  hasOverview: PropTypes.bool,
   id: PropTypes.string,
   resource: PropTypes.string,
-  isNumbered: PropTypes.bool
+  isNumbered: PropTypes.bool,
+  scrollTo: PropTypes.func.isRequired
 };
 
-Section.defaultProps = {
+SectionPlain.defaultProps = {
+  hasOverview: true,
   id: null,
   resource: null,
   isNumbered: false
 };
 
-const Subsection = ({ children, id, nested, open, resource }) => {
-  const title = t([resource, 'title'], { defaultValue: '' });
+class SubsectionPlain extends Component {
+  hitWaypoint = id => () => {
+    const { hasWaypoint, scrollTo: action } = this.props;
+    if (hasWaypoint) {
+      action(id);
+    }
+  };
 
-  return (
-    <Collapsible id={id} title={title} open={open} nested={nested}>
-      <Instruction source={`${resource}.instruction`} />
-      {children}
-    </Collapsible>
-  );
-};
+  render() {
+    const { children, id, nested, open, resource } = this.props;
+    const title = t([resource, 'title'], { defaultValue: '' });
 
-Subsection.propTypes = {
+    return (
+      <Fragment>
+        <Waypoint onEnter={this.hitWaypoint(id)} bottomOffset="90%" />
+        <Collapsible id={id} title={title} open={open} nested={nested}>
+          <Instruction source={`${resource}.instruction`} />
+          {children}
+        </Collapsible>
+      </Fragment>
+    );
+  }
+}
+
+SubsectionPlain.propTypes = {
   children: PropTypes.node.isRequired,
+  hasWaypoint: PropTypes.bool,
   id: PropTypes.string,
   nested: PropTypes.bool,
   open: PropTypes.bool,
-  resource: PropTypes.string
+  resource: PropTypes.string,
+  scrollTo: PropTypes.func.isRequired
 };
 
-Subsection.defaultProps = {
+SubsectionPlain.defaultProps = {
   resource: null,
+  hasWaypoint: true,
   id: null,
   nested: false,
   open: false
 };
 
+const Section = connect(
+  null,
+  { scrollTo }
+)(SectionPlain);
+
+const Subsection = connect(
+  null,
+  { scrollTo }
+)(SubsectionPlain);
+
 export default Section;
+
 export { Section, SectionDesc, SectionTitle, Subsection };
--- a/web/src/components/Section.js
+++ b/web/src/components/Section.js
@@ -1,10 +1,7 @@
 import PropTypes from 'prop-types';
-import React, { Component, Fragment } from 'react';
-import { connect } from 'react-redux';
-import { Waypoint } from 'react-waypoint';
+import React, { Fragment } from 'react';
 
 import Instruction from './Instruction';
-import { scrollTo } from '../actions/navigation';
 import { t } from '../i18n';
 
 const SectionTitle = ({ children }) => <h2>{children}</h2>;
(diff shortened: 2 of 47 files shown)


web/src/components/Section.js
```diff
@@ -4,9 +4,7 @@ import React, { Fragment } from 'react';
 import Instruction from './Instruction';
```
mgwalker: I didn't actually change this file, in the end - these changes are due to `prettier`

I'd initially put the waypoints into the `Section` and `Subsection` components thinking we'd want them to happen automatically, but there was *just enough* variety in when and whether we wanted waypoints that it became unwieldy.  Rather than beat it into submission, I refactored the waypoints into their own component.

web/src/containers/ApdPreviousActivityTableTotal.js
```diff
@@ -76,36 +77,9 @@ ApdPreviousActivityTableMMIS.propTypes = {
...
-    }),
-    {}
-  );
-
-  return {
-    totals
-  };
-};
+const mapStateToProps = state => ({
+  totals: selectPreviousActivityExpensesTotals(state)
```
mgwalker: Memoized selector.  It does all the math and object-building that `mapStateToProps` used to do, but only when the previous activity expense data changes rather than on every single state change.

web/src/containers/ConnectedWaypoint.js
```diff
@@ -0,0 +1,42 @@
...
+    const { scrollTo: action } = this.props;
+    action(id);
+  };
+
+  render() {
+    const { children, id } = this.props;
+    return (
+      <Fragment>
+        <Waypoint onEnter={this.hitWaypoint(id)} bottomOffset="90%" />
+        {children}
```
mgwalker: The `ConnectedWaypoint` will work identically whether it has children or not.  I made it work both ways to accommodate the case where we were returning a `<Section>` and needed a waypoint there - this way, we can wrap the `<Section>` with the `<ConnectedWaypoint>` and still just return one component - no need to introduce a `<Fragment>`.  But in places where we're already returning multiple things, we don't have to wrap anything.

web/src/containers/Sidebar.js
```diff
@@ -162,14 +166,14 @@ class Sidebar extends Component {
...
-            id: 'exec-summary-overview',
+            id: 'executive-summary-overview',
             url: '#executive-summary',
             label: 'Overview',
             onClick: (evt, id) => this.handleSelectClick(id)
           },
           {
-            id: 'executive-summary-overview',
-            url: '#executive-summary-overview',
+            id: 'executive-summary-summary',
```
mgwalker: Summary! SUMMARY! **SUMMARY!!**

web/src/containers/activity/CostAllocateFFP.js
```diff
@@ -95,30 +96,10 @@ CostAllocateFFP.propTypes = {
...
-        ffpSelectVal,
-        allocations: { federal: costs.federal, state: costs.state }
-      };
-    });
-
-  return { byYearData, costAllocation };
+export const makeMapStateToProps = () => {
+  const selectCostAllocateFFP = makeSelectCostAllocateFFP();
+  const mapStateToProps = (state, props) => selectCostAllocateFFP(state, props);
+  return mapStateToProps;
```
mgwalker: reselect memoized functions have a cache depth of 1 - it only remembers the last arguments and return value.  If the new arguments match, it returns the old return value; otherwise, it recomputes.  This is great!  Until you need distinct state for several components simultaneously, based on props.  In that case, every call to the memoized function has unique arguments and you lose the caching behavior entirely.

To work around this, the first argument to `connect` can return a function, which allows each component to have its own `mapStateToProps` function, and thus its own version of the memoized selector. 🎉

(My rule of thumb is if your selector relies on props, you should probably use this pattern.)

web/src/containers/activity/CostAllocateFFPQuarterly.js
```diff
@@ -121,7 +116,7 @@ class CostAllocateFFPQuarterly extends Component {
                             headers={`act_qbudget_fy${year} act_qbudget_fy${year}_q${q}`}
                           >
                             <Dollars>
-                              {quarterlyFFP[year].subtotal[name].dollars}
+                              {quarterlyFFP[year][q][name].dollars}
```
mgwalker: A few small display bugs in here.  The cost allocation quarterly table was showing the computed subtotal everywhere, rather than the individual quarterly amounts.

web/src/containers/activity/Description.js
```diff
@@ -63,8 +64,8 @@ Description.propTypes = {
   updateActivity: PropTypes.func.isRequired
 };
 
-const mapStateToProps = ({ activities: { byKey } }, { aKey }) => ({
-  activity: byKey[aKey]
+const mapStateToProps = (state, props) => ({
+  activity: selectActivityByKey(state, props)
```
mgwalker: **Minor tech debt:** Eventually it'd be nice to have selectors be responsible for extracting individual pieces of data from the state, like this one does.  That way the components don't have to know about the shape of the state data.

web/src/reducers/activities.selectors.js
```diff
@@ -0,0 +1,76 @@
+import { createSelector } from 'reselect';
+import { selectApdData } from './apd.selectors';
+
+export const selectActivityKeys = ({ activities: { allKeys } }) => allKeys;
```
mgwalker: First argument is state (destructured), and second is props (also destructured!).

web/src/reducers/activities.selectors.js
```diff
@@ -0,0 +1,76 @@
...
+
+export const selectActivityByKey = ({ activities: { byKey } }, { aKey }) =>
+  byKey[aKey];
+
+const selectBudgetForActivity = ({ budget }, { aKey }) =>
+  budget.activities[aKey];
+
+export const makeSelectCostAllocateFFP = () =>
+  createSelector(
+    [selectActivityByKey, selectBudgetForActivity],
```
mgwalker: `createSelector` does the magic of creating memoized functions.  The first argument is an array of selectors (it passes state and props into them), and the results are used as parameters into the second argument.  If the parameters are the same between two calls, it just returns the cached result rather than recomputing.

The second argument receives the results of the selectors in the first argument, in the order they're listed.  As far as I can tell, it doesn't have direct access to state or props itself, so it has to get everything from those other selectors.

My rule of thumb on these is if there's no data mutation or computation, use a regular selector (like the first few in this file).  If it ***does*** mutate or compute data, use a memoized selector with `createSelector`.

web/src/reducers/activities.selectors.js
```diff
@@ -0,0 +1,76 @@
...
+          return {
+            year,
+            total: costs.total,
+            medicaidShare: costs.medicaidShare,
+            ffpSelectVal,
+            allocations: { federal: costs.federal, state: costs.state }
+          };
+        });
+
+      return { byYearData, costAllocation };
```
mgwalker: The actual body of this selector came directly from a `mapStateToProps` function elsewhere.  Yay for not having to figure all this out again.  Also, the `mapStateToProps` tests on those components incidentally tests these selectors, so 👏 for that too.

web/src/reducers/navigation.js
```diff
@@ -0,0 +1,17 @@
...
+  switch (action.type) {
+    case NAVIGATION_SCROLL_TO_WAYPOINT:
+      return { ...state, activeSection: action.data };
+    default:
+      return state;
+  }
+};
+
+export const selectActiveSection = ({ navigation: { activeSection } }) =>
+  activeSection;
```
mgwalker: Thinking about whether reducers and selectors should be in the same files or separate files. I don't have strong feelings about it...

web/src/containers/IncentivePayments.js
```diff
@@ -31,69 +36,65 @@ class IncentivePayments extends Component {
...
+                  return (
+                    <tr key={id}>
                       <td
-                        className="bold mono right-align align-middle"
-                        headers={tdHdrs(year, 'subtotal')}
+                        className={`align-middle ${i % 2 === 0 ? 'bold' : ''}`}
+                        headers="incentive-payments-table-fynull1 incentive-payments-table-fynull2"
                       >
-                        {formatNumber(type, totals[id].byYear[year])}
+                        {name}
```
mgwalker: I'm not sure what these changes are.  I think they came from another PR but then I don't know why they'd show up like this here.  I didn't do it, though!  😬

mgwalker:
Acknowledging that this one is hard to test pending [#1337 - Re-stickify sidebar](#1337), the work for it is done.  If #1337 isn't started yet, I'd prefer merging this one now since #1337 might need to make changes in some of the same places.  It'd save us from merge conflicts later on.

I'm happy to pair up with anyone who wants help testing it themselves - it's not hard to hack the sidebar into being sticky using browser dev tools!

jeromeleecms:
@lauraponce I just tested this out, but had to add this through the console to view the highlight "in action." 

image

@mgwalker I couldn't get the "Executive Summary of Activities" and "Program Budget Tables" to highlight - is this because the library depends on where the header sits on the page and this will get fixed when #1321 is completed?

mgwalker:
@jeromeleecms Yes, it depends on where the headers are on the screen.  If the headers never rise high enough, they won't be highlighted.  It could be mitigated by #1321, provided #1321 creates enough space at the bottom of the app to push those headers up high enough.  😄

jeromeleecms:
Tagging @nkkl for accessibility - When a state changes in the navbar (i.e. the highlighted section), does the screen reader need to announce that something was highlighted as the user scrolls through the app?

nkkl:
> When a state changes in the navbar (i.e. the highlighted section), does the screen reader need to announce that something was highlighted as the user scrolls through the app?

We don't need to worry about that, because the screen reader is already announcing the sections they are in as they page through them. Scrolling with a scroll wheel or similar isn't going to affect where the focus is set, and the screen reader will announce what element the user is on as soon as they change focus anyway.

web/src/actions/navigation.js
```diff
@@ -0,0 +1,24 @@
...
+
+export const jumpTo = waypoint => dispatch => {
+  if (waypoint) {
+    jumping = true;
+    setTimeout(() => {
+      jumping = false;
+    }, 100);
+    dispatch({ type: NAVIGATION_SCROLL_TO_WAYPOINT, data: waypoint });
+  }
+};
```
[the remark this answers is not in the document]
hbillings: javascript is the best and I love browsers

web/src/components/Section.js
hbillings: That seems like a very smart move!

web/src/containers/ApdPreviousActivityTableTotal.js
hbillings: 👏 🌟

web/src/containers/ConnectedWaypoint.js
hbillings: This is super smart. I love how easy this makes it to use!

web/src/containers/ConnectedWaypoint.test.js
```diff
@@ -0,0 +1,47 @@
...
+import { scrollTo } from '../actions/navigation';
+
+describe('connected waypoint higher-order component', () => {
+  const props = {
+    id: 'id',
+    scrollTo: sinon.spy()
+  };
+
+  beforeEach(() => {
+    props.scrollTo.resetHistory();
```
hbillings: I didn't see anything that was updating `history`, so I'm curious about why this is necessary in the `beforeEach`?

web/src/containers/IncentivePayments.js
hbillings: welp... they seem fairly benign, at least? 🤷‍♀️

web/src/containers/activity/CostAllocateFFP.js
hbillings: That's... whoa. I'd love to chat with you in person about this, because I feel like I sort of understand it, but not really.

web/src/containers/activity/CostAllocateFFPQuarterly.js
hbillings: good catch!

web/src/reducers/activities.selectors.js
hbillings: I'd love to capture some of the thinking here in our docs somewhere. It's really smart, but I also think I'd have a hard time understanding what's going on if I were coming into the code cold. (Hell, I kind of have a hard time with it now.)

web/src/reducers/navigation.js
hbillings: They feel similar enough to me for it to not feel weird.

Issue #1331, opened 2019-02-27
Implement CMS Design System in State Medical Director child section


jenniferthibault:
_Broken up for implementation from #1269_
_____
In order to build out the transition to the CMS design system, add the pieces needed to implement the system’s new components and styling for the State Medicaid Director and Medicaid Office Address section: https://hitech-apd.app.cloud.gov/apd#apd-state-profile-office

### This task is done when...
[ ] The section is not contained in an accordion
 - [ ] The headline typography is stylistically and semantically correct
- [ ] The [Form label component](https://design.cms.gov/components/form-label/) has been added and used for each field
- [ ] The [Text field component](https://design.cms.gov/components/text-field/) has been added and used for each field
- [ ] The [Select component](https://design.cms.gov/components/select/) has been added and used for the `State` field
 - [ ] Overall layout and interaction matches the [InVision prototype](https://gsa.invisionapp.com/share/BJQNX2KXQVW#/348460956_State_Profile_Contact)

![screen shot 2019-02-27 at 12 01 24 am](https://user-images.githubusercontent.com/11636908/53468246-67231a00-3a27-11e9-87be-21daff5d7f64.png)

PR #1397, opened 2019-03-27
Implement design system for Medicaid director/office section

mgwalker:
Implements #1331 

- switches from custom input types to CMS DS `<TextField>`
- switches "state" from text box to dropdown box
- removes ace-css classes

### DIFF
--- a/web/src/containers/ApdStateProfileMedicaidOffice.js
+++ b/web/src/containers/ApdStateProfileMedicaidOffice.js
@@ -23,9 +23,7 @@ class ApdStateProfile extends Component {
     return (
       <Fragment>
         <fieldset>
-          <legend>
-            {t(`${dirTRoot}.title`)}
-          </legend>
+          <legend>{t(`${dirTRoot}.title`)}</legend>
           <TextField
             name="apd-state-profile-mdname"
             label={t(`${dirTRoot}.labels.name`)}
@@ -47,9 +45,7 @@ class ApdStateProfile extends Component {
         </fieldset>
 
         <fieldset>
-          <legend>
-            {t(`${offTRoot}.title`)}
-          </legend>
+          <legend>{t(`${offTRoot}.title`)}</legend>
           <TextField
             name="apd-state-profile-addr1"
             label={t(`${offTRoot}.labels.address1`)}
@@ -71,7 +67,7 @@ class ApdStateProfile extends Component {
               className="ds-l-col--6"
               onChange={this.handleChange('medicaidOffice', 'city')}
             />
-            <div className="ds-u-clearfix ds-l-col--3">
+            <div className="ds-u-clearfix ds-l-col--6">
               <FormLabel component="label">
                 {t(`${offTRoot}.labels.state`)}
               </FormLabel>
@@ -83,7 +79,7 @@ class ApdStateProfile extends Component {
               >
                 {STATES.map(({ id, name }) => (
                   <option key={name} value={id.toUpperCase()}>
-                    {`${id.toUpperCase()} - ${name}`}
+                    {name}
                   </option>
                 ))}
               </Select>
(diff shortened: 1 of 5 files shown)


web/src/containers/ApdStateProfileMedicaidOffice.js
```diff
@@ -21,68 +22,83 @@ class ApdStateProfile extends Component {
 
     return (
       <Fragment>
-        <h3 className="mt0">{t(`${dirTRoot}.title`)}</h3>
-        <div className="mb3">
-          <Input
+        <h4 className="ds-h4 ds-u-margin-bottom--0">
```
mgwalker: Not 100% sure on the need for this 0 margin bottom.  There isn't much margin anyway, but looking at Jen's screenshots, it looked like there should be less.

web/src/containers/ApdStateProfileMedicaidOffice.js
```diff
@@ -21,68 +22,83 @@ class ApdStateProfile extends Component {
...
-              <Input
-                name="apd-state-profile-city"
-                label={t(`${offTRoot}.labels.city`)}
-                value={medicaidOffice.city}
-                onChange={this.handleChange('medicaidOffice', 'city')}
-              />
-            </div>
-            <div className="col col-6 sm-col-3 px1">
-              <Input
+          <div className="ds-l-row">
```
mgwalker: This is the only way I could think of to get the two inputs on the same row.  One `TextField` is 9 columns, by default, which is where the 6 and 3 below come from.

jeromeleecms:
@lauraponce Not sure if this is in scope for this issue, but is there any concerns with the way the state text box cuts of all the test?

image

The rest of the implementation looks very close to what was proposed in #1331 (some minor spacing issues). Will do some light accessibility testing (if it doesn't crash my computer) in a moment.

jeromeleecms:
Tagging @nkkl for accessibility testing!

nkkl:
Didn't find any major screen reader problems, but there's a nested (empty) text element showing up that should probably be skipped:

image

mgwalker:
@lauraponce So it turns out I was wrong about the textboxes being 9 columns.  They're 460px, max width.  I suspect it's still doable but I'll need to pair with Heather to figure it out because I'm now thoroughly confused about what the design system is doing in this case.  😄

web/src/containers/ApdStateProfileMedicaidOffice.js
hbillings: It occurs to me now that we could just nix the `h4` and use the `fieldset legend` as the section title. Suggested code incoming!

jeromeleecms:
@mgwalker @lauraponce Just checking in on this -- are we still looking to adjust the text box width for the city and state fields?

mgwalker:
After scheduling a meeting with Heather yesterday and forgetting to actually invite her to it, we have the real meeting today, to figure out how to get the spacing right.  That's the only thing holding this one up!

jeromeleecms:
Sounds good - I didn't realize that unintentional solo meeting you mentioned yesterday was about this PR. I'll await updates!

mgwalker:
@hbillings If you want to look at the whitespace around the fieldsets, it's ready for you.  Otherwise it's good to merge whenever.

web/src/styles/scss/forms.scss
```diff
@@ -0,0 +1,21 @@
...
+  border: 0;
+  max-width: $input-max-width;
+}
+
+/* only legends that are first-level descendents of sections should inherit this style */
+section > fieldset {
+  legend {
+    margin-top: $spacer-3;
+    font-size: $h4-font-size;
+    font-weight: bold;
```
hbillings:
```suggestion
    float: left;
    width: 100%;
```

mgwalker:
Here's what it looks like now:

Screen Shot 2019-04-03 at 2 10 05 PM

PR #1398, opened 2019-03-27
Switch login page to design system

### LINKED ISSUE #1376: Apply design system styles to log in page
As detailed out in #1306, the log in page should be updated to match the CMS design system styles.

Log in page new

with incorrect email/password error state:

Log in page new - error

with empty field error state:

Log in page new error 2

### This task is done when...
 - [ ] Log in page has been updated with error states
 - [ ] Design/content have reviewed it and approve

hbillings:
Closes #1376.

- Tweaks card layout format
- Factors header into its own component
- Makes password strength meter optional in password component

### DIFF
--- a/web/src/components/Header.js
+++ b/web/src/components/Header.js
@@ -1,10 +1,9 @@
-import PropTypes from 'prop-types';
 import React from 'react';
 import { Link } from 'react-router-dom';
 
 import { t } from '../i18n';
 
-const Header = ({props}) => {
+const Header = () => {
   return (
     <header className="header__slim">
       <div className="ds-l-container">
--- a/web/src/components/Header.js
+++ b/web/src/components/Header.js
@@ -1,0 +1,19 @@
+import PropTypes from 'prop-types';
+import React from 'react';
+import { Link } from 'react-router-dom';
+
+import { t } from '../i18n';
+
+const Header = ({props}) => {
+  return (
+    <header className="clearfix px2 py1 bg-white">
+      <div className="left">
+        <Link to="/" className="btn px0 bold caps">
+          {t('titleBasic')}
+        </Link>
+      </div>
+    </header>
+  );
+}
+
+export default Header;
--- a/web/src/components/Header.js
+++ b/web/src/components/Header.js
@@ -6,11 +6,15 @@ import { t } from '../i18n';
 
 const Header = ({props}) => {
   return (
-    <header className="clearfix px2 py1 bg-white">
-      <div className="left">
-        <Link to="/" className="btn px0 bold caps">
-          {t('titleBasic')}
-        </Link>
+    <header className="header__slim">
+      <div className="ds-l-container">
+        <div className="ds-l-row">
+          <div className="ds-l-col">
+            <Link to="/" className="site-title">
+              {t('titleBasic')}
+            </Link>
+          </div>
+        </div>
       </div>
     </header>
   );
--- a/web/src/components/PasswordWithMeter.js
+++ b/web/src/components/PasswordWithMeter.js
@@ -1,6 +1,6 @@
 import { Choice, TextField } from '@cmsgov/design-system-core';
 import PropTypes from 'prop-types';
-import React, { Component } from 'react';
+import React, { Component, Fragment } from 'react';
 import zxcvbn from 'zxcvbn';
 
 class Password extends Component {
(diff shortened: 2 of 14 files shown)


hbillings:
@lauraponce Oh! No, I should add the email-or-password error also. Thanks for reminding me!

web/src/containers/Login.js
```diff
@@ -36,50 +37,52 @@ class Login extends Component {
...
-                onChange={this.handleChange}
-              />
+        <Header />
+        <div className="card--container">
+          <div className="ds-l-container card">
+            <div className="ds-l-row">
+              <div className="ds-l-col--1 ds-u-margin-left--auto" />
+              <div className="ds-l-col--12 ds-l-sm-col--10 ds-l-lg-col--6">
+                <h1 className="ds-h1">Log in</h1>
+                {error && error==="Unauthorized" && (
```
hbillings: This feels a little fragile to me somehow, but I wasn't sure how else to approach it.

web/src/containers/Login.js
```diff
@@ -36,50 +37,52 @@ class Login extends Component {
...
-              <input
-                id="username"
-                type="text"
-                name="username"
-                className="input"
-                value={username}
-                onChange={this.handleChange}
-              />
+        <Header />
+        <div className="card--container">
```
hbillings: This adds the gray background. I'd really rather add a body class to the page that colors the whole page background gray because it would be more robust, but figuring out how to bubble that up seemed hard. :-p

web/src/containers/admin/MyAccount.js
```diff
@@ -74,11 +74,11 @@ class MyAccount extends Component {
             </Link>
           </div>
         </header>
-
-        <div className="ds-l-container card">
-          <div className="ds-l-row ds-u-radius ds-u-padding-y--5">
-            <div className="ds-l-col--1 ds-u-fill--white ds-u-margin-left--auto" />
-            <div className="ds-l-col--6 ds-u-fill--white ds-u-padding-y--5">
```
hbillings: I made an executive decision (ha) here: I only left in the classes that governed structure. I moved rules that govern style (border radius, background color) into the CSS.

web/src/styles/scss/password-meter.scss
```diff
@@ -2,12 +2,12 @@
 
 // This code comes from https://github.com/gladchinda/password-strength-react-demo/blob/master/src/App.css#L45
 
-.password-with-meter {
+.password-input {
```
hbillings: Since the meter's optional now, it seemed odd to have the classname reflect the meter disirregardless.

web/src/containers/Login.js
mgwalker: In #1399 I did this by having the redux action convert the error message into a human string:

https://github.com/18F/cms-hitech-apd/pull/1399/files#diff-1af493675e4b7123516353a186f85abeR79

Then I have it reject the promise with that message.  Back in the UI components, I have them `await`/`catch` the action so they'll know when it completes or fails:

https://github.com/18F/cms-hitech-apd/pull/1399/files#diff-c6793634e433d7ad10c9c8ea93ee8005R52

But I'm not 100% in love with this pattern because it's hooking the UI directly to the actions instead of going through the state, which I think is only a problem because it's inconsistent.  However, I wasn't sure how else to handle it.  Maybe we use one big `error` state component that tracks every error message.  🤔  Not a bad approach, I think.  Maybe I'll try updating #1399 after this one is merged.

All that said...  this seems fine for now.  😛

web/src/components/Header.js
```diff
@@ -0,0 +1,22 @@
...
+              {t('titleBasic')}
+            </Link>
+          </div>
+        </div>
+      </div>
+    </header>
+  );
+}
+
+export default Header;
```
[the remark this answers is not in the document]
hbillings: It could be! I was thinking that it could at least be the header for the admin/card layout sorts of pages that we've got several of (they all seem to use the same header, anyway).

web/src/components/PasswordWithMeter.js
```diff
@@ -81,14 +85,16 @@ Password.propTypes = {
...
+  showMeter: PropTypes.bool
 };
 
 Password.defaultProps = {
   compareTo: [],
   onChange: () => {},
   title: 'Password',
-  value: ''
+  value: '',
+  showMeter: false
```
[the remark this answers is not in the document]
hbillings: I stole the pattern from the feedback you gave me on adding the numbered section class ^_^

PR #1399, opened 2019-03-27
Implement frontend styles for admin forms

mgwalker:
- updates "create account" admin form
   animation of "create account" admin form
- updates "manage accounts" admin form
   animation of "manage accounts" admin form


web/src/actions/admin.js
```diff
@@ -82,20 +56,48 @@ export const getRoles = () => dispatch => {
...
+          break;
+        case 'add-user-weak-password':
+          message = 'The provided password is too weak';
+          break;
+        case 'add-user-invalid-phone':
+          message = 'The provided phone number is invalid';
+          break;
+        default:
+          break;
+      }
```
hbillings: I see what you mean about hooking UI directly to actions, but I also love that there's one place for all the error messaging. Cake, why can't we have you and eat you too?

api/routes/apds/delete.endpoint.js
```diff
@@ -8,7 +8,7 @@ const {
 } = require('../../utils.endpoint');
 
 describe('APD endpoint', () => {
-  describe('Delete/archive APD endpoint | DELETE /apds/:id', async () => {
```
mgwalker: SHUTTUP JEST

api/routes/me/put.js
```diff
@@ -39,7 +39,7 @@ module.exports = (app, { deserialize = deserializeUser } = {}) => {
       } catch (e) {
         res
           .status(400)
-          .send({ error: `update-self-${e.message}` })
+          .send({ error: `edit-self.${e.message}` })
```
mgwalker: Using a different delimiter between the two parts of the error message/token/whatever lets us start to group the messages logically.  This comes up further down in the PR.

I only changed the error tokens used by this PR, since they're the only ones actually being consumed.  We can update the others as we use them.  There shouldn't be too many of them.

api/routes/users/post.js
```diff
@@ -9,21 +9,23 @@ module.exports = (app, UserModel = defaultUserModel) => {
...
+
         // If these values are not passed in, don't set them at all, even
         // to undefined.  Undefined will make the data model explode.
         if (req.body.name) {
           posted.name = req.body.name;
         }
-        if(req.body.state) {
+        if (req.body.role) {
+          posted.auth_role = req.body.role;
+        }
```
mgwalker: Allow the authorization role to be set when the user account is created.  Previously the only way to set the role was to edit the account after it was created, which seems silly.  Jen's mockups included setting the role when creating the account, so I went ahead and added it.

api/routes/users/put.js
```diff
@@ -68,7 +68,7 @@ module.exports = (
           logger.verbose(`state [${req.body.state}] is invalid`);
           return res
             .status(400)
-            .send('update-user-invalid-state')
+            .send({ error: 'edit-account.invalid-state' })
```
mgwalker: And for some reason this goofball method just sent strings instead of whole objects.  :sigh:

api/seeds/development/users.js
```diff
@@ -3,7 +3,6 @@ const bcrypt = require('bcryptjs');
 exports.seed = async knex => {
   await knex('users').insert([
     {
-      id: 57,
```
mgwalker: Back in the old days, before you could log in, we hard-coded the user ID as `57` somewhere in the app.  Something to do with ketchup.  Accordingly, we made the development user have ID 47.  Now that you can actually log in and we don't have to synchronize the hard-coded values, let's un-hard-code this one.

web/src/actions/admin.js
```diff
@@ -1,45 +1,19 @@
-import { notify } from './notification';
```
mgwalker: I'm hopeful we'll eventually get rid of the toast/snackbar entirely, but for now, we can at least ditch it from the admin alerts.

web/src/actions/admin.js
```diff
@@ -1,45 +1,19 @@
...
-export const ADMIN_CREATE_USER_REQUEST = Symbol(
-  'admin : create user : request'
-);
-export const ADMIN_CREATE_USER_SUCCESS = Symbol(
-  'admin : create user : success'
-);
-export const ADMIN_CREATE_USER_ERROR = Symbol('admin : create user : error');
+export const ADMIN_GET_USERS_REQUEST = Symbol('admin : get users : request');
+export const ADMIN_GET_USERS_SUCCESS = Symbol('admin : get users : success');
+export const ADMIN_GET_USERS_ERROR = Symbol('admin : get users : error');
```
mgwalker: prettier is kind of chaotic. Sometimes it adds whitespace, then 3 months later it takes them back.

web/src/actions/admin.js
mgwalker: These actions now pass the error message along to the reducers and don't reject back to the UI.

web/src/components/CardForm.js
```diff
@@ -0,0 +1,86 @@
+import { Alert, Button, Spinner } from '@cmsgov/design-system-core';
```
mgwalker: The login, edit own account, admin edit account, and admin create account forms shared a ***bunch*** of code.  So I pulled out the common bits and put them here.

web/src/components/CardForm.js
```diff
@@ -0,0 +1,86 @@
...
+    <div className="ds-l-container">
+      <div className="ds-l-row card">
+        <div className="ds-l-col--1 ds-u-margin-left--auto" />
+        <div className="ds-l-col--12 ds-l-sm-col--10 ds-l-lg-col--6">
+          {!!success && <Alert variation="success">{success}</Alert>}
+          {!!error && <Alert variation="error">{error}</Alert>}
+
+          <h1 className="ds-h1">
+            {sectionName.length > 0 && (
+              <h6 className="ds-h6">{sectionName.toUpperCase()}</h6>
```
mgwalker: Calling this out because the dev tools complain about having an `h6` under an `h1`.  But I didn't want to use a different sized header than the class, and I was weary of using a `div` or `p` or something for what is basically a tiny pre-header.

web/src/components/PasswordWithMeter.js
```diff
@@ -27,7 +27,7 @@ class Password extends Component {
   };
 
   render() {
-    const { title, value, showMeter, ...rest } = this.props;
+    const { compareTo, title, value, showMeter, ...rest } = this.props;
```
mgwalker: Need to pull out `compareTo`, otherwise it gets caught up in `...rest` and passed into a `<div>` which makes React very unhappy.

web/src/components/PasswordWithMeter.js
```diff
@@ -54,27 +54,34 @@ class Password extends Component {
           </Choice>
 
           <TextField
-            hint={showMeter && 'A strong password is at least 9 characters, not a commonly-used word or phrase, and not too similar to the person’s name or email address.'}
+            hint={
+              showMeter &&
+              'A strong password is at least 9 characters, not a commonly-used word or phrase, and not too similar to the person’s name or email address.'
+            }
```
mgwalker: Prettier!

web/src/containers/Login.js
```diff
@@ -35,54 +36,39 @@ class Login extends Component {
...
+            value={username}
+            onChange={this.handleChange}
+          />
+          <Password
+            title="Password"
+            value={password}
+            onChange={this.handleChange}
+          />
+        </CardForm>
+      </Fragment>
```
mgwalker: Look how much smaller it is now!

web/src/containers/admin/CreateAccount.js
```diff
@@ -1,24 +1,46 @@
...
+        email: '',
+        password: '',
+        role: '',
+        state: '',
+        success: true
+      };
+    }
+
+    return { success: false };
+  }
```
mgwalker: We have to derive whether or not the form was successfully submitted.  It can't be stored in the app state because then the next time this form was loaded, it would default to a success state, which feels wrong.  It's kinda hacky logic, but it doesn't feel too awful.  Like, it doesn't make me do this:

web/src/containers/admin/EditAccount.js
```diff
@@ -1,115 +1,112 @@
...
 
+  static getDerivedStateFromProps({ error, working }, { hasFetched }) {
+    if (!hasFetched) {
+      return { hasFetched: working };
+    }
+
+    return {
+      success: !working && !error
+    };
+  }
```
mgwalker: This one is simpler than the last because the user comes from the application state rather than the component logic.

web/src/i18n/locales/en/errors.yaml
```diff
@@ -0,0 +1,17 @@
...
+add-account:
+  invalid: Email and password are required
+
+# Errors specific to editing an existing account
+edit-account:
+  invalid-state: The state selected for the account is invalid
+  invalid-role: The authorization role selected for the account is invalid
+
+# Errors specific to editing your own account
+edit-self:
```
mgwalker: Tagging @jameshupp to get eyes on this.  This is an initial proposal from dev for how error messages might be structured in yaml.  The idea is there are some defaults (e.g., if every weak password error should display the same way, we only need to have it in the yaml once), and then the ability to override on a "family" basis - so the `add-account` family of messages could override the weak password message if that was appropriate.

What do you think?

web/src/reducers/errors.js
```diff
@@ -0,0 +1,67 @@
...
+import { t } from '../i18n';
+
+const getError = (rawMessage, fallback) => {
+  const [errorFamily, message] = rawMessage.split('.');
+  let error = t(`errors.${errorFamily}.${message}`, { defaultValue: false });
+  if (error === false) {
+    error = t(`errors.${message}`, { defaultValue: false });
+  }
+  return error === false ? fallback : error;
+};
```
mgwalker: This function pulls apart the error message from the server, with format `family.message`.  It then tries to load the overridden message.  If that fails, then it goes to the default message.  If there's also not a default, it goes with the fallback.

web/src/reducers/errors.js
```diff
@@ -0,0 +1,67 @@
...
+  [ADMIN_CREATE_USER_SUCCESS, 'addAccount'],
+  [ADMIN_EDIT_ACCOUNT_SUCCESS, 'editAccount'],
+  [ADMIN_EDIT_ME_SUCCESS, 'editOwnAccount']
+]);
+
+const errorActions = new Map([
+  [ADMIN_CREATE_USER_ERROR, ['addAccount', 'Unknown error creating account']],
+  [ADMIN_EDIT_ACCOUNT_ERROR, ['editAccount', 'Unknown error editing account']],
+  [ADMIN_EDIT_ME_ERROR, ['editOwnAccount', 'Unknown error editing account']]
+]);
```
mgwalker: These two things map action symbols to state property names.  This ended up being a lot more compact than putting each of these into the switch below individually, and pretty readable I think?

mgwalker:
Note that the error messages will be incorrect in the preview PR because it needs an API update.

api/routes/users/put.endpoint.js
```diff
@@ -42,7 +42,7 @@ describe('users endpoint | PUT /users/:userID', async () => {
       expect(body).toMatchSnapshot();
     });
 
-    describe('when updating a valid user ID', async () => {
```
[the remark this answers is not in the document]
hbillings: It's like the PR equivalent of Groundhog Day.

api/seeds/development/users.js
hbillings: Now I want some Heinz 57 sauce.

web/src/containers/Login.js
hbillings: So much cleaner!!!

web/src/components/CardForm.js
hbillings: 👍 To do this semantically within an `h1`, use `span`. That way, a screen reader will pick the whole thing up as a top-level heading and should read it as `Administrator Manage Accounts` or whatever, but we can still style the tagline as not-an-h1.

```suggestion
              <span className="ds-h6 ds-u-display--block">{sectionName.toUpperCase()}</span>
```

hbillings: `display: block` because spans are inline elements, yadda yadda

web/src/containers/admin/CreateAccount.js
hbillings: Could we throw a comment in here to this tune? This seems like the kind of thing that someone will stumble on six months from now and be confused about. (Hi, Future Heather.)

web/src/containers/admin/EditAccount.test.js
```diff
@@ -23,12 +23,16 @@ describe('admin page > edit account', () => {
         },
         auth: {
           user: 'the current user'
-        }
+        },
+        errors: { editAccount: 'that error' },
+        working: { editAccount: 'hard, playing likewise' }
```
hbillings: 😂

web/src/reducers/errors.js
hbillings: I really like this approach! Makes sense and I like how the content and logic are separated a bit.

web/src/containers/admin/CreateAccount.js
mgwalker: Done!

web/src/i18n/locales/en/errors.yaml
jameshupp: Seems like the effect of this is that any particular string of text will exist in as few places as possible, and I can't think of anything more important that would override that. So yes! This is good!

Issue #1308, opened 2019-02-21
Review error messages -> update Save interaction


jeromeleecms:
This was the original to-do on this item:

Take a look at the error messages that appear in the app. 

Currently:
1) Password Error
2) No user name Error
3) APD Save Errors
4) APD Delete Errors

At present, however, we've rolled a bunch of issues around and what this issue is _now_ is to design a new in-progress Save interaction.

#### Current version of task list

- [x] @jameshupp Reviewed language on known error messages
- [x] Get input from @lauraponce or @austinhernandez about the rendering of the save success/failure toast message
- [x] Confirmed that we aren't doing form validation on Save anymore.
- [ ] The to-do list @austinhernandez defined in [his comment with up-to-date mockups](#issuecomment-479564669) is complete. (Yes, this is a to-do list in a to-do list. This issue got away from us a bit.)

#### Saving for later to write into a different issue

- [ ] The API's use of 401 and 403 errors is corrected (currently they're reversed)
- [ ] Update the 404 and 500 error messages (available [in the doc](https://docs.google.com/document/d/1D5kOCREcV0ib-VCSYRsqdUUB_HlphAc9GsHwKOOxI8c/edit) 🔒)
- [ ] Update the login error message (also in the doc)
- [ ] Check to see whether date validation has a bug (currently, text that shouldn't parse as dates isn't triggering an error)
- [ ] Update the date-based save error messages we have clarity on (also in the doc)

mgwalker:
@jameshupp When you get to this one, if you want to sync up for 5-10 minutes, I can show you where those errors originate.  In a nutshell, we have "error tokens" coming from the API, and we convert those into strings.  I'd be very happy if the error message content structure and error token structure were the same/similar and would be eager to make whatever changes to support that.

jameshupp:
To clarify: the two lists are separate things, right? They are not intended to be overlapping or have any kind of parent/child relationship? That's certainly my understanding, just want to be sure.

mgwalker:
Correct.  It's really just one long list, but definitely no overlapping or hierarchy.  It will only be one error, not any combination.

jameshupp:
I have made a worksheet where I've started working on this. @mgwalker it will surprise you not at all it has led me to some technical questions! I have left them in comments in the doc. Can you give them a look? (There's a shorter, easier design-related one in there, too, but the technical ones are far more important for this stage.)

[...]

PR #1412, opened 2019-04-02
Update "change password" behavior for editing account

mgwalker:
- by default, password is not editable
- adds a button to enable changing the password

change-password


web/src/components/Icons.js
```diff
@@ -8,12 +9,17 @@ import {
...
   faPlusCircle,
-  faSpinner
+  faSpinner,
+  faUnlock
 } from '@fortawesome/free-solid-svg-icons';
 
 import { faCircle, faClock } from '@fortawesome/free-regular-svg-icons';
 
+const LockIcon = () => <FontAwesomeIcon icon={faLock} />;
+const UnlockIcon = () => <FontAwesomeIcon icon={faUnlock} />;
```
mgwalker: I feel like we should have been doing this all along, honestly.

web/src/containers/admin/MyAccount.js
```diff
@@ -75,66 +82,99 @@ class MyAccount extends Component {
...
+                      value={phone || ''}
+                      onChange={this.handleEditAccount}
+                    />
+                    <TextField
+                      label="Role"
+                      name="position"
+                      ariaLabel="please enter your position or role ???"
+                      value={position || ''}
+                      onChange={this.handleEditAccount}
+                    />
```
mgwalker: This comment is for the line starting below.  Our conversation on #1397 is probably relevant to this one as well.  I put the disabled password field and the "change password" button on a row so that they'd line up.

On the button side, I added a `div` with the label classes to provide the right vertical spacing and wrapped the button in a `div` styled as a field to get its spacing right, too.  Hopefully that's not asinine.

mgwalker:
Tagging @lauraponce!  I think these are the interaction questions I have based on Jerome's observations:

1. If the user selects to change their password and then leaves the new password field empty, should that be an error?  Right now we ignore an empty password field, but should we display an error message to the user since they selected to change their password?

2. Should the user be able to toggle "change password" back to off?  It seems like if an empty password field is an error, then the answer to this part is almost certainly yes.

lauraponce:
1. Yes, if the password field is empty and they try to save that change, it should trigger an error. @mgwalker can it work similarly to Austin's recommendations in #1308?

2. For toggling the `Change password` link, if we leave it as-is and just allow it to be toggleable, I think it could cause some confusion because people might think that's the link to trigger the password to be changed. Would it make sense to change that link text to `Cancel` when it's in the unlocked state so they can cancel the change password action if they change their mind? @jeromeleecms

mgwalker:
For 1., #1399 implements error alerts for the admin forms, including this form, so I'm hesitant to do anything with errors here until that one is merged.  (It also adds some general purpose error infrastructure that will hopefully make all of this easier...).  So maybe we make this PR blocked by that one?

jeromeleecms:
#1: That sounds reasonable. No need to introduce a temporary fix if you have a better infrastructure solution in place.

#2: @lauraponce Would there be two "cancel" options then? My mental picture seems like that would be confusing -- would it be sufficient to have some help text that says you can cancel the change password by selecting cancel?

lauraponce:
@jeromeleecms the more I think about this, the more I wonder if we even need the `Change password` link to stay there at all after it's clicked. What if after they click it, it goes away and they can either change their password and save changes or cancel?

jeromeleecms:
@lauraponce I would be OK with that -- or just graying it out and appear not clickable(?)

If we have an existing design pattern for something like this, I would defer to that; otherwise, I would say -- whichever is easiest to implement.

mgwalker:
Okay, so when they click the "change password" link, it goes away entirely?  Should there still be an error if the new password field is blank, since there's no way to change your mind about changing the password?

lauraponce:
After chatting about this some more with my crit group, it sounds like just making that button toggleable is what folks are thinking seems right here (which is what @jeromeleecms originally suggested 😅) and that makes sense to me!

@mgwalker does that seem right to you also?

mgwalker:
Works for me! :high_five:

mgwalker:
* "Change password" is now a toggle - on or off
* If "change password" is ***disabled***, we ignore the password entirely
* If "change password" is ***enabled*** and the password is empty when the user clicks the save button, we display an error

ezgif com-optimize

jeromeleecms:
@mgwalker I'm getting this weird behavior in Firefox. 

1) With the password not toggled, when I select save, I initially get the error message, which than flashes to the success message.

2) With the change password toggled, I don't get an error message when I try to save with a blank field.

Thoughts?

mgwalker:
@jeromeleecms Will look into it

mgwalker:
@jeromeleecms Just tried it on Firefox and it's working normally for me.  🤔  If you had ever looked at the preview URL before about 1:00 PT, then maybe the previous JS is cached in your browser.  (#1422 is going to fix that!)  If clearing your cache doesn't help, maybe we can pair up on Monday to take a look?

web/src/containers/admin/MyAccount.js
```diff
@@ -110,12 +117,37 @@ class MyAccount extends Component {
...
-          <Password
-            className="mb2"
-            title="Change password"
-            value={password}
-            onChange={this.handleEditAccount}
-          />
+          <div className="ds-l-row ds-u-padding-x--2">
+            <TextField
+              label="Password"
+              ariaLabel="Current password"
```
hbillings: Since this doesn't read the current password aloud, can this say something like "Current password obscured for security" or somesuch? Otherwise I fear this might read out, "Current password: Asterisk Asterisk Asterisk"

mgwalker: Because the current password text field is disabled, the screen reader will skip it altogether and go straight to the "change password" button.  So that part seems okay, but the button text is the same no matter what state the form is in.  The only thing that changes is the lock/unlock icon.  So in bde586d, I added an ARIA label to the button.

PR #1417, opened 2019-04-03
Create admin landing page

hbillings:
Yay, admins can do adminy things from one convenient location!

Screen Shot 2019-04-03 at 5 53 17 PM

- Adds an `/admin` page
- Starts building out the `Header` component more -- `Log in` only appears when user is authenticated (the only other page this component is currently used on is the login page)


web/src/components/Footer.js
```diff
@@ -0,0 +1,24 @@
...
+              A federal government website managed and paid for by the <br/>U.S. Centers for Medicare and Medicaid Services.
+            </p>
+          </div>
+        </div>
+      </div>
+    </footer>
+  );
+}
+
+export default Footer;
```
hbillings: At some point, this will be the global footer! Right now it's just a start.

web/src/components/Header.js
```diff
@@ -1,22 +1,64 @@
...
-          <div className="ds-l-col">
-            <Link to="/" className="site-title">
-              {t('titleBasic')}
-            </Link>
+class Header extends Component {
+  handleLogout = e => {
+    e.preventDefault();
+    const { pushRoute } = this.props;
+    pushRoute('/logout');
+  };
```
hbillings: I stole this and the other auth checks from `TopBtns.js`. Since this is half-baked still, I thought it wasn't worth adapting the tests (but I can if @mgwalker disagrees with me).

web/src/containers/admin/AdminDashboard.js
```diff
@@ -0,0 +1,60 @@
...
+import Footer from '../../components/Footer';
+import Header from '../../components/Header';
+
+const AdminDashboard = () => {
+  return (
+    <Fragment>
+      <Header />
+      <div className="site-body ds-l-container">
+        <div className="ds-l-row">
+          <div className="ds-l-col--8 ds-u-margin-x--auto ds-u-padding-y--6">
```
hbillings: I'm not crazy about putting the padding in with utility classes (ideally, our spacing would be standardized), but I'm planning to move it once things start to look a bit more uniform.

web/src/containers/admin/AdminDashboard.js
```diff
@@ -0,0 +1,60 @@
...
+            <h1 className="ds-h1">
```
hbillings: For some reason, the design system doesn't style plain ol' `h` tags -- you have to use the class in conjunction with the heading. We could apply the class styles to unclassed headings at some point in the future.

web/src/containers/admin/AdminDashboard.js
```diff
@@ -0,0 +1,60 @@
...
+                    className="ds-u-display--block"
+                  >
+                      Create new user account
+                    </Link>
+                </h2>
+                <p>
+                  Make new accounts for federal analysts and state filers
+                </p>
+              </li>
+              <li className="ds-l-col--12 ds-l-md-col--6 ds-u-padding-right--0">
```
hbillings: I'm okay with styling this in utility classes, though, because it's unlikely we'll need to use these styles anywhere else on the site.

web/src/routes.js
```diff
@@ -20,6 +21,7 @@ const routes = [
     nonPrivate: false
   },
   { path: '/admin/edit-account', component: EditAccount, nonPrivate: false },
+  { path: '/admin', component: AdminDashboard, nonPrivate: false },
```
hbillings: I don't fully understand this, but when I put the `/admin` path preceding `/admin/*` paths, the former overwrote the latter. This is probably because I was expecting it to work like Django views. 😄

web/src/styles/scss/header.scss
```diff
@@ -0,0 +1,21 @@
...
+  background-color: $color-primary-alt-darkest;
+  padding: $spacer-1 0;
+  .site-title {
+    display: block;
+    margin: $spacer-1 0;
+    text-decoration: none;
+    font-weight: bold;
+    color: $color-white;
+    letter-spacing: 1.6px;
+  }
```
hbillings: This mostly existed before, I just moved it to its own file.

NAretakis:
How might one get to the Admin page? I can't seem to figure it out using /admin

web/src/components/Icons.js
```diff
@@ -9,7 +9,9 @@ import {
   faEdit,
   faExclamationTriangle,
   faPlusCircle,
-  faSpinner
+  faSpinner,
+  faUserCog,
+  faUserPlus
```
mgwalker: If these are new to the project, we need to update the `LICENSE.md` to cite them.

web/src/containers/admin/AdminDashboard.js
mgwalker: Would it make sense to have a set of React components for `<H1>`, `<H2>`, etc., that applied the classes?  Then if we used those components, we'd get the classes for free everywhere, and if we wanted to change the global appearance of a header, we could do it in one place?

```js
const H1 = ({ children }) => (<h1 className="ds-h1">{ children }</h1>);
```

But that may be a layer of abstraction too far, I dunno. :shrugging-person-made-of-symbols:

web/src/routes.js
mgwalker: React-router stops on the first path object that matches the page location, based on the order the path objects are inserted.  It's also basically a `startsWith` match, so it can be aggressive.  There are several workarounds - putting the more general paths later in the insertion order is one.  Another is to add an `exact` prop which tells the router to only match exactly, not generally.  I think I'd recommend we add the `exact` prop anyway, to avoid any weirdness if we add more pages in the future.

```suggestion
  { path: '/admin', component: AdminDashboard, nonPrivate: false, exact: true },
```

(The `nonPrivate` property is pulled from these objects, and all other properties are passed on to a react-router [`<Route>` component](https://github.com/ReactTraining/react-router/blob/master/packages/react-router/docs/api/Route.md).)

web/src/components/Icons.js
hbillings: Jen already did when she closed the design issue! ^_^

randyhart:
nice

PR #1422, opened 2019-04-04
Build optimizations: vendoring and "chunking"

mgwalker:
- splits up Javascript files and adds lazy-loading; should be transparent to users

This pull request adds build optimizations to make the user experience better.  It accomplishes this with three big changes:

1. **Vendoring** - Vendoring is when you separate the 3rd-party libraries from your application's custom code.  This PR changes our build so that instead of one giant `app.js` file, we have separate `app.js` and `vendor.js` files.  The ultimate goal here is to allow users' browsers to cache files more effectively.  Since our 3rd-party libraries rarely change, the `vendor.js` file will be the same for a long time, so when browsers ask for it, the server will tell them it hasn't changed since they fetched it last, and the browser will use its local cached copy instead of downloading it all again.  Our 3rd-party libraries are about 2.1 MB, so letting users' browsers cache it is a big transfer savings.
   - To make sure we bust through cacheing when necessary, the actual filename is `vendors~app.[content hash].js`, where `content hash` is a hash of the content of the final Javascript file.  It will only change when one of our 3rd-party libraries change, but having a unique filename will guarantee that we bust through the cache when we need to.
   - The custom file is also cache-busting, named `app.[content hash].js`
2. **Chunking** - Chunking is similar to vendoring in that it allows you to break your app into multiple output files.  In this PR, the `zxcvbn`library is chunked separately.  So in fact, we end up with three Javascript files:  `app.[hash].js`, `vendors~app.[hash].js`, `vendors~zxcvbn.[hash].js`.  Chunking is useful for pieces of the app that are large and not always needed - zxcvbn is about 400 KB, and only used when creating/changing a password.
3. **Lazy loading** - Lazy loading is when you don't load some of the code until you need it.  If you never need it, you never load it!  It's a nice way to save network transfer time and memory by only loading the things you need.  This PR makes zxcvbn lazy-loaded, which means most of the time you use the app, it just won't be loaded.  :tada:
   - This is especially useful for zxcvbn as it's the single largest library we use and it could reasonably be expected to grow (albeit slowly) over time.  It contains a bunch of common passwords in it that it uses to make sure user passwords are strong.  The list could grow as big compromises happen and new common passwords arise.

   - I'll update the wiki page on our frontend build and check this box when it's one


web/package.json
```diff
@@ -7,7 +7,7 @@
     "storybook": "start-storybook --ci -p 9001 -c .storybook",
     "build-storybook": "build-storybook -c .storybook -o .storybook-dist/public",
     "make-state-icons": "babel-node scripts/stateIcons.js",
-    "scaffold": "mkdir -p dist && cp src/index.html dist/ && cp ../Staticfile dist/ && cp -r src/static dist/",
+    "scaffold": "rm -rf dist && mkdir dist && cp ../Staticfile dist/ && cp -r src/static dist/",
```
mgwalker: Deletes the `dist` directory and starts over every time.  Otherwise those JS files with hashes in their filenames would accumulate in here forever.  In deployment it doesn't matter because we're starting with a clean filesystem every time anyway, but for building locally, it's nice to start with a clean slate.

web/package.json
```diff
@@ -152,7 +155,8 @@
   "babel": {
     "plugins": [
       "react-hot-loader/babel",
-      "@babel/plugin-proposal-class-properties"
+      "@babel/plugin-proposal-class-properties",
+      "@babel/plugin-syntax-dynamic-import"
```
mgwalker: This doohickey has to be turned on for webpack's dynamic import (née lazy-loading) to work, because shut up I guess

web/src/index.html
```diff
@@ -36,5 +34,6 @@
       gtag('js', new Date());
       gtag('config', 'UA-120133131-1');
     </script>
+    <!-- built scripts will get loaded below here -->
   </body>
 </html>
```
mgwalker: This was just being straight-up copied as part of the npm `build` script.  Now it's being used as a template by the webpack config, and the CSS and Javascript files are dynamically injected.  🎉

web/src/lazy/zxcvbn.js
```diff
@@ -0,0 +1,9 @@
+let zxcvbn = () => ({ score: 0 });
```
mgwalker: This file does the lazy loading.  It starts by creating a temporary object that does some default behavior for the things we need (we only use `zxcvbn().score`, so that's the only thing provided here).  Then once the real one is loaded, it replaces the temporary object with the real one.

web/src/lazy/zxcvbn.js
```diff
@@ -0,0 +1,9 @@
...
+
+export default (...args) => {
+  import(/* webpackChunkName: "zxcvbn" */ 'zxcvbn').then(real => {
```
mgwalker: Setting `/* webpackChunkName: "whatever" */` is a hint to webpack that this should be a separate file.

web/webpack.config.js
```diff
@@ -59,17 +70,42 @@ const config = {
       },
       {
         test: /\.(woff2?|ttf|otf|eot|svg)$/,
-        loader: 'file-loader'
+        use: [
+          {
+            loader: 'file-loader',
+            options: {
+              name: '[name].[ext]',
+              outputPath: 'static/fonts'
```
mgwalker: This change is about moving fonts into a subfolder, just so there's less stuff at the top level.  No biggie.

web/webpack.config.js
```diff
@@ -59,17 +70,42 @@ const config = {
...
+
+    // Inject our app scripts into our HTML kickstarter
+    new HtmlWebpackPlugin({
+      minify: { removeComments: true },
+      template: 'src/index.html'
+    }),
+    new HtmlWebpackIncludeAssetsPlugin({
+      assets: ['legacy.css', 'app.css'],
+      append: true,
+      hash: true
```
mgwalker: I tried to put the content hash in the CSS filenames, but then this plugin wouldn't find them.  I got this to work though: the filenames are constant, but they get a `?[hash]` appended when they're injected into the `index.html`.  Unfortunately the hash is for the build, so it'll change pretty much every time meaning we'll get no cacheing.  Maybe we shouldn't have this hash?  I'm not clear on the best way to do this one.

web/package.json
hbillings: 😂

web/webpack.config.dev.js
```diff
@@ -64,7 +66,16 @@ const config = {
     new webpack.EnvironmentPlugin({
       API_URL: null
     }),
-    new webpack.HotModuleReplacementPlugin()
+    new webpack.HotModuleReplacementPlugin(),
+    // Inject our app scripts into our HTML kickstarter
+    new HtmlWebpackPlugin({
+      minify: { removeComments: true },
```
hbillings: :party_parrot:

web/webpack.config.js
hbillings: Maybe we leave the hash in for now, while stuff is changing rapidly, and we take it out later when the styles are more stable?

mgwalker: That's smart.  @hbillings is smart.

PR #1426, opened 2019-04-08
Route admin user to admin dashboard

### LINKED ISSUE #1420: Admin Users should be routed to the admin dashboard when logging in
As part of the ATO requirements, Admin users must have separate accounts. After admin users have logged in, they should be routed to the admin dashboard that was described in #1273. Non-admin users should not be allowed to access the dashboard.

### This task is done when...
 - [ ] Upon logging in, Admin users will be routed to the Admin Dashboard where they can do admin-y related actions
- [ ] Non-Admin users who attempt to access the admin screens will be re-routed or given a permission denied notification
 - [ ] Someone else has looked at the code

mgwalker:
- admin users will be shown the admin dashboard on login
- admin pages will produce a 404 error for non-admin users
- closes #1420

This PR adds a general-purpose `<Dashboard>` component that checks the user's credentials and then renders either the `<StateDashboard>` or `<AdminDashboard>` accordingly.  It also updates our `routes.js` listing to include a property for `isAdmin` which flags that a route should only be accessible to administrators.  Those routes are wrapped in an `<AdminRoute>` component that will render a 404 for logged-in users who are not administrators.

The `<CardForm>` utility component is also updated to handle the cancel button itself, by navigating back one position in the browser history.  That made more sense to me than implementing that behavior in 4 places.  😄


web/src/routes.js
```diff
@@ -1,30 +1,23 @@
...
 import Logout from './containers/Logout';
 
-import StateDash from './containers/StateDashboard';
-
 const routes = [
-  { path: '/', component: StateDash, exact: true, nonPrivate: false },
-  { path: '/apd', component: ApdApplication, exact: true, nonPrivate: false },
-  { path: '/login', component: Login, nonPrivate: true },
-  { path: '/logout', component: Logout, nonPrivate: false },
+  { path: '/', component: Dashboard, exact: true, isPublic: false },
```
mgwalker: Changed `nonPrivate` to `isPublic` because that seems easier to read to me.  I'll revert if anyone disagrees, but the negation there always threw me off.  "Yes, it is non-private" vs "Yes, it is public."

jeromeleecms:
This is a weird behavior -- when I log into my admin account, it routes correctly. But when I log into my non-admin account, the cached APDs show up from my non-admin account. Clicking on the APDs doesn't open it.

web/src/components/CardForm.js
```diff
@@ -75,12 +78,13 @@ CardForm.propTypes = {
...
   primaryButtonText: ['Save changes', 'Working'],
   sectionName: '',
   success: false,
   working: false
 };
 
-export default CardForm;
+export default withRouter(CardForm);
+
+export { CardForm as plain };
```
hbillings: exporting this as `plain` and then re-importing it as `CardForm` -- what does that do??

web/src/routes.js
hbillings: Thank you!! This really confused me the first time I saw it. ^_^

web/src/routes.js
```diff
@@ -1,30 +1,23 @@
...
-  {
-    path: '/admin/create-account',
-    component: CreateAccount,
-    nonPrivate: false
-  },
-  { path: '/admin/edit-account', component: EditAccount, nonPrivate: false },
-  { path: '/admin', component: AdminDashboard, nonPrivate: false, exact: true },
-  { path: '/me', component: MyAccount, nonPrivate: false },
+  { path: '/create-account', component: CreateAccount, isAdmin: true },
+  { path: '/edit-account', component: EditAccount, isAdmin: true },
```
hbillings: I love that the `/admin` prefix is no longer necessary thanks to the smart routing you did in the files above!

web/src/components/CardForm.js
mgwalker: It lets you test the unwrapped component for better unit isolation.  The default export is wrapped by `withRouter` and I don't want that to interfere with testing just this component.

Issue #1090, opened 2018-11-02
Tech debt: deployment documentation is out of date


mgwalker:
Need to update deployment documentation.  Might be worth holding on to this issue until we're a little further along in our migration.

### This task is done when...
 - [ ] deployment documentation matches reality

mgwalker:
Recommend holding until #1014 and #1013 are done.  Documentation would be speculative right now and would just have to be updated again.

mgwalker:
Updated [dev deployment](https://github.com/18F/cms-hitech-apd/wiki/Dev%3A-Deployment), [infrastructure architecture](https://github.com/18F/cms-hitech-apd/wiki/Infrastructure-Architecture), and [authentication](https://github.com/18F/cms-hitech-apd/wiki/Authentication) pages in the wiki, with a small smattering of updates elsewhere as needed (things like library updates, mostly).

@NAretakis @jeromeleecms Do one of you want to review the updated docs before we close this?

NAretakis:
Awesome. No, don't need to review via this issue. If anything comes up during our ATO/compliance stuff, we'll track it there, if updates are needed. Closing this one out.

PR #1429, opened 2019-04-09
Automate deployment to AWS

mgwalker:
- adds a script for deploying the API to AWS EC2
  - also adds a `user-data` script that gets run in the EC2 instance after it starts up
  - scripts are commented, so I won't rehash them here
- refactors our CircleCI config file
  - adopts a consistent naming convention - frontend jobs are prefixed with `frontend`, backend with `backend`; ideally names are in increasing order of specificity, so something like `frontend deploy prod` (which is a real job now)
  - puts jobs in alphabetical order
  - lists workflows first, since those are most useful for understanding ***what*** the config does; the jobs describe the how, and those are now pushed further down
  - got rid of npm dependencies caches
    - CircleCI caches are persisted across builds, but that isn't helpful to us since we use `npm ci` which deletes the `node_modules` directory before installing
    - attaches a workspace instead and loads the two `node_modules` directories into it (a workspace is basically a shared drive between workflow jobs)
    - keeps the npm system cache, which npm will use with `npm ci` rather than pulling all the packages from the internet
  - separated build and deploy steps
    - the deployment job did four things: built the frontend, saved a zip file of the frontend into the CircleCI artifacts, and deployed the frontend and backend
    - these are now four distinct jobs: `backend deploy prod`, `frontend build`, `frontend deploy prod`, and `store artifacts`
    - separating these means we can reuse them more easily, for example...
  - preview deploy job now relies on `frontend build` instead of doing its own build
    - See?  😄  If we need to change how the frontend build happens, we can do it once and it'll cascade down to all the other places!  🎉

Because this PR change CircleCI job names, we'll also have to update our Github settings to change which jobs are required in order to merge pull requests.  If there are other open pull requests when we do that, they will not be mergeable until this PR is merged, so we want to delay that settings change as long as possible (we can wait until this PR is completely ready and approved).

  - there is a separate issue for updating the deployment documentation in the wiki, so that is _not_ covered in this PR


mgwalker:
@hbillings Happy to pair up sometime to talk through this, though if you take a look at it and it's not fairly easy to read, that's an indication that it needs a little more love.  There's also #1090 for updating deployment documentation, so any issues you run into here could be useful for that too.

mgwalker:
@randyhart Okay you can review it now.

PR #1430, opened 2019-04-09
Stick it to the sidebar

### LINKED ISSUE #1337: Re-stickify sidebar

hbillings:
Closes #1337, the best issue number ever. Re-introduces the sticky sidebar.

- Sticks the sidebar.
- Adds [StickyBits](https://www.npmjs.com/package/stickybits) polyfill for older browser support (StickyBits will check to see if it's needed before it starts listening to scroll events)


hbillings:
@jameshupp, this is ready for you!

web/src/containers/Sidebar.js
```diff
@@ -13,6 +14,10 @@ import { selectActivitiesSidebar } from '../reducers/activities.selectors';
 import { selectActiveSection } from '../reducers/navigation';
 
 class Sidebar extends Component {
+  componentDidMount() {
+    stickybits('.site-sidebar__sticky');
+  }
+
```
hbillings: I think this works? We'll see if @jeromeleecms gets a sticky sidebar on IE.

web/src/containers/Sidebar.js
```diff
@@ -206,7 +211,7 @@ class Sidebar extends Component {
                 {t('title', { year: '2018' })}
               </h1>
             </div>
-            <VerticalNav selectedId={activeSection} items={links} />
+            <VerticalNav selectedId={activeSection || "apd-state-profile-overview"} items={links} />
```
hbillings: With the new active item highlighting, there was no active item on page load, which caused a weird jump as the font bolded for the first time. That seems to be the only time there's no highlighted item, so this just sets the active item to the first section if `activeSection` is `null`.

hbillings:
@mgwalker I'm not sure why this `check docker-compose image tags` test is failing -- it says something about checksums not matching but I don't really know what's wrong. :-/

mgwalker:
If you run `npm run postinstall` in the web directory, it should update the `docker-compose` file and fix that.

The `docker-compose` file creates images and tags them with the hashes of the `package-lock.json` files.  There's a test to make sure the file hash and the image tags stay in sync.  If you install something new but the `docker-compose` file doesn't get updated, they'll be out of sync.  The `postinstall` npm script is supposed to keep them in sync but apparently sometimes it doesn't run right - and it _definitely_ doesn't run right if you do your `npm install` inside the docker container.

hbillings:
@mgwalker 🙌 No, that's what I love about stickybits -- it checks for `position: sticky` support, and quits if it's there. Most of the React-ready libs I looked at just appended a scroll listener regardless of whether a polyfill was necessary, which seemed like a potential performance issue.

jameshupp:
Same page as Nick: everything is awesome as long as there's also a behavior to make sure the active nav item is visible on screen. If auto-collapse is the easiest way to do that, let's do it! This is very cool!

PR #1440, opened 2019-04-11
Add APD unique name to builder header

### LINKED ISSUE #1388: Add unique ID to the APD builder
Created from #1346

So that users can have a unique identifier for keeping track of their APD, add the unique ID that's created to the top of the builder. For example:

image

### This task is done when...
 - [ ] ID has been added to the top of the builder

mgwalker:
There wasn't an overarching header for the builder, so I added that and put the APD name in it.

- adds an APD header, including unique name
- removes "<year> HITECH APD" text from the sidebar
  - fun fact, `<year>` was hard-coded as 2018.  Forever.  No matter what.  Oops!
- closes #1388

Screen Shot 2019-04-11 at 4 33 01 PM


web/src/components/Section.js
```diff
@@ -21,7 +21,7 @@ const Section = ({ children, id, isNumbered, resource }) => {
   const helptext = t([resource, 'helpText'], { defaultValue: false });
 
   return (
-    <section id={id} className={isNumbered && 'numbered-section'}>
+    <section id={id} className={isNumbered ? 'numbered-section' : ''}>
```
mgwalker: This was my suggestion, but it turns out React complains about `className` having a boolean value.  My mistake!

web/src/styles/scss/section.scss
```diff
@@ -32,6 +32,10 @@ body {
   padding-top: $spacer-2;
   padding-left: $number-padding-size;
   border-top: 1px solid $color-gray-dark;
+
+  &:first-of-type {
+    margin-top: $spacer-2;
+  }
```
mgwalker: The first section on the APD builder page (at `/apd`) needs to have less margin from the title to match the mockup.  An attempt was made.  😆

mgwalker:
@jeromeleecms We can get those aligned, but we'd need to wait until the site-wide header is implemented.  Right now, the yellow nav buttons at the top are actually part of the APD builder content, so they're pushing all the other builder content down - but they don't affect the sidebar.  Once we have the site-wide header, it'll push everything down equally.  If we then wanted to get the alignment perfect, we could.  In short: doing it now would be a temporary fix.  😛

web/src/containers/ApdApplication.test.js
```diff
@@ -40,7 +40,9 @@ describe('apd (application) component', () => {
     const state = {
       apd: {
         data: {
-          id: 'bloop'
+          id: 'bloop',
+          name: 'florp',
+          years: ['dinkus', 'dorkus']
```
hbillings: ah yes, the years my brother and I were born in

web/src/components/Section.js
hbillings: Thanks for fixing it!

web/src/styles/index.scss
```diff
@@ -13,3 +13,7 @@
 @import '/scss/password-meter';
 @import '/scss/print';
 @import '/scss/section';
+
+h1.apd--title {
+  margin-left: 72px;
+}
```
hbillings: In my open PR, I added a `type.scss` file. This would be right at home there -- want to go ahead and create it in your PR also?

web/src/containers/ApdApplication.js
```diff
@@ -71,6 +73,10 @@ class ApdApplication extends Component {
           <Sidebar place={place} />
           <div className="site-main p2 sm-p4 md-px0 ds-l-col--9">
             <TopBtns />
+            <h1 className="ds-h1 apd--title">
```
hbillings: I can't figure out what targets the `apd--title` class. If it's not used, let's take it out, otherwise, I need more coffee!

web/src/containers/ApdApplication.test.js
mgwalker: I spelled it wrong.  It's [Dorcas](https://en.wikipedia.org/wiki/Dorcas_Hoar).  😬

web/src/styles/index.scss
mgwalker: Done!

Issue #1269, opened 2019-02-07
Restyle the State Profile Contact Info using the CMS Design System


jenniferthibault:
_Part of #1250_

In order to transition a few more complex components and layouts to the [CMS Design System](https://design.cms.gov/), restyle the State Profile Contact Info portion of the form. 

### This task is done when...
 - [x] We feel good about the component and layout restyling
 - [x] Design files have been updated with new library components
 - [x] A new issue has been created for development changes for the concept
 
**Current styles:**
![screen shot 2019-02-06 at 11 41 25 pm](https://user-images.githubusercontent.com/11636908/52390910-ce712e00-2a68-11e9-8ba1-25ecb3cd6d66.png)
![screen shot 2019-02-06 at 11 41 33 pm](https://user-images.githubusercontent.com/11636908/52390911-ce712e00-2a68-11e9-8278-883d4f43a909.png)

jenniferthibault:
Holy moly I think I've cracked it!

## What I've learned about the design system & it's components through rebuilding this section
### The CMS design system **does not** have an accordion component. 
This was an initial snag because we use accordions heavily in the prototype. We could choose to use ours, or build one, but the accordions were previously causing some headache in the layout, so I wanted to see where they came from before making that choice. After checking in with @lauraponce on previous research, here's that summary:
 > no research points to the fact that we should definitely use the accordions. But the accordions allow for states to quickly see what’s next. And potentially for the fed experience, the accordions could be expanded by default to make it easier for them to go through the review process (more research needs to be done for the fed users though to validate that hypothesis)

Based on this, I recommend we try releasing the content from the accordions, and instead rely on:
- The new navigation styling (which does not truncate the heading text)
- Numbering the sections to help people orient them in their progress
I think this will both make our codebase simpler since it removes a custom component, AND make the app space appear visually cleaner since we're removing some containers (we were backing ourselves into a corner through so many nested boxes)

### The CMS design system has components and patterns that solve a few of our pesky old problems
- Their [React Checkbox/ `<Choice>`](https://design.cms.gov/components/choice/) component has a prop called `checkedChildren` that would allow us to do what we planned in https://github.com/18F/cms-hitech-apd/issues/1173 to replace the yes/no buttons.
[...]

PR #1451, opened 2019-04-12
Implement design system in APD key personnel section

### LINKED ISSUE #1330: Implement CMS Design System in Key Personnel child section

mgwalker:
- Implements the design @jenniferthibault outlined in #1269 for the key personnel portion of the state profile section. A couple of exceptions, which @lauraponce and I worked through after I ran into some code-y problems:
   - In the mockup, there is an "Add" button at the bottom of the key personnel form.  We changed it to "Done" because by the time the user is filling out this form, the person has already been added. The ultimate purpose of the button is to collapse the personnel form into the review view, so we may revisit this in the future to change the button text or otherwise change the way users collapse that form.
   - The mockup has a heading for "New Contact" for filling out the information for a new person, but there's no way in the code to distinguish between filling out a new person versus just editing a previous person.  We changed the heading to "Additional APD Point of Contact" to match the heading for the primary person, and not try to distinguish between when people were added.
   - Issue #1330 has a really handy checklist. One of the checkboxes says that there should be a cancel button for adding personnel costs.  We decided to remove the cancel button because the user can just click the "No" radio button again, and the cancel button looks confusing so close to the "Add"/"Done" button.

   - ### Initial view, new APD
     Primary user is created and expanded by default.  Also defaults to not having any direct costs.
     Screen Shot 2019-04-12 at 2 51 29 PM

   - ### With direct costs
     Screen Shot 2019-04-12 at 2 51 51 PM

   - ### Collapsed person
     Screen Shot 2019-04-12 at 2 52 43 PM
 and expanded by default.  Also defaults to not having any direct costs.

   - ### Another person added
     Screen Shot 2019-04-12 at 2 53 04 PM

- For print media, only the review view is rendered, fully expanded, with all of the data fields displayed.
   Screen Shot 2019-04-12 at 3 05 03 PM

- closes #1330


web/src/actions/apd.js
```diff
@@ -22,7 +22,6 @@ export const SAVE_APD_REQUEST = 'SAVE_APD_REQUEST';
 export const SAVE_APD_SUCCESS = 'SAVE_APD_SUCCESS';
 export const SAVE_APD_FAILURE = 'SAVE_APD_FAILURE';
 export const SELECT_APD = 'SELECT_APD';
-export const SET_KEY_PERSON_PRIMARY = 'SET_KEY_PERSON_PRIMARY';
```
mgwalker: Since the first key person is now always primary and that one can't be removed, no need to set which one is primary.

wipe it away

web/src/containers/ApdStateKeyPerson.js
```diff
@@ -0,0 +1,248 @@
...
+            ))}
+            <div className="ds-l-col--auto">
+              <FormLabel>Total</FormLabel>
+              <Dollars>{personTotalCost(person)}</Dollars>
+            </div>
+          </div>
+        }
+      >
+        Yes
+      </Choice>
```
mgwalker: Should these `<Choice>` components somehow be tied back to the `<FormLabel>` to link the two for screenreaders?  I'm not sure how to do that in the design system components, but if it's the right thing to do, I'll figure it out.

figure it out

web/src/containers/ApdStateKeyPerson.js
```diff
@@ -0,0 +1,248 @@
...
+              {primary ? (
+                <Fragment>
+                  <br />
+                  Primary APD Point of Contact
+                </Fragment>
+              ) : null}
+            </Review>
+          )}
+        </div>
+        <div className="visibility--print">
```
mgwalker: I don't recommend this as a pattern going forward, but it feels like a pretty powerful demo to me of how useful it'd be to have a designed-for-print page with no interactions, no math, no logic (or very little, anyway), etc.  Just display the danged thing, and end up with fewer pages!

printers gone wild

web/src/i18n/locales/en/apd/stateProfile.yaml
```diff
@@ -43,7 +43,7 @@ keyPersonnel:
     email: Email
     percentTime: Percent commitment to program
     primary: 'This person is the primary APD point of contact:'
-    hasCosts: 'This person has costs directly attributable to the program:'
+    hasCosts: Does this person have costs directly attributable to this project?
```
mgwalker: @jameshupp 

shhh

web/src/util/serialization/activities.test.js
```diff
@@ -186,7 +186,7 @@ describe('APD activity serializer', () => {
       expect(fromAPI(activityAPI, ['1947', '1949'])).toMatchObject(
         expect.objectContaining({
           id: 998,
-          key: expect.stringMatching(/[a-f0-9]{8}/),
+          key: expect.stringMatching(/^[a-f0-9]{8}$/),
```
mgwalker: Tightens up these regexes.  The entire string should be matched, hence adding the start-of-line and end-of-line markers to the check.  There was one set of these that was only looking for 6 characters and without the start/end markers, it passed just fine.  But it was wrong.

sad ant

web/src/i18n/locales/en/apd/stateProfile.yaml
jameshupp: This should say `program` not `project` to maintain consistency. [This exact discrepancy came up a little while back](https://github.com/18F/cms-hitech-apd/issues/1145#issuecomment-471100068); both the activities and the personnel costs support a _program_. The antecedent of _project_ isn't really specified in the app.

But phrasing it as a question is great!

jeromeleecms:
This looks great!

Some comments/questions:

1. Was there a rationale to make the print view and collapsed view of the individual different? This is not deal breaker for me, but seems like we'll have to revisit this when we do reviews in the app in subsequent releases. After playing with the app more, it may be worth sitting down and reviewing what the review "view" of the APD would be like (instead of just showing the reviewer the filled in APD builder).. though layer 2+ discussion :)

2. The calculated dollars don't line up with the form field. 

image

Can we put the calculated dollars into a grayed out form field so it line up horizontally (barring any known/documented usability issues with that)? Maybe a question for @lauraponce?

mgwalker:
@jeromeleecms:
1. The print view shows all of the data the user entered, whereas the collapsed view only shows some of it.  The collapsed view comes from Jen's mockups and I think it's just meant to allow state users to identify who the key people are so they know who to edit and can make sure they've entered everyone.  But I assumed a CMS reviewer would want to see all of the information, otherwise we shouldn't be collecting it.  😉  But that said, I think you're right, this and a lot of other things will need to be rethought when it's time to implement a reviewer mode.  (I will keep gently pushing that when that time comes, we investigate whether it'd be better to build a different web app for CMS users, but we're not there yet!)
2. I'll get some time with Heather to see about getting that to line up.  The disabled text box would work visually, but screen readers don't read the content of disabled text boxes.  (I think they don't even announce that the text boxes exist.)

web/src/containers/ApdStateKeyPerson.js
```diff
@@ -0,0 +1,250 @@
...
+                  size="small"
+                  mask="currency"
+                  value={person.costs[ffy]}
+                  onChange={handleYearChange(ffy)}
+                />
+              </div>
+            ))}
+            <div className="ds-l-col--auto">
+              <FormLabel>Total</FormLabel>
+              <div className="form-value-input-aligned">
```
mgwalker: This is not an amazing class name, but I couldn't think of anything.  It's a computed form value that should be visually aligned like it's inside a text box but without being in a text box.

web/src/styles/scss/forms.scss
```diff
@@ -20,3 +20,7 @@ section > fieldset {
     margin-top: 0;
   }
 }
+
+div.form-value-input-aligned {
+  margin: ($spacer-half + $input-padding) 0;
+}
```
mgwalker: Add the same vertical spacing that the text box elements get.

hbillings:
When I load an APD with existing data, the first user is expanded by default. It seems like we probably should collapse fields with data in them by default? (Also it's weird because when it's expanded, the `done` button is visible, but... I didn't do anything! I swear!)

web/src/containers/ApdStateKeyPerson.js
```diff
@@ -0,0 +1,250 @@
...
+                        onClick={remove}
+                      >
+                        Remove
+                      </Button>
+                    </Fragment>
+                  )}
+                </div>
+              }
+            >
+              {person.position}
```
hbillings: I think it might make sense to make these into an unordered list, so that each item is its own `li`.

web/src/containers/ApdStateKeyPerson.js
```diff
@@ -0,0 +1,250 @@
...
+                  <br />
+                  Primary APD Point of Contact
+                </Fragment>
+              ) : null}
+            </Review>
+          )}
+        </div>
+        <div className="visibility--print">
+          <Review heading={`${number}. ${person.name}`}>
+            {person.email}
```
hbillings: A `ul` would also work well here. There's a class `ds-c-list--bare` that removes the bullet points and extra padding!

web/src/util/serialization/activities.test.js
hbillings: judging you

web/src/containers/ApdStateKeyPerson.js
mgwalker: YESSS, thank you!  When I was doing it, I was thinking to myself this was surely not the best approach, but a list just...  didn't...  occur to me.

PR #1462, opened 2019-04-15
Fix issue where admins can still see APDs

mgwalker:
- If a user had previously logged in with a non-admin account
- and the user had viewed an APD
- and the user logged out
- and the user browsed to `{url}/apd`
- and the user logged in with an admin account
- and the admin account was associated with the same state as the non-admin account
- then the user would be redirected to the admin dashboard briefly and then redirected to the APD page

- for admin users, does not flag that the page should be redirected to `{url}/apd`
- the APD page redirects admin users to `{url}/` (the dashboard)
- updates the database seed data to restrict admin users' permissions; this helps satisfy [NIST 800-53 AC-6](https://nvd.nist.gov/800-53/Rev4/control/AC-6)

Separately, I will update the authorization roles in the staging database to admin users no longer have permission to view/edit/submit documents, etc.

  - https://github.com/18F/cms-hitech-apd/wiki/Dev%3A-Users-And-Roles


api/seeds/shared/roles-and-activities.js
```diff
@@ -32,16 +32,7 @@ const roleToActivityMappings = {
...
-    'submit-clearance',
-    'edit-comments',
-    'submit-document',
-    'submit-state-response',
-    'create-draft',
-    'edit-document',
-    'edit-response',
-    'delete-users',
-    'view-document'
+    'delete-users'
```
mgwalker: [NIST 800-53 AC-6(2)](https://nvd.nist.gov/800-53/Rev4/control/AC-6) says admins shouldn't have access to non-admin functions, so this is a useful change for ATO.

web/src/actions/apd.js
```diff
@@ -33,7 +34,12 @@ export const WITHDRAW_APD_SUCCESS = Symbol('withdraw apd success');
 export const WITHDRAW_APD_FAILURE = Symbol('withdraw apd failure');
 
 export const SET_SELECT_APD_ON_LOAD = 'SET_SELECT_APD_ON_LOAD';
-export const selectApdOnLoad = () => ({ type: SET_SELECT_APD_ON_LOAD });
+export const selectApdOnLoad = () => (dispatch, getState) => {
```
mgwalker: This function is called when a user lands on the `/apd` page before their list of APDs has loaded.  It sets a flag in our application state to load the previously-selected APD once the list of APDs has loaded.  This process could be simplified since we changed how the APD list is loaded, but a story for another time, perhaps.

Anyway, this change just disables setting that flag if the logged-in user is an admin.

web/src/containers/ApdApplication.js
```diff
@@ -44,10 +48,15 @@ class ApdApplication extends Component {
...
       apdSelected,
       dirty,
+      isAdmin,
       place,
       selectApdOnLoad: dispatchSelectApdOnLoad
     } = this.props;
 
+    if (isAdmin) {
+      return <Redirect to="/" />;
+    }
```
mgwalker: Always redirect to the dashboard for admin users.

web/src/containers/ApdApplication.js
```diff
@@ -89,20 +98,16 @@ class ApdApplication extends Component {
...
-  }
-}) => ({
-  apdSelected: !!data.id,
-  dirty,
-  place: state
+const mapStateToProps = state => ({
+  apdSelected: getIsAnAPDSelected(state),
+  dirty: getIsDirty(state),
+  isAdmin: getIsAdmin(state),
+  place: getUserStateOrTerritory(state)
```
mgwalker: Converted all of these to selectors because it was cleaner.  A little piece at a time...

Issue #1468, opened 2019-04-16
Implement top navigation (horizontal header)


austinhernandez:
We ready to start implementing #1366  !

After a dev + design meeting we decided on exploring the following header:

![Header](https://user-images.githubusercontent.com/23194720/55653706-5780b900-57bd-11e9-9c8e-32029498ca4f.png)

![Header Hover](https://user-images.githubusercontent.com/23194720/55653705-5780b900-57bd-11e9-9df1-56a03037e8ee.png)

Notes: 

- The rationale for the header is the name or email will indicate system status: "You are logged in as `xxxx xxxx` and doing things on that account. Should you need to make changes to the profile or log out it is on this account."
  - Not sure if the `xxxx xxxx` will be the email address or user name, the goal is just to let a user know that's the account changes are being made under in the case of a shared computer or multiple log ins. 
  - Information will pull from information input on the `manage account` page
- The "state dashboard" button is now in the location of a "home" link or logo. It takes you back to the "start page" whether thats the APD builder / APD list screen or the manage users screen for the admin. 
  - @jameshupp still working on content. 
  - There is an arrow back button to indicate you are a one level down from the homepage.
- There are no horizontal header interactions on the current design system, [only vertical](https://design.cms.gov/components/vertical-nav/) so I am using [the interaction styles from USWDS V2 for horizontal (top) nav](
 https://federalist-proxy.app.cloud.gov/preview/uswds/uswds/release-2.0/components/preview/header--basic.html) also [here](https://components.designsystem.digital.gov/components/detail/header--default.html)
- Icons on [Font Awesome](https://fontawesome.com/icons/sign-out-alt?style=solid)
- Background color for the header is `primary darkest`

### This task is done when...
 - [x] Header is implemented with dropdown states as referenced above
 - [x] New design system horizontal header pattern is added to the library (in USWDS, but not CMS design system.)
 - [x] We are able to see the header in situ, and the design is ready to be reviewed by a designer (so we finalize content and give feedback on interactions)

mgwalker:
What should the `< NY APD Home` text be when the user is on a page they can't go "back" from?  Eg, the consent banner, the login page, or the dashboard.

jameshupp:
Does it make sense for it to be blank in those contexts? I don't know what that does to the patterns of people's usage — will they find it again on internal pages? — but if it's one-way traffic (or they're already on that page) we may not need anything at all.

jameshupp:
[...]

PR #1470, opened 2019-04-16
Login consent screen

### LINKED ISSUE #1424: Government Consent Screen before User Login
Government Consent Screen should appear prior to user logging into the app (ATO Requirement).

Current recommendation is to take the one from cloud.gov, but need to be have CMS lawyers approval. Also available, is the language from NIST. 

Reference: https://login.fr.cloud.gov/login

### This task is done when...
 - [ ] User is presented with the Government Consent Screen
 - [ ] User can access the login screen after the Consent Screen

mgwalker:
- adds a government IS consent banner before allowing access to the web app
  - Once a user logs in, we store a cookie in their browser that's valid for 3 days and won't show them the consent banner again as long as the cookie is valid.  The cookie expiration is extended when the user logs in again, but not every time they visit the app.
- closes #1424

### DIFF
--- a/web/src/components/ConsentBanner.js
+++ b/web/src/components/ConsentBanner.js
@@ -1,6 +1,6 @@
 import { Button } from '@cmsgov/design-system-core';
 import PropTypes from 'prop-types';
-import React, { PureComponent } from 'react';
+import React, { Fragment, PureComponent } from 'react';
 
 let cookie = name => {
   const cookieMap = (document.cookie || '').split(';').reduce((c, s) => {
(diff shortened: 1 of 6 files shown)


web/src/components/ConsentBanner.js
```diff
@@ -0,0 +1,135 @@
...
+      this.agreeAndContinue();
+      return null;
+    }
+
+    return (
+      <div className="card--container">
+        <div className="ds-l-container">
+          <div className="ds-l-row card">
+            <div className="ds-l-col--1 ds-u-margin-left--auto" />
+            <div className="ds-l-col--12 ds-l-sm-col--10 ds-l-lg-col--6">
```
mgwalker: This is all borrowed from the admin form card work @hbillings did.

mgwalker:
Added the header and tweaked the agree button text.

@jeromeleecms Does the "CMS HITECH APD" header work?  It uses the same header that the admin forms do, so hopefully as the site-wide header is developed and refined, this will get those benefits (but without any links to anything, since you haven't consented yet...!)

jeromeleecms:
The header "CMS HITECH APD" works. We have some branding questions down the road that can better inform this. 

Can you update the color of the header to the one that Austin put in for #1468 ("primary darkest")? That way we won't have to come back and do it later.

mgwalker:
It's the same header, so when #1468 is implemented, this one should be updated for free.  😉

web/src/components/ConsentBanner.js
hbillings: aww yiss!!

web/src/components/ConsentBanner.js
```diff
@@ -0,0 +1,135 @@
...
+              {showDetails ? (
+                <Fragment>
+                  <p>
+                    This warning banner provides privacy and security notices
+                    consistent with applicable federal laws, directives, and
+                    other federal guidance for accessing this Government system,
```
hbillings: @jameshupp do we need to caps `Government` in any of the instances here?

mgwalker: If yes, let's create an issue to correct it!

PR #1471, opened 2019-04-16
Minor fixes

mgwalker:
This PR addresses some runtime warnings from React as well as some tests that were causing minor problems (that would eventually become breaking issues).

- no visible changes

### DIFF
--- a/web/src/components/PasswordWithMeter.test.js
+++ b/web/src/components/PasswordWithMeter.test.js
@@ -2,7 +2,13 @@ import { shallow } from 'enzyme';
 import React from 'react';
 import sinon from 'sinon';
 
+// Import the lazy-loading zxcvbn so we can use jest's mocking functionality
+// eslint-disable-next-line no-unused-vars
+import zxcvbn from '../lazy/zxcvbn';
+
 import Password from './PasswordWithMeter';
+
+jest.mock('../lazy/zxcvbn');
 
 describe('PasswordWithmeter component', () => {
   it('renders with no props', () => {
@@ -11,19 +17,21 @@ describe('PasswordWithmeter component', () => {
 
   it('renders with various passwords', () => {
     // weak, strength = 0
-    expect(shallow(<Password value="a" />)).toMatchSnapshot();
+    expect(shallow(<Password value="a" showMeter />)).toMatchSnapshot();
 
     // weak, strength = 1
-    expect(shallow(<Password value="abCD" />)).toMatchSnapshot();
+    expect(shallow(<Password value="abCD" showMeter />)).toMatchSnapshot();
 
     // weak, strength = 2
-    expect(shallow(<Password value="abCD321" />)).toMatchSnapshot();
+    expect(shallow(<Password value="abCD321" showMeter />)).toMatchSnapshot();
 
     // good, strength = 3
-    expect(shallow(<Password value="abCD321!," />)).toMatchSnapshot();
+    expect(shallow(<Password value="abCD321!," showMeter />)).toMatchSnapshot();
 
     // great, strength = 4
-    expect(shallow(<Password value="abCD321!,@_" />)).toMatchSnapshot();
+    expect(
+      shallow(<Password value="abCD321!,@_" showMeter />)
+    ).toMatchSnapshot();
   });
 
   it('shows the password', () => {
--- a/web/src/containers/BudgetSummary.test.js
+++ b/web/src/containers/BudgetSummary.test.js
@@ -40,7 +40,7 @@ describe('budget summary component', () => {
         <DataRow
           category="category text"
           data={{
-            total: { total: 0 }
+            total: { medicaid: 0, federal: 0, state: 0, total: 0 }
           }}
           entries={[]}
           title="title text"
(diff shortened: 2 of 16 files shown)


web/src/components/PasswordWithMeter.test.js
```diff
@@ -2,28 +2,36 @@ import { shallow } from 'enzyme';
 import React from 'react';
 import sinon from 'sinon';
 
+// Import the lazy-loading zxcvbn so we can use jest's mocking functionality
+// eslint-disable-next-line no-unused-vars
+import zxcvbn from '../lazy/zxcvbn';
```
mgwalker: The lazy-loading wrapper uses the `import()` function, which isn't actually supported by, like, anything yet, except Webpack.  Jest would reject a promise when `import()` was called, and that rejection wasn't handled anywhere so it bubbled up to an `unhandledRejection`.  It didn't interfere with our tests completing properly, but a future version of Node.js will crash on unhandled rejections.

Jest has built-in module mocking, though.  In your test, you import the module you want to mock, call `jest.mock(modulePath)`, and then when the thing being tested imports the module, Jest will give it the mock instead.  Handy dandy.

web/src/components/PasswordWithMeter.test.js
```diff
@@ -2,28 +2,36 @@ import { shallow } from 'enzyme';
...
+
 describe('PasswordWithmeter component', () => {
   it('renders with no props', () => {
     expect(shallow(<Password />)).toMatchSnapshot();
   });
 
   it('renders with various passwords', () => {
     // weak, strength = 0
-    expect(shallow(<Password value="a" />)).toMatchSnapshot();
+    expect(shallow(<Password value="a" showMeter />)).toMatchSnapshot();
```
mgwalker: It turns out these tests were also just wrong, because `showMeter` defaults to `false`, so all the snapshots were identical - no meter.  😬

web/src/containers/BudgetSummary.test.js
```diff
@@ -40,7 +40,7 @@ describe('budget summary component', () => {
         <DataRow
           category="category text"
           data={{
-            total: { total: 0 }
+            total: { medicaid: 0, federal: 0, state: 0, total: 0 }
```
mgwalker: Since we're not testing invalid data, make sure the data is valid.  😛  This gets rid of some warnings about the `<Dollars>` component

web/src/containers/activity/All.js
```diff
@@ -31,8 +31,8 @@ const All = ({ activityKeys, addActivity }) => (
         </Btn>
       </Subsection>
       {activityKeys.map((key, idx) => (
-        <Waypoint id={key}>
-          <EntryDetails key={key} aKey={key} num={idx + 1} />
+        <Waypoint id={key} key={key}>
```
mgwalker: Oops.  Forgot to move this key up from `<EntryDetails>` when I wrapped it in a `<Waypoint>`!

web/src/containers/admin/MyAccount.test.js
```diff
@@ -87,10 +87,7 @@ describe('my account page', () => {
       position: 'Table Cleaner'
     };
 
-    // sinon.stub().rejects() wraps strings in Error objects, but the edit
-    // account action rejects a plain string, so return a rejected promise
-    // eslint-disable-next-line prefer-promise-reject-errors
-    const editAccount = sinon.stub().returns(Promise.reject('error message'));
+    const editAccount = sinon.spy();
```
mgwalker: For a hot minute, these admin forms relied on Redux actions returning promises.  That wasn't a good design and we changed it, but this test was still built around returning promises.  The promise was rejected, and the `MyAccount` component no longer handled it.  Easy fix, though - just take out the promise!

web/src/reducers/navigation.js
```diff
@@ -1,7 +1,7 @@
 import { NAVIGATION_SCROLL_TO_WAYPOINT } from '../actions/navigation';
 
 const initialState = {
-  activeSection: null
+  activeSection: 'apd-state-profile-overview'
```
mgwalker: Default the active section of the nav bar to the state profile overview.  The `SideBar` component requires an active section be set, so...  set one.

web/src/components/PasswordWithMeter.test.js
hbillings: I can't even tell you how much I love preemptive technical debt resolution. It's so beautiful :sob:

web/src/components/PasswordWithMeter.test.js
hbillings: Ahh, that's my bad!

web/src/reducers/navigation.js
hbillings: This raises a larger question about whether it _should_ require an active section to be set, IMO. But I guess we're only going to be using the sidebar on the builder page for now, so... maybe it's okay? (Having it hardcoded like this makes me a little uncomfortable, even though I've done the same thing elsewhere in the app.)

web/src/components/PasswordWithMeter.test.js
mgwalker: In fairness, it made test output ugly because Node warned about it, and I was tired of ugly test output. 😆

web/src/reducers/navigation.js
mgwalker: We could also set a `defaultProp` on the sidebar itself.  Then at least this hard-coded value would be in the same place as its hard-coded counterparts?

PR #1473, opened 2019-04-17
Adds export section

### LINKED ISSUE #1449: Create the Export section
Implements the design described in #1321, closes both 1321 and #1256.

Here's the original mockup, which will need an update after recent changes to the Save experience and the full design system implementation.

App - 9: Export

The language is in #1256, not this mockup. If it's helpful I can also pull that over here, but I'm not worried about it because the whole point is that the content part is easy to edit in YAML later!

### This task is done when...
 - [ ] The interface direction is aligned with the rest of the app.
 - [ ] The appropriate code changes have been made and the right place for the content to live in YAML is identified.
 - [ ] The content is up to date and appropriate.
 - [ ] Product has signed off on the experience.

mgwalker:
- removes the "save pdf" button from the sidebar
- adds an "Export and Submit" section
  - ~needs email address - does this need to be a separate issue, or do we know that mailbox address now?~ done

56317297-01e3cf00-6122-11e9-86cf-4f06af442632

- closes #1449


web/src/containers/ApdExport.js
```diff
@@ -0,0 +1,83 @@
...
+
+import Waypoint from './ConnectedWaypoint';
+import { printApd } from '../actions/print';
+import { Section } from '../components/Section';
+import { FileDownload } from '../components/Icons';
+
+const Bold = ({ children }) => (
+  <span className="ds-u-font-weight--bold">{children}</span>
+);
+Bold.propTypes = { children: PropTypes.node.isRequired };
```
mgwalker: Made these as helpers rather than write out those long class names over and over. 😬

web/src/containers/ApdExport.js
```diff
@@ -0,0 +1,82 @@
...
+
+const Italic = ({ children }) => (
+  <span className="ds-u-font-style--italic">{children}</span>
+);
+Italic.propTypes = { children: PropTypes.node.isRequired };
+
+const ExportAndSubmit = ({ printApd: print }) => (
+  <Waypoint id="export-and-submit">
+    <Section isNumbered id="export-and-submit" resource="exportAndSubmit">
+      <p>
```
mgwalker: @jameshupp I put the content in the code instead of the YAML because the YAML's Markdown converter thing does lists really poorly.  In order to get proper numbered lists, I needed to put the text here directly.

jeromeleecms:
Official Mailbox: CMS-EAPD@cms.hhs.gov

lauraponce:
@jeromeleecms thanks for being our resident IE tester! 😄Great catch with the IE-specific print flow.

web/src/containers/ApdExport.js
hbillings: What converter are we using, and would it make sense to consider adjusting it?

web/src/containers/ApdExport.js
```diff
@@ -0,0 +1,82 @@
...
+        To submit your APD for analyst review, export the document as a PDF,
+        then send it to CMS just like you’ve done before.
+      </p>
+      <ol>
+        <li>
+          Select <Bold>Export</Bold> to open up your browser’s print options.
```
hbillings: Y'know, we _could_ just use `strong` and `em` tags here and not bother with the helper functions (although if you'd rather use the span with the style, I really like that method of ensuring consistency!)

web/src/containers/ApdExport.js
mgwalker: We're using [markdown-it](https://www.npmjs.com/package/markdown-it) and I'm totally open to switching to something else.

web/src/containers/ApdExport.js
mgwalker: Back in the old days, it'd have been `<b>` and `<i>` but it got drilled into my head not to use those.  Now `<strong>` and `<em>` are okay I guess?  But my brain still tells me never to use things like that.  😛

But if I switched to `<strong>` and `<em>`, we could get consistency by targeting those with CSS and setting the properties the same as the design system utility classes, right?

PR #1484, opened 2019-04-18
Implement new tables for program summary federal FFY share

hbillings:
Implements the second set of tables in #1394. 

Looks like:
Screen Shot 2019-04-18 at 5 14 06 PM

### This pull request changes...
- Breaks apart fiscal years and totals

### This pull request is ready to merge when...
- [ ] Tests have been updated (and all tests are passing)
- [ ] This code has been reviewed by someone other than the original author
- [ ] The experience passes a basic manual accessibility audit (keyboard nav, screenreader, text scaling) OR an exemption is documented

### This feature is done when...
- [ ] Design has approved the experience
- [ ] Product has approved the experience

Issue #1394, opened 2019-03-25
Implement new styles and layout for tables


lauraponce:
Carried forward from #1160 to update the tables in section 6.

### This task is done when...
 - [ ] The code implemented matches the design mockups, using the CMS Design System and its components
 
#### TABLE 1 - Summary Budget Table

<img width="544" alt="Screen Shot 2019-03-25 at 1 55 26 PM" src="https://user-images.githubusercontent.com/21222704/54953531-b284f700-4f05-11e9-8101-8a73a6617876.png">
<img width="539" alt="Screen Shot 2019-03-25 at 1 56 21 PM" src="https://user-images.githubusercontent.com/21222704/54953588-d6e0d380-4f05-11e9-81db-6a3ba2cb2bdb.png">
<img width="536" alt="Screen Shot 2019-03-25 at 1 56 28 PM" src="https://user-images.githubusercontent.com/21222704/54953599-dc3e1e00-4f05-11e9-8e28-b981674453f1.png">
<img width="550" alt="Screen Shot 2019-03-25 at 1 56 12 PM" src="https://user-images.githubusercontent.com/21222704/54953608-e233ff00-4f05-11e9-85e2-ed75ef3310ce.png">

#### TABLE 2 - Federal Share by FFY

<img width="538" alt="Screen Shot 2019-03-25 at 1 58 53 PM" src="https://user-images.githubusercontent.com/21222704/54953746-33dc8980-4f06-11e9-86de-e7089bf8d0b2.png">
<img width="537" alt="Screen Shot 2019-03-25 at 1 59 09 PM" src="https://user-images.githubusercontent.com/21222704/54953750-3808a700-4f06-11e9-9868-5844f19c5a01.png">

### TABLE 3 - Incentive Payments by FFY Quarter

<img width="534" alt="Screen Shot 2019-03-25 at 1 59 43 PM" src="https://user-images.githubusercontent.com/21222704/54953916-99307a80-4f06-11e9-9328-63d0f8f45e6d.png">

lauraponce:
### Implementation styling notes

- `HIT Activities (and other table headers)` use H4 visual styling
- Text inside tables is all font size 12

Issue #1469, opened 2019-04-16
Audit the table headers


hbillings:
With all of the table rejiggering that's happened lately, we should do a full-page audit of the table headers to make sure they're coherent and consistent.

### This task is done when...
 - [ ] all of the tables have been audited and any new errors have been corrected
 - [ ] the header cells are associated with the correct data
 - [ ] screen readers read tables coherently
 - [ ] any content that is currently in what should be empty cells has been removed (see #1480 for examples)
 - [ ] Instead of `subtotal`, use `total` across the board for tables that are now broken apart by FFY

hbillings:
This is mostly obsoleted by other tickets that have been opened.

Issue #1459, opened 2019-04-15
Content Review of Table Headers


NAretakis:
With #1394 being implement, this gave us a major leap forward in viewing and understanding tables. Now that the tables are understandable, this issue is to review the content for each of the table headers (screenshot), and make sure the headers match the content we are actually looking for. 

![image.png](https://images.zenhubusercontent.com/5c536083d11206601037eb29/d900352c-052a-4d55-9101-0ef55e0125d4)

This issue is should not be started until all the tables have been cleaned up in the way #1394 cleaned up the Summary Budget Tables.

### This task is done when...
 - [ ] Content in tables and table headers is reviewed
 - [ ] Language changes made to table headers where needed.

PR #1485, opened 2019-04-18
Implement new tables for incentive payments

hbillings:
The last TWO sets of tables for #1394. (Closed #1484 in favor of this one, since all of that code is also in here, whups.)

- Removes some bold text
- changes headers from being specified to being `scope="row"` and `scope="col"`


web/src/containers/IncentivePayments.js
```diff
@@ -45,32 +43,32 @@ class IncentivePayments extends Component {
                   <th id={thId('null2')} />
                   <Fragment key={year}>
                     {QUARTERS.map(q => (
-                      <th key={q} className="right-align" id={thId(year, q)}>
+                      <th key={q} className="right-align" id={thId(year, q)} scope="col">
```
mgwalker: Since we're not using the `headers` attr anymore, can we also get rid of the `id`s on these header cells?

web/src/containers/QuarterlyBudgetSummary.js
```diff
@@ -7,150 +7,130 @@ import { t } from '../i18n';
...
-                        id={`quarterly-budget-summary-${source}-fy-${year}`}
-                      >
-                        {t('ffy', { year })}
-                      </th>
-                    ))}
+            <h3 className="ds-h3">{sourceDisplay}</h3>
+            {years.map((year) => (
+            <table className="table-cms">
+              <colgroup>
+                <col className="table-cms--col-header__fixed-width" />
```
mgwalker: Whoa, what is this witchcraft?

mind blown

web/src/containers/QuarterlyBudgetSummary.js
```diff
@@ -7,150 +7,130 @@ import { t } from '../i18n';
...
+                <col />
+                <col />
+                <col />
+                <col />
```
mgwalker: Could these be collapsed to `<col span="4" />`?

web/src/containers/QuarterlyBudgetSummary.js
```diff
@@ -7,150 +7,130 @@ import { t } from '../i18n';
...
+              </colgroup>
+              <thead>
+                <tr>
+                  <th
+                    key={year}
```
mgwalker: Don't need this `key` anymore.  I think it should be on the `<table>` up on line 31 instead?

web/src/containers/QuarterlyBudgetSummary.js
```diff
@@ -7,150 +7,130 @@ import { t } from '../i18n';
...
-                      id={`quarterly-budget-summary-${source}-total2`}
-                    />
-                  </tr>
-                </thead>
-                <tbody>
-                  {Object.keys(EXPENSE_NAME_DISPLAY).map(name => (
-                    <tr
-                      key={name}
-                      className={`${name === 'combined' ? 'bold' : ''}`}
+                      headers={`quarterly-budget-summary-${source}-null1 quarterly-budget-summary-${source}-null2`}
```
mgwalker: Does `headers` apply to `th` elements?  This may fall more under the purview of #1459, not sure.

web/src/containers/QuarterlyBudgetSummary.js
```diff
@@ -7,150 +7,130 @@ import { t } from '../i18n';
...
-                            <Dollars>{data[year].subtotal[name]}</Dollars>
-                          </td>
-                        </Fragment>
-                      ))}
-                      <td
-                        className="bold mono right-align nowrap bg-gray-light"
-                        headers={`quarterly-budget-summary-${source}-total2 quarterly-budget-summary-${source}-total`}
+                        className={`mono right-align nowrap ${
+                          name === 'combined'
+                        }`}
```
mgwalker: Two questions here.  One, is there a plan to come back and change up the classes on this one?  Looks like it's still using the old atomic styles.

Two, it looks like the class will either be `mono right-align nowrap` if the name is "combined" or else `mono right-align nowrap false`?  Is that intended?

scratch chin

lauraponce:
Swapped in @austinhernandez for the design review on this one since he's been 👀at table layouts

web/src/containers/QuarterlyBudgetSummary.js
hbillings: Writing ancient HTML is like writing runes, man. (You _could_ just put a class on the actual header, but since it's in a loop, I kind of feel like writing it out as a colgroup makes it a little more readable? Or maybe that's just because I, like HTML, am ancient?)

web/src/containers/QuarterlyBudgetSummary.js
hbillings: I DID NOT KNOW THIS WAS POSSIBLE

web/src/containers/QuarterlyBudgetSummary.js
hbillings: I just now ripped these out and did my best approximation of the structure #1459 will move us towards. 🤷‍♀️

web/src/containers/QuarterlyBudgetSummary.js
hbillings: Yeah, I figured we'd clean up those classes when we know for sure what we want the tables to look like. Good catch on the `combined` one -- that was me apparently not cleaning up _all_ of a ternary operator. 😬

hbillings:
@jeromeleecms I added a note to #1469 about changing `Subtotal` to `Total`. Seemed to make sense there because then I can test the full screen reader flow of things!

PR #1487, opened 2019-04-22
Implement new styles and layout for state dashboard

### LINKED ISSUE #1378: Implement new styles and layout for Layer 1 state dashboard/landing page

### LINKED ISSUE #1389: Add unique ID to the state dashboard
Created from #1346. Depends on #1378 

So that users can have a unique identifier for keeping track of their APD, add the unique ID in the dashboard. For example:

image

### This task is done when...
 - [ ] Unique ID has been added to dashboard

mgwalker:
- uses the standard header and big footer
- implements the new dashboard style from #1378
- displays the APD name
- displays the APD last update time
- closes #1378 
- closes #1389


web/package.json
```diff
@@ -11,7 +11,7 @@
     "build": "npm run scaffold && NODE_ENV=production webpack -p",
     "lint": "eslint 'src/**/*.js'",
     "start": "webpack-dev-server --config webpack.config.dev.js",
-    "test": "jest --collectCoverage",
+    "test": "TZ=utc jest --collectCoverage",
```
mgwalker: Set the timezone so the human-readable thingy will be consistent when running tests, regardless of the test machine's actual timezone.

web/src/containers/StateDashboard.js
```diff
@@ -33,63 +33,75 @@ const StateDashboard = (
...
-                      className="right"
-                      onClick={delApd(apd)}
-                    >
-                      Delete
-                    </Button>
-                  </div>
-                ))}
+                  <span
+                    className="ds-u-fill--primary-alt-lightest ds-u-padding--2"
+                    style={{ marginLeft: '-16px' }}
```
mgwalker: Unclear to me why this was necessary, but this is the width of the icon and this pulls the span back so that it's actually centered. I don't know why it isn't centered anyway because the containing `div` has a `ds-u-text-align--center` and :all-the-shruggies:

web/src/i18n/locales/en/stateDashboard.yaml
```diff
@@ -1,13 +1,14 @@
 title: My State Dashboard
 subheader:
-helpText: >-
-  The CMS HITECH APD app is the new way to create and manage your state’s HITECH
-  Advanced Planning Documents (APDs). Follow the steps to fill in the required information,
-  and the app will generate a completed APD you can submit to a CMS State Officer
-  for review.
+instruction:
```
mgwalker: @jameshupp @austinhernandez Changed this to an instruction detail to match the style in #1378.  Is that correct, or should this still be help text, inside the little blue box and everything?

web/src/containers/StateDashboard.js
hbillings: I'm working on a patch PR for ya (because explaining "take this class from here and put it here" just doesn't work so well with the commit suggestions). This is a good question, though! The columns in nested rows are smaller than the columns in their parents, which is an eternal pain point between design and front end dev. If you look at the original screenshot, you'll see that this icon is supposed to be "one column" wide. Except now we're inside a container inside a container, so this row makes 12 columns out of what used to be 8. (You're accounting for this numerically already.) One column is now too small to actually hold this icon at size, so it squeezes it too far to the right. Because it's a `span`, it'll overflow the column and do its own thing. Here's a screenshot that makes it easier to see what's happening, with `span` set to `display: block`:
Screen Shot 2019-05-17 at 10 39 30 AM

CSS is the worst.

hbillings: Oh, this is also a prime example of why a designer may want to make a static visual element a column wide, but implementing it so that it actually is a column width gets messy :-/

web/src/i18n/locales/en/stateDashboard.yaml
jameshupp: No, if this page is styled according to the same rules as the actual app, this is right.

PR #1488, opened 2019-04-24
Dashboard update

### LINKED ISSUE #1028: Focused Review of the Language within the Dashboard
Review the Dashboard language and statuses for provide clarity + evaluate content from the Round 7 (and subsequent) usability testing.

### This task is done when...
 - [ ] The dashboard introduces _what_ the app is, not just how to use it, per testing feedback.
 - [ ] The dashboard fulfills the app's design principles.
 - [ ] Product has reviewed any changes.

jameshupp:
### This pull request changes
* closes #1028 
* changes the language on the dashboard to reflect a desire for a description of what the app is
* simplifies the language about what to do with the app

### This pull request is ready to merge when
- [ ] The language passes a copy review
- [x] The dashboard reflects line breaks in the YAML

### This feature is done when
- [x] Product has approved that the language serves the org and user needs
- [ ] Design has approved that the language serves the desired experience


mgwalker:
I think we should merge this one before #1487 and then port the changes over.  Does that make sense to folks?  #1487 implements content changes based on the mockups, but it makes more sense to me to hold until this one is approved and just use the new content from the start.  🙂

jameshupp:
Seems right to me! One other note for product reviewers pulled from a commit message (it's very annoying that I write the details there and then forget to put them in the pull request):

I wanted to clarify that this app is new (to help suggest that it replaces the long-hand written version), that it makes requirements clear, and that it generates the completed doc. The reason is that I want to be really clear with folks what the advantage is of using the new app vs the old way. Ideally it's obvious because the app is so good! But maybe it's not yet and will only get there once we've got the second and third layers done. So that's the first paragraph.

The second paragraph is about what’s on the homepage/dashboard. I didn’t call it a homepage or dashboard, though, since it doesn’t feel like _obviously_ either of those things until there’s more capability. What this page is should speak for itself while it’s still simple.

One other assumption that goes into all this: we can always change the language. The homepage/dashboard of a focused app like this is a good place to give people updates as things change. We can just keep forefronting what they need next. Right now they need to know what this app _is_ in addition to what it _does_. In a few months or a year they'll need something different.

jameshupp:
I don't think there's anything wrong with the term in navigation or a header; eventually that's what the page will be. A page title or header can be aspirational or help people get used to a concept. It may not be perfectly accurate, but people understand what titles and headers _are_ from looking at them. It's a bit more dangerous to make reference to a word in narrative copy without doing the work of explaining why it's appropriate, if it's not obvious already why it's the right word. So for the top nav to use `homepage` and for the header on this page to use `dashboard` both feel okay, especially in a pilot environment, as we're building toward it. But if we said `All your APDs will be on this dashboard` it can trigger an expectation for more than the list we're currently giving them.

That doesn't mean we _must_ use those words. But I'm fine with them.

I added the email address. I'm not sure it adds much on this page, but I like the idea of giving people more specificity rather than less.

PR #1489, opened 2019-04-25
Add new header

### LINKED ISSUE #1468: Implement top navigation (horizontal header)

hbillings:
Closes #1468. Restyles the header consistently across pages.

- Removes the `Header` and `TopBtns` component from individual pages in favor of adding a single `Header` to `Wrapper.js`
- Moves the actions formerly in buttons to a dropdown menu
- Displays `CMS HITECH APD` when a user is at a top-level page and either `Admin dashboard` or `[STATE] APD home` when the user is at a point that he can navigate back "home"

WIP until these things are done:
- [x] Stop `aria-expanded` state from persisting, possibly by making the dropdown open on mouseover instead of on click (@austinhernandez, I'd like your input on this)
- [x] Tests


hbillings:
@austinhernandez This is still WIP while I fix bugs/write tests, but other than the bug I documented in the description, I think this is ready for you to review.

web/src/components/Wrapper.js
```diff
@@ -17,7 +18,12 @@ class Wrapper extends Component {
 
   render() {
     const { children } = this.props;
-    return <div className="site">{children}</div>;
+    return (
+      <div className="site">
+        <Header key={location.pathname}/>
```
hbillings: @mgwalker I'm not sure that putting the header in `Wrapper.js` makes sense, but I couldn't figure out where else to put it so that every page would pick it up. I'm open to other ideas!

web/src/components/Header.js
```diff
@@ -46,19 +101,29 @@ class Header extends Component {
...
+Header.defaultProps = {
+  currentUser: null,
+  ariaExpanded: false
+};
 
-export default connect(
-  mapStateToProps,
-  mapDispatchToProps
-)(Header);
+export default withRouter(
```
hbillings: I'm using `withRouter` here to give the component access to `location.pathname`, but after looking at `Login.js`, I'm not sure I need to do this?

mgwalker: You need `withRouter` if your component is not in a `<Route>`.  The `<Route>` component will provided `location` and `history` props to the thing it renders, so that's why `Login.js` doesn't need it, but `Header.js` probably does.

web/src/components/Header.js
```diff
@@ -3,62 +3,134 @@ import Button from '@cmsgov/design-system-core/dist/components/Button/Button';
...
+
   render() {
-    const { authenticated } = this.props;
+    const { authenticated, currentUser, isAdmin, location } = this.props;
+    const { ariaExpanded } = this.state;
+    const username = currentUser ? currentUser.username : 'Your account';
+    const isTopLevel = location.pathname == '/';
     return (
-      <header className="header__slim">
+      <OutsideClickHandler
```
hbillings: This is the Airbnb handler for this. Looking at the source, it seemed like it did exactly what I was going to write from scratch anyway. https://github.com/airbnb/react-outside-click-handler

mgwalker: This is my only concern with the Airbnb library:  https://github.com/airbnb/react-outside-click-handler/issues/24

hbillings: Hrm. Good catch. I suppose I could just write my own!

web/src/components/Header.js
```diff
@@ -1,42 +1,110 @@
...
+                <ul className="nav--dropdown" aria-expanded={ariaExpanded}>
+                  <li>
+                    <Button
+                      size="small"
+                      variation="transparent"
+                      className="nav--dropdown__trigger"
+                      onClick={this.toggleDropdown}
+                    >
+                      { currentUser ? currentUser.username : 'Your account' }
+                      <Icon icon={faChevronDown} style={{ width: '8px'}} />
```
hbillings: Not a huge fan of this pattern, but it's the only way I can find to reliably set the SVG width. Setting it in the CSS and not inline on the element seems to get ignored. In summary, SVGs are black magic.

web/src/components/Header.js
```diff
@@ -6,37 +8,103 @@ import { connect } from 'react-redux';
...
+  componentDidMount() {
+    document.addEventListener('mousedown', this.handleOutsideClick);
+  }
+
+  handleOutsideClick = e => {
+    if (this.node.contains(e.target)) {
+      return;
+    }
+    this.setState({ ariaExpanded: false })
+  };
```
hbillings: This method and the `document.addEventListener` in `componentDidMount` seem like they ought to be tested, but I can't seem to figure out the right way to stub things and I don't want to hold up this PR any longer while I bang my head against the wall.

web/src/components/Header.js
```diff
@@ -6,37 +8,102 @@ import { connect } from 'react-redux';
...
+    this.setState(prev => ({ ariaExpanded: !prev.ariaExpanded }));
+  };
+
   render() {
-    const { authenticated } = this.props;
+    const { authenticated, currentUser, isAdmin, showSiteTitle } = this.props;
+    const { ariaExpanded } = this.state;
     return (
-      <header className="header__slim">
+      <header ref={node => this.node = node}>
```
hbillings: This is some black magic right here. But it's what makes the `outsideClickHandler` work, so I'll take it.

web/src/components/Header.js
```diff
@@ -6,37 +8,102 @@ import { connect } from 'react-redux';
...
-                  className="action--logout"
-                  variation="transparent"
-                  onClick={this.handleLogout}
-                >
-                  {t('logout')}
-                </Button>
+              <div className="ds-l-col--12 ds-l-md-col--8">
+                <ul className="nav--dropdown" aria-expanded={ariaExpanded}>
+                  <li>
+                    <button
```
hbillings: Made this a regular button instead of a Button element because testing a click event on a child element is a den of vipers.

web/src/components/Header.test.js
```diff
@@ -0,0 +1,117 @@
...
+    component.find('.nav--dropdown__trigger').simulate('click');
+    expect(component).toMatchSnapshot();
+    expect(component.state().ariaExpanded).toBeTruthy();
+
+    component.find('.nav--dropdown__trigger').simulate('click');
+    expect(component.state().ariaExpanded).toBeFalsy();
+  });
+
+  it('renders the admin home title when an admin user is at a secondary page', () => {
+    const component = mount(
```
hbillings: In order to get the link text of the `Link` element, I had to mount the whole dang thing. Which required `MemoryRouter`, for some reason I don't pretend to understand.

web/src/components/Header.test.js
```diff
@@ -0,0 +1,117 @@
...
+        }
+      },
+      user: {
+        data: {
+          role: 'admin'
+        }
+      }
+    };
+
+    expect(mapStateToProps(state)).toEqual({ authenticated: 'some value', currentUser: { user: { role: 'admin' }}, isAdmin: true});
```
hbillings: This test is failing and I can't figure out why. It seems to be expecting a data structure that is different than what's being passed in, but I can't actually tell from looking at [the error on Circle](https://circleci.com/gh/18F/cms-hitech-apd/28928?utm_campaign=vcs-integration-link&utm_medium=referral&utm_source=github-build-link) what it needs.

web/src/components/Wrapper.js
```diff
@@ -17,7 +20,13 @@ class Wrapper extends Component {
 
   render() {
     const { children } = this.props;
-    return <div className="site">{children}</div>;
+    let showSiteTitle = location.pathname === '/'
+    return (
+      <div className="site">
+        <Header key={location.pathname} showSiteTitle={showSiteTitle} />
```
hbillings: I'm not sure this should go in `Wrapper`, but 1) I could avoid using `withRoute` in the `Header` component this way, 2) I wasn't sure if there were another globally used component.

web/src/containers/Login.js
```diff
@@ -7,7 +7,6 @@ import { TextField } from '@cmsgov/design-system-core';
 import ConsentBanner from '../components/ConsentBanner';
 import { login } from '../actions/auth';
 import CardForm from '../components/CardForm';
-import Header from '../components/Header';
```
hbillings: Yay, we don't need to import this on every page any more!

web/src/styles/scss/header.scss
```diff
@@ -1,21 +1,115 @@
...
-  padding: $spacer-1 0;
-  .site-title {
-    display: block;
-    margin: $spacer-1 0;
-    text-decoration: none;
-    font-weight: bold;
+header {
+  background-color: $color-primary-darkest;
+  padding: $spacer-2 0;
+  letter-spacing: 0.6px;
```
hbillings: @austinhernandez I kept the original letter spacing here since we haven't moved to the new font family yet. If you'd rather I nix it, let me know!

web/src/styles/scss/header.scss
```diff
@@ -1,21 +1,115 @@
...
   }
+}
 
-  .action--logout {
-    float: right;
-    margin-top: $spacer-1;
-    color: $color-white;
+.site-title {
+  svg {
+    margin-right: $spacer-1;
```
hbillings: Applies to the back carat only.

web/src/styles/scss/header.scss
```diff
@@ -1,21 +1,115 @@
...
+}
+
+.nav--dropdown {
+  padding-left: 0;
+  margin-bottom: 0;
+  float: right;
+  text-align: right;
+  position: relative; /* needed for submenu absolute positioning */
+  min-width: 200px; /* don't render submenu oddly if email is very short */
+  &[aria-expanded='true'] {
```
hbillings: By keying off of the `aria` attributes, we ensure they're present.

web/src/styles/scss/header.scss
```diff
@@ -1,21 +1,115 @@
...
+    .nav--dropdown__trigger {
+      border-bottom: 2px solid $color-primary;
+    }
+    .fa-chevron-down {
+      transform: rotate(180deg);
+    }
+  }
+  .fa-chevron-down {
+    transition-timing-function: ease-in;
+    transition: 0.2s;
```
hbillings: smooth

web/src/components/Header.js
mgwalker: i got you

### For clicking directly on the toggly bit

In the `opens and closes the dropdown when clicked` test, add this after you create the component:
```js
jest.spyOn(document, 'addEventListener');
jest.spyOn(document, 'removeEventListener');
```

Then after the first click simulation, add these assertions:
```js
expect(document.addEventListener).toHaveBeenCalled();
expect(document.addEventListener.mock.calls[0][0]).toEqual('click');
const handler = document.addEventListener.mock.calls[0][1];
```

And after the second click simulation, these assertions:
```js
expect(document.removeEventListener).toHaveBeenCalled();
expect(document.removeEventListener.mock.calls[0][0]).toEqual('click');
expect(document.removeEventListener.mock.calls[0][1]).toEqual(handler);
```

### For testing dom events

```js
it('handles outside click events', () => {
  // Create an instance of the component object but don't mount it. This
  // breaks us out of all the React requirements, which is handy because
  // we don't actually need them for these tests.
  const component = new Header({
    authenticated: true,
    currentUser: {
      role: 'admin',
      state: { id: 'wa', name: 'Washington' },
      username: 'frasiercrane@kacl.com'
    },
    isAdmin: true,
    pushRoute: () => {},
    ariaExpanded: false,
    showSiteTitle: true
  });

  // React sets the node for us, so since we're not Reacting, do it manually.
  // In the first case, make the contains fn return false.
  component.node = { current: { contains: jest.fn(() => false) } };

  jest.spyOn(document, 'removeEventListener');
  // By default, spyOn calls the real method as well, but that'll cause a
  // React error here because the component isn't mounted
  jest.spyOn(component, 'setState').mockImplementation(() => {});

  // Call the method under test
  component.handleOutsideClick({});

  // Make sure the state was set and the event handler was removed.
  expect(component.setState).toHaveBeenCalled();
  expect(component.setState.mock.calls[0][0]).toEqual({
    ariaExpanded: false
  });
  expect(document.removeEventListener).toHaveBeenCalled();
  expect(document.removeEventListener.mock.calls[0][0]).toEqual('click');
  // we can't check the function because of weird binding

  // Reset the mocks
  document.removeEventListener.mockClear();
  component.setState.mockClear();

  // Change the node contains function to return true this time.
  component.node.current.contains = jest.fn(() => true);

  component.handleOutsideClick({});

  expect(component.setState).not.toHaveBeenCalled();
  expect(document.removeEventListener).not.toHaveBeenCalled();
});
```

web/src/components/Header.js
```diff
@@ -1,3 +1,5 @@
+/* eslint-disable no-return-assign */
```
mgwalker: One last comment since I forgot to add this one.  Do we need this exception still?  If so, can you move it next to the spot where the issue is and change it to `// eslint-disable-next-line`?

web/src/components/Header.js
mgwalker: Is this issue here that there are two `Button`s and it's hard to get Enzyme to give you just the one you want?  Because if so...  yep.  😆

web/src/components/Header.js
hbillings: I did try that! But for some reason, the icons were all slightly different sizes and it looked _weird_. no idea

web/src/components/Header.test.js
hbillings: well duh

web/src/components/Icons.js
```diff
@@ -28,13 +30,15 @@ export {
...
   faChevronDown,
+  faChevronLeft,
   faChevronUp,
   faCircle,
   faClock,
   faEdit,
   faExclamationTriangle,
   faLock,
   faPlusCircle,
+  faSignOutAlt,
```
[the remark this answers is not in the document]
hbillings: Oooh, good catch, thank you!

web/src/components/Header.js
hbillings: We don't need this any more! Yay!

web/src/components/Wrapper.js
```diff
@@ -1,5 +1,8 @@
+/* eslint-disable no-restricted-globals */
```
mgwalker: What do you think about using `document.location` instead of just `location`?  Then we can get rid of this eslint exception.

web/src/components/Wrapper.js
```diff
@@ -17,7 +20,13 @@ class Wrapper extends Component {
 
   render() {
     const { children } = this.props;
-    return <div className="site">{children}</div>;
+    let showSiteTitle = location.pathname === '/'; // eslint-disable-line prefer-const
```
mgwalker: Can this go ahead and be a `const`?

web/src/components/Wrapper.js
```diff
@@ -17,7 +20,13 @@ class Wrapper extends Component {
...
+    return (
+      <div className="site">
+        <Header key={location.pathname} showSiteTitle={showSiteTitle} />
```
mgwalker: I'm not sure what the `key` prop is for here.  It's not listed in the propTypes for the Header.

web/src/components/Wrapper.js
hbillings: I ran into some issues with it being a `const` because the wrapper doesn't always rerender itself... or something? It was failing to reassign the value when I navigated between pages, so it was just always equal to `/`. If there's a different way to approach it, I'm all ears!

mgwalker: Ahhhhhh.  I don't know why using a `let` would fix that, but the problem makes sense.  If none of the wrapper's props change, it won't re-render.  If you use `withRouter()`, the component should re-render whenever the location changes because the location props are passed in.

web/src/components/Wrapper.js
hbillings: Oh, whoops! I'll add it. The key forces the `Header` to rerender when `location.pathname` changes; otherwise it doesn't consistently update itself on navigation. (Mostly I was using this to force `ariaExpanded` to reset itself when the user navigated to a page from the dropdown menu... maybe I could just add a handler to those items... 🤔 )

web/src/components/Wrapper.js
hbillings: 👯 I didn't realize `document.location` was accessible from here! 👯

austinhernandez:
Heading looks sharp! Good on my end

web/src/components/Header.js
```diff
@@ -6,37 +6,111 @@ import { connect } from 'react-redux';
...
+              }
+              {!showSiteTitle && !authenticated &&
+                <Link to="/">
+                  {t('titleBasic')}
+                </Link>
+              }
+              {!showSiteTitle && authenticated && (
+                <Link to="/">
+                  <Icon icon={faChevronLeft} size="sm" />
+                  { isAdmin ? "Admin dashboard" : `${currentUser.state.id.toUpperCase()} APD home` }
```
jeromeleecms: Small thing, but can we title case this to "APD Home" vs "APD home" to match the design sketch?

web/src/components/Header.js
oliviertassinari: Did you know https://next.material-ui.com/components/click-away-listener/?

mgwalker: Hi @oliviertassinari! We're actually in the process of removing Material-UI from the project. Thanks for the info, though!

oliviertassinari:
Thanks for sharing this info. What's your biggest pain point with it?

mgwalker:
@oliviertassinari No pain point, it's just not what we need.

PR #1490, opened 2019-04-29
Add floating "save" button, error behavior, and button state

### LINKED ISSUE #1467: Implement save button, interactions, and error messages
We are going to start implementing #1308 !

The Save state interaction styles are as follows:

Unsaved changes 
Screen Shot 2019-04-11 at 1 41 47 PM

Save failed - the error message persists and doesn't disappear until they try saving again
Screen Shot 2019-04-11 at 1 39 02 PM

Save success (what they see immediately)
Screen Shot 2019-04-11 at 1 41 22 PM

All changes saved (no new changes made) 5 seconds later after the toast popup disappears. (tentatively we say after 5 seconds, but we are going to investigate further interactions once everything is implemented and final error messaging is in)

Screen Shot 2019-04-11 at 1 42 57 PM

Once a change is made, the button turns into the Save changes button again

Screen Shot 2019-04-11 at 1 41 47 PM

While content is being saved (Save loading state)
([loading spinner](https://design.cms.gov/components/spinner/)):
Save interactions - new placement - save loading

### This task is done when...
 - [ ] Button states are implemented as referenced above
 - [ ] We are able to see the button state interactions in situ, and the design is ready to be reviewed by a designer (so we can choose the best interactions for the error / toast messages)

mgwalker:
- adds a floating "save" button to the APD form
  - button is green and says "Saved" when there are no changes to save
    green "saved" button
  - button is blue and says "Save" when the form can be saved
    blue "save" button
  - button is blue with a spinner and says "Saving" while the form is being saved
    blue "saving" button with spinner
- adds alerts above the button as appropriate
  - red error alert if there is an error saving the form
    red error alert
  - green success alert if the form saves successfully; disappears after 5 seconds
    green success alert
- closes #1467


web/src/actions/apd.js
```diff
@@ -154,19 +154,18 @@ export const notifyNetError = (action, error) => {
...
     dirty
   } = state();
 
   if (!dirty.dirty) {
-    dispatch(notify('Save successful!'));
     return Promise.resolve();
   }
 
+  dispatch(requestSave());
+
```
mgwalker: Switched this up so it only sends the save request action if it actually does the network request.

web/src/actions/apd.js
```diff
@@ -221,12 +220,10 @@ export const saveApd = ({ serialize = toAPI } = {}) => (dispatch, state) => {
   return axios
     .put(`/apds/${updatedApd.id}`, apd)
     .then(res => {
-      dispatch(notify('Save successful!'));
       dispatch(saveSuccess(res.data));
       return res.data;
     })
     .catch(error => {
-      dispatch(notifyNetError('Save', error));
```
mgwalker: go away gif

web/src/containers/SaveButton.js
```diff
@@ -0,0 +1,126 @@
...
+    }
+
+    const buttonVariant = dirty ? 'primary' : 'success';
+
+    return (
+      <div className="ds-l-row save-button--container">
+        <div className="ds-l-col--3">
+          <div className="ds-u-margin-bottom--3">
+            <div className="save-button--alert">{alert}</div>
+            <Button variation={buttonVariant} onClick={this.save}>
```
mgwalker: So my thinking was if I put this in a row, then I could have an offset column pushed to the right and that'd align things, but it turned out that didn't work like I thought.  Once the button's parent element was `fixed`, it no longer obeyed the alignment of the column.

If I'm completely honest, I don't know which of these various `<div>`s are actually necessary at this point.  The `ds-l-col--3` was meant to constrain the alerts so they didn't just grow forever, but that didn't work so I set the width directly - and now I'm not sure if this column class is doing anything.

web/src/styles/scss/save-button.scss
```diff
@@ -0,0 +1,17 @@
+@import '@cmsgov/design-system-support/src/settings/index';
+
+.save-button--container .ds-l-col--3 {
+  margin-left: 601px;
```
mgwalker: This is based on the 1040px width of the page at its biggest, which is not a great approach but I wasn't sure what else to do.  I guess it could have been percentage based.  I tried putting `ds-u-margin-left--auto` on this one and it kinda/sorta did what I wanted, but not quite.  This got closer.

web/src/styles/scss/save-button.scss
```diff
@@ -0,0 +1,17 @@
...
+  text-align: right;
+}
+
+.save-button--container .ds-l-col--3 > div {
+  bottom: 0;
+  position: fixed;
+  width: 375px;
+  z-index: 10;
```
mgwalker: Sets a width to make sure the alerts don't grow too big horizontally.  Sets a z-index so the alerts will float above textareas.

mgwalker:
@NAretakis I wasn't sure where to put the button horizontally so I tried to put it in the rightmost column of the grid (and only somewhat successfully).  My concern was that putting it outside the grid would cause problems on narrower screens.

I'll need to pair with @hbillings to fix the grid positioning it has now, so if we should move it further to the right, that'd be a great time to work that in.

web/src/reducers/errors.test.js
```diff
@@ -0,0 +1,152 @@
...
+      editOwnAccount: 'error message',
+      saveApd: 'error message'
+    };
+    [
+      ['a user account is created', ADMIN_CREATE_USER_REQUEST, 'addAccount'],
+      ['a user account is edited', ADMIN_EDIT_ACCOUNT_REQUEST, 'editAccount'],
+      ['a user saves their account', ADMIN_EDIT_ME_REQUEST, 'editOwnAccount'],
+      ['an APD is saved', SAVE_APD_REQUEST, 'saveApd']
+    ].forEach(([test, type, prop]) => {
+      it(test, () => {
```
hbillings: whoa, I didn't realize you could run tests in a loop like this! 0_0

web/src/styles/scss/save-button.scss
hbillings: I don't think there is a great answer; I've got it as percentage based in my PR just because that will inherit different pixel widths from `site-body` when the page is rendered at a smaller size. 🤷

web/src/containers/SaveButton.test.js
```diff
@@ -0,0 +1,104 @@
...
+    expect(props.saveApd.calledOnce).toEqual(true);
+  });
+
+  describe('handles changing based on behavior', () => {
+    test('shows error alert if save fails', () => {
+      // Render it in a good state, click save, then...
+      const component = shallow(<SaveButton {...props} />);
+      component.find('Button').simulate('click');
+
+      component.setProps({ error: 'Now there is an error' });
```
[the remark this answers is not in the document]
mgwalker: It's part of Enzyme

PR #1495, opened 2019-04-30
Suggestion: minor performance tweak [skip deploy]

mgwalker:
Since Github comments don't support multi-line suggestions, I'll do this one as a PR.  It modifies the header so that it only adds a global document event handler when the dropdown is expanded.

### DIFF
--- a/web/src/components/Header.js
+++ b/web/src/components/Header.js
@@ -22,15 +22,13 @@ class Header extends Component {
     this.handleOutsideClick = this.handleOutsideClick.bind(this);
   }
 
-  componentDidMount() {
-    document.addEventListener('mousedown', this.handleOutsideClick);
-  }
-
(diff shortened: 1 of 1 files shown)


web/src/components/Header.js
```diff
@@ -40,7 +38,13 @@ class Header extends Component {
   };
 
   toggleDropdown = () => {
-    this.setState(prev => ({ ariaExpanded: !prev.ariaExpanded }));
+    this.setState(prev => {
+      if (!prev.ariaExpanded) {
+        // add global click handler when the dropdown is expanded
+        document.addEventListener('click', this.handleOutsideClick);
```
hbillings: 🙌 I didn't realize I could add a global event listener outside of `componentDidMount` -- thank you!

PR #1497, opened 2019-05-01
Implement design system in program summary

### LINKED ISSUE #1442: Implement design system in section 2 (Program Summary)
The Program Summary section contains two different types of new components: checkboxes (#1444) and the rich text editor (#1443). 
 
2_Program Summary

- `Program Overview` is an H3
- `Program Introduction`, `HIT Overview`, `HIE Overview`, and `MMIS Overview` are H4s
- The help text box uses #E1F3F8 $color-primary-alt-lightest for the light blue and #205493
$color-primary-darker for the left triangle

### This task is done when...
 - [ ] The Program Summary section is using the new design system and matches the mock-ups

mgwalker:
Waiting for a little more input on h3 vs h4 in #1442, then this one will be ready to fly.

- implements the design system in the program summary section
- modifies how our yaml is parsed to add custom list support
  - lists can be either ordered or unordered:
    ```This is some paragraph text
    [ol]
    ordered list item 1

    ordered list item 2

    ordered list item 3
    [/ol]

    [ul]
    unordered list item 1

    unordered list item 2

    unordered list item 3
    [/ul]
    ```
  - supports multiple lists within the same content block
- closes #1442


web/src/components/Md.js
```diff
@@ -3,12 +3,89 @@ import React from 'react';
...
+  //   <br>
+  //  </ul>
+  //
+  // To workaround this, we can use some custom format tokens, such as [ul],
+  // to mark where lists should be.  Then we can parse out the content, build
+  // an HTML string with no extraneous whitespace or newlines, and put that
+  // back into the content before passing it on to the Markdown library.
+  //
+  // Declare this regex locally since we will be using exec(), which relies on
+  // internal regex state.  We don't want previous calls to impact this one!
```
hbillings: `internal regex state` is the most terrifying thing I've heard in recent memory. (the comments below explaining the regexes are beautiful, however.)

mgwalker:
Oh, good catch!  I think this is a little different.  The `<Instruction>` component has a `short` property that is bolded, and in the content, "Provide a high-level introduction to the state program" part is currently a short.  The `short` is also on a separate line from the `detail` property of the content. So probably two questions:

1. In this case, should the "Provide a..." content be moved out of the `short` and into the non-bolded `detail` property?
2. Should the `short` be un-bolded globally?

lauraponce:
1. Yes.

2. I don't _think_ so. Because there might be instances where it makes sense to separate out the instructions to have one part bolded with `short` and when that's not the case, it would be part of the `detail` text. @jameshupp can you check me on that? From the looks of the designs, it seems like we're straying away from the bold instruction text for the most part, but there might still be some places where it's used. Does that line up with what you're envisioning with the content style?

lauraponce:
@mgwalker let's just go with yes on question 1 and no on 2 for this ticket, and we can always add an issue later if the global styles should change for `short` so that this PR can move forward!

mgwalker:
@lauraponce That sounds great!  I'd written you a response two hours ago and forgot to push the green button, but I guess late is better than never?  😬

PR #1499, opened 2019-05-01
Patch for #1467

hbillings:
Just a few tweaks to the div structure for #1467.

### DIFF
--- a/web/src/styles/scss/save-button.scss
+++ b/web/src/styles/scss/save-button.scss
@@ -1,17 +1,18 @@
 @import '@cmsgov/design-system-support/src/settings/index';
 
-.save-button--container .ds-l-col--3 {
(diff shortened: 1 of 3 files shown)


web/src/styles/scss/save-button.scss
```diff
@@ -1,17 +1,23 @@
 @import '@cmsgov/design-system-support/src/settings/index';
 
-.save-button--container .ds-l-col--3 {
-  margin-left: 601px;
+.save-button--container {
+  max-width: inherit;
+  width: 100%;
```
hbillings: By inheriting the max-width and setting the width to 100%/margins to auto, we get a container that's the size of the `site-body` element and we can get rid of the grid system (since it won't really work here anyway).

PR #1513, opened 2019-05-02
Language change to reflect APD Portfolio

NAretakis:
Changes the sidebar and header to reflect both the eAPD and APD Portfolio (consistent language)

- Sidebar and header language adjustment

### DIFF
--- a/web/src/components/__snapshots__/Header.test.js.snap
+++ b/web/src/components/__snapshots__/Header.test.js.snap
@@ -15,7 +15,7 @@ exports[`Header component opens and closes the dropdown when clicked 1`] = `
           replace={false}
           to="/"
         >
-          CMS HITECH APD
+          CMS HITECH eAPD
         </Link>
       </div>
       <div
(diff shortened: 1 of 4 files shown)


NAretakis:
@jameshupp just a quick content read on this minor language tweak to the header and sidebar (on the dashboard). Ready for your review (please and thank you).

jeromeleecms:
@NAretakis #1378 removes the side bar entirely. Do we need to make this change for the "APD Portfolio?"

Also, we did have a conversation about include eAPD in the header - James had some reservations about the inclusion of "e" because it was would denote that there is an alternative method to submit APDs and the future intent was that the app was "the" way to submit APDs (even though the name of the product is eAPD).  Would be interested to hear if you had any new thoughts around this @jameshupp!

NAretakis:
@jeromeleecms you're right, I think I jumped the gun on this one. Closing it out.

jameshupp:
I know this is closed out and review no longer needed, but just to complete the conversation: 

Outside the context of the app itself, `eAPD` and `APD` are fine distinctions. But once someone is an app user, they should begin to treat this as one and the same thing. What they build in this app _is_ their APD. That it has to be printed as a PDF for routing is simply a stopover point on the way to the app managing the entire process. I would advocate for saying only `APD` within the app so that people who are pilot users begin to think of them as a single thing.

PR #1514, opened 2019-05-03
Replace bcrypt with pbkdf2 [skip deploy]

### LINKED ISSUE #1506: Switch password hashing algorithm
We currently use [bcrypt](http://bcrypt.sourceforge.net/) to hash passwords to store in the database. However, bcrypt is not approved by FIPS 140-2 so we cannot use it.  We should instead switch to [pbkdf2](https://en.wikipedia.org/wiki/PBKDF2) using [SHA-256](https://en.wikipedia.org/wiki/SHA-256).

### This task is done when...
- [x] implement pbkdf2 in API
   - [x] replace all uses of bcrypt
   - [x] update database seeds
- [x] benchmark pbkdf2 in Node.js to determine a good number of iterations for...
   - [x] development - **default is fine**
   - [x] preview deployments - **330,000, to achieve about 300ms**
   - [x] testing/staging/production deployments - **375,000, to achieve about 300ms**
- [x] update deployment scripts to set iteration count based on deployment target

mgwalker:
- implements pbkdf2
- updates everywhere bcrypt was being used to use pbkdf2 instead
- updates database seeds to initialize database with pbkdf2 hashes
- removes bcrypt
- closes #1506 


api/auth/passwordHash.js
```diff
@@ -0,0 +1,160 @@
...
+ *
+ * @returns {string} A string that is the hashed password, salted, and iterated
+ *                   or throws if there is an error. Hash is 60 characters
+ *                   using the default salt generator.
+ */
+const hashSync = (
+  password,
+  { iterations = process.env.PBKDF2_ITERATIONS || DEFAULT_ITERATIONS } = {}
+) => {
+  const salt = module.exports.generateRandomSalt();
```
apburnes: Never seen this convention with the `module.exports.function` assigned to a variable.  What's the background?

mgwalker: If I did:

```js
const salt = generateRandomSalt();
```

...it would *always* use the local `generateRandomSalt()` method and it couldn't be overridden.  I needed to be able to override it in order to have deterministic tests.  The test can't reach into the module and change the local `generateRandomSalt`, but it ***can*** change the value on `module.exports.generateRandomSalt`, so that's why this is the way it is.

Here's where the test changes it:
https://github.com/18F/cms-hitech-apd/pull/1514/files#diff-07625e265089b39c6dfef07ab8dd36d7R27
```

apburnes: Ah, very clever. Nice.

PR #1517, opened 2019-05-06
Update section 7 (assurances and compliance) to use design system

### LINKED ISSUE #1447: Apply the design system to section 7 (Assurances and Compliance)
For the Assurances and Compliance section, all of the yes/no radio buttons should default to unselected with the same branching behavior to provide the explanation box if no is selected.

7_Assurances and Compliances

### This task is done when...
 - [ ] The Assurance and Compliance section is using the new design system

mgwalker:
- closes #1447

### DIFF
--- a/web/src/containers/AssurancesAndCompliance.js
+++ b/web/src/containers/AssurancesAndCompliance.js
@@ -1,3 +1,10 @@
+import {
+  Button,
+  Choice,
+  FormLabel,
+  Review,
+  TextField
+} from '@cmsgov/design-system-core';
 import PropTypes from 'prop-types';
 import React, { Component } from 'react';
 import { connect } from 'react-redux';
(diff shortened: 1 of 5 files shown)


mgwalker:
@lauraponce Yes! On a new APD, they are unselected by default. The reason they all show "yes" for you is that previously they defaulted to yes, so when the shared "Nick's APD" was created, it got that value.  I've cleared that part out of the database, so you should now have them all reset to unselected.  Let me know if that worked!

lauraponce:
@mgwalker ah gotcha! Okay, that worked, and I'm seeing it all default to unselected now. This section looks so much better now — good to go!

web/src/containers/AssurancesAndCompliance.js
```diff
@@ -72,43 +68,48 @@ class AssurancesAndCompliance extends Component {
                 {apdSections[name].map(
                   ({ title, checked, explanation }, index) => (
                     <div key={title} className="mt2">
```
hbillings: `mt2` is from the old atomic system -- can we replace with ` ds-u-margin-top--2`? (There's also an `mb3` on line 66, but GH won't let me comment on it since it wasn't changed in this PR.)

Also, can we make this `div` a `fieldset`? ([`fieldset` is the only semantic parent for `legend`, weirdly enough](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/legend).)

hbillings:
Okay so I approved because I was excited, but I do have some changes to request and GH will only let me dismiss my review, not edit it. WHATEVER GITHUB

web/src/containers/AssurancesAndCompliance.js
mgwalker: Oh, I guess.

PR #1518, opened 2019-05-06
Update section 8 (executive summary), activity summary with design system

### LINKED ISSUE #1491: Implement the Executive Summary section
From the design ticket #1457 we are ready to implement!

Final mockup with text updates from @jameshupp and product approval!

image

### This task is done when...
 - [ ] The Executive Summary section looks like the mockup
 - [ ] Use the [review pattern](https://design.cms.gov/patterns/review/) from the CMS design system 
 - [ ] Designers approve! 
 
Details:
- This design was created 4/22 as part of our coworking session.
- This mockup has updated copy - just one line "Total Funding Request" might be subject to change pending @jameshupp approval.
- Don't worry about the exact pixels heights from this mockup, I am using the [review pattern](https://design.cms.gov/patterns/review/) from the CMS design system.
  - The action on edit is to jump you up to the activity section.

Text to copy in is here:

Activity 1: `activity name`
Continued Operations of the Medicaid EHR Incentive Payment Program, includes modifications to the SLR, HIT staff, auditing, outreach, and non-personnel expenses for administering the program.

Date: `start date xx/xx/xxxx` - `enddate xx/xx/xxxx`
Total cost of activity: `$XXX,XXX,XXX`
Medicaid share: `$XXX,XXX` (`$XXX,XXX` Federal Share)

Total cost:
Federal Fiscal Years requested: `FFY xxxx, xxxx, xxx`
Medicaid Share: `$XXX,XXX` (`$XXX,XXX` Federal Share)
Total Funding Request: `$XXX,XXX`

mgwalker:
This PR does not update the executive summary budget table, only the activity summary view, as designed in #1491.

- closes #1491

### DIFF
--- a/web/src/containers/ExecutiveSummary.js
+++ b/web/src/containers/ExecutiveSummary.js
@@ -1,6 +1,6 @@
 import { Review } from '@cmsgov/design-system-core';
 import PropTypes from 'prop-types';
-import React, { PureComponent } from 'react';
+import React, { Fragment, PureComponent } from 'react';
 import { connect } from 'react-redux';
 
 import ExecutiveSummaryBudget from './ExecutiveSummaryBudget';
@@ -35,24 +35,35 @@ class ExecutiveSummary extends PureComponent {
           >
             {data.map((activity, i) => (
               <Review
-                heading={`Activity ${i + 1}: ${activity.name ||
-                  t('activities.noNameYet')}`}
+                key={activity.key}
+                heading={
+                  <Fragment>
+                    Activity {i + 1}:{' '}
+                    <a
+                      href={`#activity-${activity.key}`}
+                      onClick={this.jump(activity.key)}
+                    >
+                      {activity.name || t('activities.noNameYet')}
+                    </a>
+                  </Fragment>
+                }
+                headingLevel={4}
                 editHref={`#activity-${activity.key}`}
-                onEditClick={this.jump(`#activity-${activity.key}`)}
+                onEditClick={this.jump(activity.key)}
                 className={i === data.length - 1 ? 'ds-u-border-bottom--0' : ''}
               >
                 {activity.summary && <p>{activity.summary}</p>}
 
                 <ul className="ds-c-list--bare">
                   <li>
-                    <strong>Date:</strong>
+                    <strong>Date:</strong> {activity.dateRange}
                   </li>
                   <li>
                     <strong>Total cost of activity:</strong>{' '}
                     <Dollars long>{activity.combined}</Dollars>
                   </li>
                   <li>
-                    <strong>Medicaid share:</strong>
+                    <strong>Medicaid share:</strong>{' '}
                     <Dollars long>{activity.medicaid}</Dollars> (
                     <Dollars long>{activity.federal}</Dollars> Federal share)
                   </li>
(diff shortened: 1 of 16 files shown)


web/src/containers/ExecutiveSummary.js
```diff
@@ -8,85 +9,112 @@ import { expandActivitySection } from '../actions/activities';
...
+
+import { jumpTo } from '../actions/navigation';
 import { selectApdYears } from '../reducers/apd.selectors';
-import { selectBudgetExecutiveSummary } from '../reducers/budget.selectors';
+import {
+  selectBudgetExecutiveSummary,
+  selectBudgetGrandTotal
+} from '../reducers/budget.selectors';
+
+class ExecutiveSummary extends PureComponent {
```
hbillings: If you have the time/inclination, I'd love to know more about why you changed this from a function-based component to a class-based pure component! Just my curiousity, though. :)

web/src/containers/ExecutiveSummary.js
```diff
@@ -8,85 +9,112 @@ import { expandActivitySection } from '../actions/activities';
...
-                </div>
-              </div>
-            </div>
-          </div>
-        ))}
-      </Subsection>
+            {data.map((activity, i) => (
+              <Review
+                heading={`Activity ${i + 1}: ${activity.name ||
+                  t('activities.noNameYet')}`}
```
hbillings: In line with Laura's suggestion, semantically this should be:
```suggestion
                  t('activities.noNameYet')}`}
                headingLevel={4}
```

mgwalker:
@lauraponce Whoops!  Clearly I was ready to be done and completely forgot to add the activity name links and dates. 😬

Everything you noted should be fixed now. One thing I bumped into is the case where maybe the user hasn't entered a start and end date yet.  If either one is missing, right now it will say `Dates not set`. It's very easy to change if there's a better way to handle that!

web/src/containers/ExecutiveSummary.js
hbillings: lol BEAT

web/src/containers/ExecutiveSummary.js
mgwalker: I changed it to a class because it has an instance method now, and functional components can't do that.  I used `PureComponent` because they render more efficiently.  `PureComponent` implements the React `shouldComponentUpdate` lifecycle method and does a shallow comparison of the current state and props to the incoming state and props.  If they match, it does not re-render.  The standard `Component` class, meanwhile, doesn't do any comparison so it will re-render every time the state or props change (or the parent re-renders!), even if the new and old are internally identical.  Those words don't make a lot of sense to me, so here's an example:

```js
const props1 = {
  name: 'Alice',
  age: 37
};

const props2 = {
  name: 'Alice',
  age: 37
};
```

In a standard `Component`, switching from props1 to props2 would cause a re-render because they are different objects.  A `PureComponent`, however, would shallowly compare the values in the two objects, see that nothing changed, and _*not*_ re-render.

jeromeleecms:
@mgwalker Is the total cost meant to be an `<H3>` or an `<H4>`? It's currently an `<H3>`, but seems to line up that this would fall under the previous `<H3>Activities summary</H3>`.

image

mgwalker:
@jeromeleecms Good catch.  Fixing that now.

PR #1519, opened 2019-05-06
Add footer to wrapper

hbillings:
All the pages have footers now! yay!

- Adds the footer to the `Wrapper` element so all pages inherit it
- Removes footer from files it previously existed in.

### DIFF
--- a/web/src/components/Wrapper.js
+++ b/web/src/components/Wrapper.js
@@ -1,6 +1,7 @@
 import PropTypes from 'prop-types';
 import React, { Component } from 'react';
 import Header from './Header';
+import Footer from './Footer';
 
 class Wrapper extends Component {
   componentDidMount() {
@@ -23,6 +24,7 @@ class Wrapper extends Component {
       <div className="site">
         <Header showSiteTitle={showSiteTitle} />
         {children}
+        <Footer />
       </div>)
     ;
   }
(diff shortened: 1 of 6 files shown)


hbillings:
@jeromeleecms Yes, please file the save button issue as a bug! 

Good catch with the height! There were styles that had been applied to the card layout that conflicted with the footer styles -- I just didn't realize it when I wrote the latter because it was only on a single page!

hbillings:
@jeromeleecms Whee!

The background color is a holdover, yes, but I think Jen kept it so that it would give the "card" some definition and the content on very small pages wouldn't float in too much space. @austinhernandez can probably weigh in on whether it's a construct we should keep or not. :)

hbillings:
@mgwalker wut. Does this only happen on smaller screens?

NAretakis:
@hbillings are we talking about the tota11y (sunglasses) icon? If so, not sure why its there, but pretty sure its something from tota11y.

mgwalker:
@hbillings Nah, I just narrowed the window so the screenshot wouldn't be huge.  It happens when the window is bigger, too.

hbillings:
@mgwalker The only way I can reproduce this on my end is if I zoom the browser window out, which makes sense because the thing that's gray is set to `100vh`. I'd be worried if it were broken at levels that were zoomed _in_ because that's an accessibility thing, but I'm less worried about it looking weird zoomed out. I'm going to go ahead and merge, and if it still looks weird to you (and you're at 100% zoom), let's open a bug!

Issue #1502, opened 2019-05-02
Design review of the Activity Schedule Summary Table


NAretakis:
This issue is to do a design review of the Activity Schedule Summary Table, per how it will look via the dev implementation issue #1446 . The goal of this issue is to make sure the start and end dates of activities are correctly being applied, as well as the dates for milestones (within activities). A few points to consider:

* Activities have start and end dates, specified within each activity's schedule
* Activities have milestone dates, specified within each activity's schedule

### This task is done when...
 - [x] Design review has been completed of the Activity Schedule Summary Table
 - [x] Mockups have been designed that give users information about activity start and end dates & milestone dates
 - [x] A mockup is selected to implement
 - [x] a development issue is created with the selected mockup

lauraponce:
![Screen Shot 2019-05-16 at 10 33 27 AM](https://user-images.githubusercontent.com/21222704/57875711-6ac57200-77c8-11e9-97c5-8ef4f7318e1e.png)

- Activity date is start date – end date
- Singular date for a milestone
- Formatting should be MM/DD/YYYY

lauraponce:
Created implementation issue #1537 for this!

PR #1536, opened 2019-05-16
Restructure activity schedule table

### LINKED ISSUE #1446: Restructuring Activity Schedule Summary Table
The activity list in section 5 uses the same table pattern as the Results of Previous Activities section:

5_Activity Schedule Summary - Relaxed

### This task is done when...
 - [ ] The Activity Schedule Summary is using the CMS design system

mgwalker:
- Switches the activity schedule summary table to a regular table
- Groups milestones under activities
- Sorts milestones by end date
- closes #1446 

### DIFF
--- a/web/src/containers/ScheduleSummary.js
+++ b/web/src/containers/ScheduleSummary.js
@@ -1,5 +1,5 @@
 import PropTypes from 'prop-types';
-import React, { Fragment } from 'react';
+import React from 'react';
 import { connect } from 'react-redux';
 
 import { t } from '../i18n';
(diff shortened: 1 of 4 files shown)


web/src/containers/ScheduleSummary.js
```diff
@@ -1,74 +1,57 @@
...
         ) : (
-          <ReactTable
-            showPagination={false}
-            resizable={false}
-            minRows={0}
-            className="h6 -striped"
-            {...tableData}
-          />
+          <table className="table-cms">
+            <tbody>
```
mgwalker: It feels a little weird to put headers in a `tbody`, but a table can only have one `thead` and we have header rows spread throughout...  So I guess we can't use `thead`? 😬

web/src/containers/ScheduleSummary.js
```diff
@@ -1,74 +1,57 @@
...
+              {activities.map(({ name: activityName, milestones }, i) => (
+                <Fragment key={activityName}>
+                  <tr>
+                    <th colSpan="2" className="ds-u-font-weight--bold">
+                      Activity {i + 1}: {activityName}
+                    </th>
+                  </tr>
+                  {milestones.map(({ end, name: milestoneName, start }) => (
+                    <tr>
+                      <td className="ds-u-padding-left--3 ds-u-border-right--0">
```
mgwalker: This is to get rid of the horizontal padding between cells in a row, to match the mockup in #1446

lauraponce:
@NAretakis should the singular date for the milestones be a start or an end date (or another date)? 

Is this what you're picturing?

Screen Shot 2019-05-16 at 10 33 27 AM

I also just noticed that the dates were written YYYY/MM/DD instead of MM/DD/YYYY — was there a reason behind that?

NAretakis:
@lauraponce yes, that's exactly what we need. Should just be a singular date for a milestone. With start and end dates for the activity itself (just like you have it). Also, formatting should be MM/DD/YYYY. Do you want to drop that design into #1502 ?

lauraponce:
@NAretakis added to #1502 ✔️

mgwalker:
I thought that date format was funny, but I went with it because it matched the mockup. Downside to coding while sick: you don't question *anything*. 😛

web/src/containers/ScheduleSummary.js
hbillings: Hmm. Could we maybe make the `activities.map` function create multiple tables stacked vertically, instead of making it one big table? Then it would be clear which header goes with which rows.

web/src/containers/ScheduleSummary.test.js
```diff
@@ -4,38 +4,59 @@ import React from 'react';
...
+        // Billboard with "Love Me Do" and wrap things up with "The Long and
+        // Winding Road"
+        { end: '1970/05/23', name: 'one-two', start: '1962/10/05' }
+      ],
+      // For the first time, Captain Picard gives Commander Riker his most
+      // famous order:  "Make it so, Number One."
+      start: '1988/12/10'
+    },
+    {
+      name: 'Number 💩',
```
hbillings: lol

web/src/containers/ScheduleSummary.js
mgwalker: Brilliant!

PR #1538, opened 2019-05-17
Patch for #1487

hbillings:
This just tweaks the CSS of #1487  to fix some spacing and grid weirdness.


web/src/containers/StateDashboard.js
```diff
@@ -3,8 +3,6 @@ import PropType from 'prop-types';
 import React, { Fragment } from 'react';
 import { connect } from 'react-redux';
 
-import Footer from '../components/Footer';
-import Header from '../components/Header';
```
hbillings: I originally merged master into this, which made it super messy. It also means I deleted the header, because it's in the wrapper in master. Oops?

Issue #1540, opened 2019-05-20
Agree and Consent Screen has an inconsistent look between IE, Chrome, and Firefox


jeromeleecms:
There is odd spacing that appears when viewing the consent screen across browsers. Screenshots below:

IE:
![image.png](https://images.zenhubusercontent.com/5c547884d11206601037f0c7/7b880f34-d836-483f-a09f-4ada8811d261)

Chrome:
![image.png](https://images.zenhubusercontent.com/5c547884d11206601037f0c7/cafbb9c4-04fd-40eb-bac1-d0091d91ccaa)

Firefox:
![image.png](https://images.zenhubusercontent.com/5c547884d11206601037f0c7/9b5d5731-b191-463d-9715-adfca853d2e7)

Also, this appears similar to the issue we encountered with the login screen (band below the login card in Chrome), but now only appears in Firefox:

![image.png](https://images.zenhubusercontent.com/5c547884d11206601037f0c7/5c7f6b6e-9d4a-491f-862a-cb0fec30575b)

Is this related to the page template and/or some one off issue with the way the card layout is structured for the consent screen? 

### This task is done when...
 - [ ] Consent screen appears consistent on browsers
 - [ ] Underlying issue is reviewed/addressed (i.e. adding new pages similar to this style won't result in one off styling corrections)

hbillings:
#1541 should address this.

jeromeleecms:
:thumbsup: Assigning @hbillings on this so we can track the consent and the login screen per the note above.

PR #1541, opened 2019-05-20
Fix white gap under small cards

hbillings:
Greg noticed this on the footer PR, but I couldn't reproduce it locally. Now I see it! So I fixed it!


jeromeleecms:
@hbillings I just spun up a bug under #1540 for this, does this also solve the issue described there?

mgwalker:
FWIW, I still see the gap on a normal-sized window on my external monitor. The height of the white gap is proportional to the height of the window, so if I make it exaggeratedly tall, the gap gets huge, or the gap goes away entirely if I shorten the window height.

hbillings:
@mgwalker I think in order to address that, we're going to need to find a way to make the page background gray, rather than the current solution which changes the color of the card container background. The only way I know of to for sure fix it the way it currently exists is to set the card container to `height: 100vh`, which will always push the footer off the page, even when there's not much content in the card. Maybe that looks less weird than a gap?

mgwalker:
Oof, setting the page background would be a _thing_...  It's outside the React ecosystem, so we could use some DOM APIs to do it, but I'm not sure where would be the right place to do that.  It almost seems like the only place we could reliably do it would be in the route handler.  🤔

OH!  OH!  This might be a good use of a context - they're essentially a way of creating global state within a React app.  https://reactjs.org/docs/context.html  The `Wrapper` could add a class to the `<div className="site">` based on the value in the context?

PR #1557, opened 2019-05-28
Update production deploy cleanup

mgwalker:
Currently, the production deployment will only successfully cleanup instances that have completed their deployments (i.e., they've been created and added to the load balancer).  However, in the case that a deployment begins while a previous deployment is still ongoing, the latter deployment may fail to clean up the former and we could end up with two active deployments.  This could cause unpredictable behavior if the API changed, as different versions of the API are responding to requests.

- cleans up all previously-existing production instances when the new instance is ready, regardless of their status

### DIFF
--- a/bin/prod-deploy/aws.sh
+++ b/bin/prod-deploy/aws.sh
@@ -35,9 +35,10 @@ function deployAPItoEC2() {
   # Configure AWS CLI with defaults
   configureAWS
 
-  # Get the instance ID of the current API EC2 instance, so we can delete it later
-  PREV_INSTANCE_ID=$(getCurrentTargetGroupInstanceID)
-  echo "• Found previous instance $PREV_INSTANCE_ID"
+  # Get the instance IDs of the current API EC2 production instances, so we can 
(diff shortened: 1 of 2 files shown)


bin/prod-deploy/aws.sh
```diff
@@ -178,21 +177,6 @@ function getTargetHealth() {
...
-#   AWS_PROD_API_PORT - Port the EC2 instance is listening on
-#   AWS_PROD_API_TARGET_GROUP_ARN - The ARN of the target group to remove the
-#     instance from.
-function removeInstanceFromTargetGroup() {
-  aws elbv2 deregister-targets \
-    --target-group-arn $AWS_PROD_API_TARGET_GROUP_ARN \
-    --targets Id=$1,Port=$AWS_PROD_API_PORT \
-    > /dev/null
-}
-
```
mgwalker: Terminating the instance also removes it from the target group. Removing this step should speed up deployments a little bit because it's one less thing we have to wait on.

PR #1565, opened 2019-05-30
update activity goals section

### LINKED ISSUE #1453: Add review pattern to Performance Goals and Benchmarks (Program Activities subsection) and update to use design system tokens
Within the Program Activities section, the 2nd subsection is Performance Goals and Benchmarks. 

image

### This task is done when...
 - [x] The `Collapsible` component has been replaced with the design system review pattern.
 - [x] "Performance Goals and Benchmarks" is `h4.ds-h4`
 - [x] Each section is wrapped in a `div.data-entry-box`
 - [x] Available entry boxes are "Goal Name" (short text entry field) and "Benchmarks" (textarea).
 - [x] The collapsed view shows "Goal [number]: [goal name]" in `h5.ds-h4` and the "benchmark" text.
 - [x] Goals may be edited once collapsed or removed if there is more than one goal in the list. (Note that the current mockup does not allow for removing the first goal, but probably all we care about is making sure that there is at least one goal.)
 - [x] New goals may be added with the "Add another goal" button (would work similar to the contacts in the Key Personnel section using the review component.)
 - [x] Section has an `hr` at the bottom

mgwalker:
- updates activity goals section

  image

  - goals can be removed as long as there is more than one in the list
  - users are prompted for confirmation before goals are deleted
  - new goals are expanded by default

- updates dev seed to remove HTML formatting from goal name and benchmark text
- closes #1453 

### DIFF
--- a/web/src/components/Section.js
+++ b/web/src/components/Section.js
@@ -53,7 +53,7 @@ const Subsection = ({ children, id, nested, resource }) => {
         </h3>
       )}
       {nested && (
-        <h4 id={id} className="ds-h3">
+        <h4 id={id} className="ds-h4">
           {title}
         </h4>
       )}
--- a/web/src/containers/activity/Goals.js
+++ b/web/src/containers/activity/Goals.js
@@ -58,8 +58,10 @@ const GoalForm = ({ goal, idx, done, handleChange }) => (
     <TextField
       name="name"
       className="data-entry-box"
-      label="Goal name"
-      hint={t('activities.goals.goal.instruction.detail', {
+      label={t('activities.goals.description.input.label', {
+        defaultValue: ''
+      })}
+      hint={t('activities.goals.description.input.hint', {
         defaultValue: ''
       })}
       value={goal.description}
@@ -69,10 +71,12 @@ const GoalForm = ({ goal, idx, done, handleChange }) => (
     <TextField
       name="milestones"
       className="data-entry-box"
-      label="Benchmarks"
+      label={t('activities.goals.objective.input.label', {
+        defaultValue: 'Benchmarks'
+      })}
       multiline
       rows={6}
-      hint={t('activities.goals.objective.instruction.detail', {
+      hint={t('activities.goals.objective.input.hint', {
         defaultValue: ''
       })}
       value={goal.objective}
@@ -175,7 +179,7 @@ class Goals extends Component {
           className="ds-u-margin-top--4 visibility--screen"
           onClick={this.handleAdd}
         >
-          {t('activities.goals.addGoalButtonText')}
+          Add a goal
         </Button>
         <hr />
       </Subsection>
--- a/web/src/containers/activity/Goals.js
+++ b/web/src/containers/activity/Goals.js
@@ -16,9 +16,9 @@ import { t } from '../../i18n';
 const GoalReview = ({ goal, idx, edit, handleDelete }) => (
   <Review
     heading={
-      <h5 className="ds-h4">
+      <h4 className="ds-h4">
         {idx + 1}. {goal.description}
-      </h5>
+      </h4>
     }
     editHref="#"
     onEditClick={edit}
--- a/web/src/containers/activity/Goals.js
+++ b/web/src/containers/activity/Goals.js
@@ -38,7 +38,9 @@ const GoalReview = ({ goal, idx, edit, handleDelete }) => (
       </div>
     }
   >
-    <p className="ds-u-margin-top--2">Benchmark: {goal.objective}</p>
+    <p className="ds-u-margin-top--2">
+      <strong>Benchmark:</strong> {goal.objective}
+    </p>
   </Review>
 );
 
--- a/web/src/containers/activity/Goals.js
+++ b/web/src/containers/activity/Goals.js
@@ -76,7 +76,7 @@ const GoalForm = ({ goal, idx, done, handleChange }) => (
         defaultValue: ''
       })}
       value={goal.objective}
-      onChange={handleChange('idx', 'objective')}
+      onChange={handleChange(idx, 'objective')}
     />
 
     <Button variation="primary" className="ds-u-margin-top--4" onClick={done}>
@@ -141,11 +141,9 @@ class Goals extends Component {
     addActivityGoal(activityKey);
   };
 
-  handleDelete = goal => this.getDeleter(goal.key)();
-
-  handleSync = (index, field) => html => {
+  handleChange = (index, field) => ({ target: { value } }) => {
     const { activityKey, updateActivity } = this.props;
-    const updates = { goals: { [index]: { [field]: html } } };
+    const updates = { goals: { [index]: { [field]: value } } };
     updateActivity(activityKey, updates);
   };
 
@@ -164,7 +162,7 @@ class Goals extends Component {
                 goal={goal}
                 idx={i}
                 initialExpanded={goal.expanded}
-                handleChange={this.handleSync}
+                handleChange={this.handleChange}
                 handleDelete={
                   goals.length > 1 ? this.getDeleter(goal.key) : null
                 }
@@ -208,3 +206,12 @@ export default connect(
   mapStateToProps,
   mapDispatchToProps
 )(Goals);
+
+export {
+  Goals as plain,
+  Goal,
+  GoalForm,
+  GoalReview,
+  mapStateToProps,
+  mapDispatchToProps
+};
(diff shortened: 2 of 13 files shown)


web/src/components/Section.js
```diff
@@ -53,7 +53,7 @@ const Subsection = ({ children, id, nested, resource }) => {
         </h3>
       )}
       {nested && (
-        <h4 id={id} className="ds-h3">
+        <h4 id={id} className="ds-h4">
```
mgwalker: Note to @hbillings, this changes the styling for all of the sections.  Is that the intended result?  If not I'll need to go back in and tweak it.

web/src/containers/activity/Goals.js
```diff
@@ -1,61 +1,133 @@
...
+            <Button size="small" variation="transparent" onClick={handleDelete}>
+              Remove
+            </Button>
+          </Fragment>
+        )}
+      </div>
+    }
+  >
+    <p className="ds-u-margin-top--2">Benchmark: {goal.objective}</p>
+  </Review>
```
mgwalker: This is the third time I've done this exact pattern and I'm thinking it'd make sense to create a custom wrapper around `<Review>` that does all the header stuff for us automatically, and also consistently...

web/src/containers/activity/Goals.js
```diff
@@ -1,61 +1,133 @@
...
+  handleDelete: null
+};
+
+const GoalForm = ({ goal, idx, done, handleChange }) => (
+  <div className="ds-u-padding-bottom--2 ds-u-border-bottom--2">
+    <TextField
+      name="name"
+      className="data-entry-box"
+      label="Goal name"
+      hint={t('activities.goals.goal.instruction.detail', {
```
mgwalker: Not wild about coopting the `instruction` content structure here. Should we also have an `input` structure?

```yaml
input:
  label: Goal name
  hint: Write your goal name here, obviously
```

I don't know.  Our YAML situation is kind of a mess, with some stuff being there and other stuff not...

web/src/reducers/activities.js
```diff
@@ -32,7 +32,12 @@ export const setKeyGenerator = fn => {
   generateKey = fn;
 };
 
-const newGoal = () => ({ key: generateKey(), description: '', objective: '' });
+const newGoal = () => ({
+  key: generateKey(),
+  description: '',
+  expanded: true,
```
mgwalker: New goals default to expanded

jeromeleecms:
@mgwalker in Chrome/IE, I can't seem to type anything into the boxes (new or editing old ones)...

mgwalker:
@jeromeleecms I was so focused on getting the appearance right I completely forgot to make sure the fields were actually editable...  🤦‍♂  Should be fixed.

jeromeleecms:
@mgwalker The dev work matches the mockup :tada:

I did want to get @austinhernandez and/or @jameshupp opinion on this:

image

In our instruction text, we ask the state to "Explain the state's goal for this activity." But in the actual section, we have "Goal **Name**" and then jump right to the "Benchmarks" with no explicit area for the state to provide an explanation. Should we have a separate section for the state to provide the explanation .. or can/should we just modify the Benchmarks -> Description and Benchmarks + update the instruction text to have the state provide the explanation + benchmarks in that provided text box (should that become a text editor box?)?

I'm leaning toward the latter for the pilot, but wanted to see if there might be some user experience implications that I might be missing here.

Also, if we need to move this conversation into a separate issue, I'm OK with that too, and I can approve the PR and the instruction text can come afterwards.

web/src/containers/activity/Goals.js
```diff
@@ -1,61 +1,133 @@
...
 import { t } from '../../i18n';
 
-const plaintext = txt => {
-  const e = document.createElement('div');
-  e.innerHTML = txt;
-  return e.innerText;
+const GoalReview = ({ goal, idx, edit, handleDelete }) => (
+  <Review
+    heading={
+      <h5 className="ds-h4">
```
mgwalker: @hbillings Should this be `<h4 className="ds-h4">`?  The header above is an `h3` and the React nesting validator is complaining about having a 5 directly under a 3.  I wasn't clear on why it was a 5 so I didn't change it yet, but raising because...  I think maybe it should be a 4.  😛

web/src/containers/activity/Goals.js
jameshupp: It's not especially helpful to just say 💯 here but 💯. My own take is:

* Content of any significance should be in YAML. It's pretty rough going back and forth.
* YAML things should have semantic meaning to the editors. `instruction` should be about telling people what to do and, if they are confused, how to do it. It should not govern interactive things. If we have inputs, we should have something else. `input` if that's it; something else if we want to govern, say, inputs and widgets and games of hangman. This is doubly true because the YAML doesn't correspond directly to the structure of the site.

mgwalker: 👌 I've changed the YAML.  I added a key for each field, and put an input key under those, like this:

```
objective:
  input:
    label: Benchmarks
    hint: >-
      Briefly describe how the state will measure progress toward the identified goal.
```

I also took the "Add a goal" button text out of the YAML, but happy to put it back if you think it belongs.  I struggle with what qualifies as having "significance."  I assume things like button text generally don't, and it's more about text that describes what something is or how to use it.

At one time, ***all*** of the user-facing text was in YAML, but that seemed like such overkill, and it was hard to keep track of too.  Finding the middleground is proving difficult (to me, at least), too.  😬

jameshupp:
> Should we have a separate section for the state to provide the explanation .. or can/should we just modify the Benchmarks -> Description and Benchmarks + update the instruction text to have the state provide the explanation + benchmarks in that provided text box (should that become a text editor box?)?

@jeromeleecms can you say briefly what explanation you actually want? I am inclined not to label that field `Description and Benchmarks`; a single input should generally have a single ask. Otherwise we get a lot of different structures: benchmark-by-benchmark descriptions, a description and then a list of benchmarks, a description with benchmarks mixed in at various points, etc.

We could do something with the instructions themselves to make that clearer, though? Maybe just `Description` as a title, and then an instruction like: 
> Briefly describe how this goal relates to the vision identified in the program summary. Then provide any benchmarks the state will use to measure progress toward the identified goal.

(That would involve moving the first sentence from the section instructions to the ones for this specific input field.)

---
If this feels like too big a set of things to discuss, go ahead and approve the merge and we can figure this out as a separate need.

jeromeleecms:
Thanks @jameshupp I'll spin up a separate issue now. Approving in 3...2...

austinhernandez:
As long as the format makes sense for the content (e.g. input vs text area) everything is good with me! The only feedback I have is to bold the "benchmarks" label to keep it consistent with the rest of the app. Sorry that some of the mockups are unbolded, they were in varying stages of approval.

web/src/containers/activity/Goals.js
hbillings: The heading levels are my best attempt at guessing what things should be, which is super hard to do piecemeal like this. If the validator is complaining, it knows better than I do!

The one thing that seems like it might throw a wrench into this is where reviews are used in nested elements (activities). Those probably need to have a heading level that is lesser than the heading levels of non-nested components, so maybe we make the level/class overrideable but with sensible defaults? (I did something like this in the cost allocation PR with the `Instruction` component: https://github.com/18F/cms-hitech-apd/pull/1569/files#diff-1f5a9d3d843586bd0de4807c9b3f7442R7)

web/src/containers/activity/Goals.js
hbillings: yesplz

web/src/containers/activity/Goals.test.js
```diff
@@ -0,0 +1,273 @@
...
+describe('activity goals subsection', () => {
+  const sandbox = sinon.createSandbox();
+  const dispatchProps = {
+    addActivityGoal: sandbox.spy(),
+    removeActivityGoal: sandbox.spy(),
+    updateActivity: sandbox.spy()
+  };
+
+  beforeEach(() => {
+    sandbox.resetHistory();
```
hbillings: if only it were that easy

web/src/containers/activity/Goals.test.js
```diff
@@ -0,0 +1,273 @@
...
+
+      expect(handleObjectiveChange.calledOnce).toEqual(true);
+    });
+  });
+
+  describe('the GoalReview component', () => {
+    const edit = sandbox.spy();
+    const handleDelete = sandbox.spy();
+
+    it('renders properly without a delete delete', () => {
```
hbillings: okay okay

PR #1570, opened 2019-06-03
Bring activity overview section in line with design system

### LINKED ISSUE #1452: Bring design elements in line with design system for Activity Overview (Program Activities subsection)
Within the Program Activities section, the first subsection is `Activity Overview`. Don't worry about the exact pixels or size discrepancies in the mockup, please use default Design System and we can fine tune global styles later.

Note that the content of the overview is shown as nested, but we're not going to implement that right now. Keep the margins aligned with the higher-level header.

image

### This task is done when...
- [ ] `Activity Overview` is `h4.ds-h4`.
- [ ] `Textarea` has been swapped for the design system component
- [ ] Each section is wrapped in a `div.data-entry-box` (controls the spacing/margin between each section -- there's probably a better way to do this but I haven't figured it out yet)
- [ ] There is an `hr` at the end of the Overview section

mgwalker:
- renames `container/activity/Description.js' to `containers/activity/Overview.js`
- replaces instructions with form labels/hint text
- replaces textarea with design system textfield
- use memoized text and event handlers for better performance
- closes #1452


web/src/containers/activity/Overview.js
```diff
@@ -0,0 +1,144 @@
+import { FormLabel, TextField } from '@cmsgov/design-system-core';
```
mgwalker: git helpfully decided this was a delete/add situation rather than a rename situation, so congrats, you get to review the whole file!

web/src/containers/activity/Overview.js
```diff
@@ -0,0 +1,144 @@
...
+        rows={6}
+        className="data-entry-box"
+        value={summary}
+        onChange={overviewOnChange}
+      />
+
+      <div className="data-entry-box">
+        <FormLabel hint={descriptionHint}>{descriptionLabel}</FormLabel>
+        {activity.fundingSource === 'HIE' && (
+          <Instruction source="activities.overview.activityDescriptionInput.hie" />
```
mgwalker: HIE activities have additional helptext that the `FormLabel` doesn't support.  If we have that pattern a lot, we should consider wrapping `FormLabel`, but this felt fine for a one-off.

mgwalker:
Ha!  @hbillings and I had [discussed the spacing on Slack](https://gsa-tts.slack.com/archives/C8E5EJK9V/p1559578951118000?thread_ts=1559578579.117300&cid=C8E5EJK9V) briefly, but the short answer is the design system form elements (including input boxes as well as their labels) have a maximum width, but the rich text area is not (yet, at least) constrained the same way.

Heather said these are on the next round of standardization, so I'd defer to her on whether this is something we need to address in this PR or if it'll be fixed more broadly in the near future.

jeromeleecms:
Does this weird spacing also apply to the "Statement of Alternative Considerations and Supporting Justifications?" It doesn't look like a label...

mgwalker:
Yes, that text is using the design system `<FormLabel>` component, which has the same width restriction as inputs.

web/src/containers/activity/Overview.js
hbillings: yay

web/src/containers/activity/Overview.js
```diff
@@ -0,0 +1,144 @@
...
+
+  const descriptionLabel = useMemo(
+    () =>
+      t(
+        `activities.overview.activityDescriptionInput.${
+          activity.fundingSource === 'HIE' ? 'hie' : 'standard'
+        }.label`,
+        {
+          defaultValue:
+            'Include as much detail as is necessary to explain the activity.'
```
hbillings: Just so I'm sure I follow: this is coming from the YAML in some cases, but in case it's not, we're defining the default value here?

mgwalker: Correct, though I'm not sure it's necessary.  Maybe we'd prefer there be an error if these are missing from the YAML?  🤷‍♂

Issue #1180, opened 2018-11-27
Further refine content guide to account for Layer 1 updates


jameshupp:
We have a basic content guide in place on the wiki. We are updating both content elements and some style choices. Those need to be reflected back into the guide.

### This task is done when...
 - [x] Incomplete sections of the content guide are complete
 - [x] The content work described in a comment on #1083 has been completed. (1083 itself will be complete once that comment is submitted, so that issue will show up as closed. But it is a useful reference point.)

jameshupp:
The 1083 work is either closed or in the icebox to look at again later, so I'm calling that closed for now.

jameshupp:
Here's the basic list of things I'm working on for the content wiki. If there's more that should be included, lemme know.

1. style
    a. the basics (capitalization, numbers, punctuation, etc.)
    b. terminology (abbreviations, choices, etc.)
2. Elements
    a. front end (section, subsection, etc.)
    b. content management (YAML elements, things in YAML vs other things, etc.)
3. content workflows
    a. relationship 
    b. publishing
    c. getting things into the testing workflow
    d. getting support

If there's time I might spin up another issue to figure out how to work some of this into Storybook but that's a secondary concern.

jameshupp:
Lots of new stuff in here. Doesn't mirror the outline above exactly, but there's a lot more info for someone about style, about certain terms that someone needs to understand, about what's managed where, etc. https://github.com/18F/cms-hitech-apd/wiki/Content-guide

PR #1608, opened 2019-06-14
lots of little adjustments to copy

### LINKED ISSUE #1177: Clarify how the app treats key personnel for subsequent cost breakdowns
This builds on #1128, which changed the way the app processes key personnel costs.

The key personnel costs are included under Program Administration, which is always an HIT cost and therefore a 90-10 cost. And then any adjustment to the quarterly share of Program Administration would also apply to the way the federal share of the key personnel cost is broken out.

The task here is: make that clear to state staff in a useful way.

### This task is done when...
 - [x] Appropriate help text has been added alongside the key personnel fields to clarify this need.
 - [ ] The presentation of cost breakdowns later in the app have been reviewed and any reminders or additional helpful commentary for those sections have been added.
 - [ ] Review of whether this content is helpful has been added to the protocol for an upcoming round of testing.

### LINKED ISSUE #1599: Fix discovered typos and style issues on the way out
In completing #1180, I noticed a number of issues that need to be addressed in the content. Some of these I can do; some I'll just have to poke an engineer in the arm about over the next day or two. Maybe they turn into new issues, or maybe they get resolved. The checkboxes in the list are for whether we know what to do next, not whether they've been resolved.

Sorry, I know it's bad to throw something directly into this sprint, but, well, I don't have another option if I'm going to take this on and it's a really light lift.

### To be completed
- [x] HIE Overview helptext: “prior to developing HIE funding request” should have “a” or be plural
- [x] HIO is organization not office
- [x] Linebreaks in standards and conditions please!
- [x] Medicaid director and Medicaid office don’t align left correctly
- [x] Goal names aren’t editable suddenly?
- [x] Private contractor costs field labels are h3 tags but no other field labels anywhere seem to be. Am I wrong about what these are?
- [x] Title case for Proposed Budget tables pretty much top to bottom
- [x] Title case for things under Cost Allocation

jameshupp:
Closes #1599

Might close #1177, depends on someone’s interpretation, I’ll leave that to the rest of the team.

- This addresses a number of minor typos and stylistic errors I noticed while I was updating the content style guide per #1180.
- This also adds information about Key Personnel being 90-10 HIT expenses for #1177.

This is ready to merge when nothing breaks and product says the content is clearer than it was before on key personnel. It can certainly use additional work but I want it to be more clear rather than less.

### DIFF
--- a/web/src/i18n/locales/en/activities/costAllocate.yaml
+++ b/web/src/i18n/locales/en/activities/costAllocate.yaml
@@ -41,8 +41,8 @@ ffp:
     other: Other Funding Amount
     fedStateSplit: Federal-State Split
 quarterly:
-  title: Breakout of federal share by quarter
   instruction:
+    heading: Breakout of federal share by quarter
     detail: >-
       Indicate allocations of state personnel & contract resources in current FFY
       by quarter.
@@ -54,7 +54,7 @@ otherFunding:
   instruction:
     heading: Other funding description
     detail: >-
-      Write a high level description to describe funding used to support this activity.
+      Write a high level description of funding used to support this activity.
       (Review Letter 10‐016 for examples of grants for inclusion.)
     helpText: >-
       Ensure new grants are included, and remove expired grants.
--- a/web/src/i18n/locales/en/apd/apd.yaml
+++ b/web/src/i18n/locales/en/apd/apd.yaml
@@ -28,13 +28,13 @@ hie:
     heading: HIE Overview
     detail: >-
       Summarize planned HIE-funded activities within this APD.
-      
+
       If there are no HIE activities, enter “No HIE activities.”
 
 
       [ol]
-      Description of the state's HIE approach (statewide, sub-state
-      health information office (HIOs), etc.);
+      Description of the state’s HIE approach (statewide, sub-state
+      health information organization (HIO), etc.);
 
       Anticipated risks and mitigation strategies;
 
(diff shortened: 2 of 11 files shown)


web/src/i18n/locales/en/activities/costAllocate.yaml
```diff
@@ -54,7 +54,7 @@ otherFunding:
   instruction:
     heading: Other funding description
     detail: >-
-      Write a high level description to describe funding used to support this activity.
+      Write a high level description of funding used to support this activity.
```
jeromeleecms: "of" or "of the"

web/src/i18n/locales/en/apd/apd.yaml
```diff
@@ -44,13 +44,13 @@ hie:
       [/ol]
     helpText: >-
       CMS recommends that states connect with regional CMS HITECH contacts prior to
-      developing HIE funding request. HIE funding requests are complex. Early
+      developing an HIE funding request. HIE funding requests are complex. Early
```
jeromeleecms: "a HIE"? Did some searching, using "a" before H when the sound is a consonant and use "an" before H when the sound is a vowel sound...

mgwalker: Depends on whether you pronounce the initialism (like "hi" or "hee" or something), or if you pronounce the letters ("eich eye ee").  In the former case, it'd be "a HIE" because the word begins with a hard sound; but in the latter, it'd be "an HIE" because the pronunciation of the letter "H" begins with a soft sound.  This is one of those quirky areas of the English language because the written form is dictated by the pronunciation, which of course varies from person to person.

And I don't feel bad about stepping into James's turf because James's watch is ended.

PR #1611, opened 2019-06-17
Limit failed logons within a period of time

### LINKED ISSUE #1508: Enforce a limit on consecutive failed logon attempts
[NIST 800-53 AC-7](https://nvd.nist.gov/800-53/Rev4/control/AC-7) requires action after a defined number of consecutive failed logon attempts within a set period of time.  The possible actions are:
- the account is locked until some time has passed
- the account is locked until an administrator unlocks the account
- the logon prompt is delayed according to a delay algorithm

Need product input to decide which of these three to adopt.  The first is easiest to implement.  😄

### This task is done when...
 - [ ] when consecutive logon attempt threshold is reached, the account action is taken

mgwalker:
If a user attempts and fails to log into an account `N` times within `X` minutes, that account is then locked for `Y` minutes and no one may log into it.

- adds environment variables
   - `AUTH_LOCK_FAILED_ATTEMPTS_COUNT` - the number of failed logon attempts before the account is locked; default is 5.
   - `AUTH_LOCK_FAILED_ATTEMPTS_WINDOW_TIME_MINUTES` - the number of minutes in which the failed logon attempts must occur; default is 1. All five failed logon attempts must occur within this time window.
   - `AUTH_LOCK_FAILED_ATTEMPTS_DURATION_MINUTES` - the number of minutes to lock the account; default is 30.
- saves timestamps of failed logons, counts how many in the past `X` minutes
- if `X` reaches `N`, marks the account as locked until `Y` minutes in the future
- when the lock is expired, erases the timestamps of failed logons and the lock timeout
- on successful logons, erases the timestamps of failed logons and the lock timeout (so if you fail to logon twice and then succeed, if you for some reason logged out and failed two more times, the second two would represent a new list of failed logons)
- attempting to log into a locked account appears exactly the same to the outside world as attempting to log into an invalid account or using the wrong password (this way we don't reveal the existence of valid accounts)
- closes #1508


docs/database-diagram.dot
```diff
@@ -1,5 +1,5 @@
 /*
```
hbillings: unknown filetype

Issue #1385, opened 2019-03-20
Implement replacement of typography system - use CMS Design System styles


jenniferthibault:
The prototype site used Source Sans Pro, but the CMS Design System uses Open Sans universally.

We should replace the font family universally, and start tackling how to remove the old prototype's atomic styling in favor of CMS Design System classes.

Describe the task here...

### This task is done when...
 - [x] Source Sans Pro has been smited from the code base

cc @hbillings who can probably get more specific :)

hbillings:
SMITE IT, SMITE IT ALL

We'll need to remove classes that set the font face to the font we're no longer using. We'll also need to make sure that changing the font face doesn't inadvertently change some sort of spacing, etc.

mgwalker:
Is this one finished by #1612, or are there a few more places that still need attention?

PR #1597, opened 2019-06-10
Fix up activity schedule dates

mgwalker:
### This pull request changes...
- adds activity date range to activity header row
- switches milestone rows to just end date
- formats dates as `mm/dd/yyyy`
- closes #1537

### This pull request is ready to merge when...
- [ ] Tests have been updated (and all tests are passing)
- [ ] This code has been reviewed by someone other than the original author
- [ ] The experience passes a basic manual accessibility audit (keyboard nav, screenreader, text scaling) OR an exemption is documented
- [ ] The change has been documented
  - [ ] Associated OpenAPI documentation has been updated

### This feature is done when...
- [ ] Design has approved the experience
- [ ] Product has approved the experience

PR #1612, opened 2019-06-18
Update all the tables

### LINKED ISSUE #1544: Implement new, coherent design for tables
From the design ticket #1480 we are ready to implement!

image

Rationale:
- This project is super table heavy and there are a lot of different styles and functions of these tables. I tried to make it clear when to use the style of each one, but they are pretty similar. 
- I thought it might be important to add a special class to the "sum" or "total" rows and columns because that is the result of the data you are manipulating, but open to suggestions if this is not feasible.
- To help read long forms and rows of data, I thought it would be helpful to add a hover class to the tables to help focus on the row of data yr mouse is over

### This task is done when...
 - [ ] all the tables look like the mockups
 - [ ] design approves!

Here are the design colors, fonts and border styles used!
Fonts used ([from design system](https://design.cms.gov/style/color/), with the exception of courier for the numerals):
- [x] Open Sans semi-bold for headers (horizontal and vertical)
- [x] Open Sans bold for table caption, subtotals, totals, and sums
- [x] Courier for #s and $$$$s
- [x] Open Sans for regular fonts inside the table
- [x] Don’t use center alignment - numerical data is right-aligned, textual data is left-aligned, and headers are aligned with their data if possible.

Colors and border styles ([from CMS design system](https://design.cms.gov/style/color/))
- [x] Header `#D6D7D9` `$color-gray-lighter` header background, border-top and bottom solid border: `#AEB0B5` `$color-gray-light`
- [x] TRs/TD `#FFFFFF` header background, border-top and bottom dotted border: `#D6D7D9` 
 `$color-gray-lighter`
- [x] Special row or column call out sums `#FFF1D2` `$color-warn-lightest` background - Let's discuss if this is feasible or a good idea!
- [ ] Hover highlight row `#F1F1F1`  `$color-gray-lightest` keeps the same inherited borders
- [ ] inherit link colors when necessary, without an underline

*Also somewhat weirdly blocking #1425

hbillings:
Closes #1544 and #1385. This was a doozy to write, but I think it should be pretty simple to review!

- Changes the `table-cms` class to `budget-table` so that we can BEM-ify modifier classes.
- Replaces Source Sans Pro with Open Sans.
- Replaces our prior "monospace" font (that wasn't really monospace) with Courier.
- Rearranges a few tables for another inch toward consistency.
- Introduces the concept of subtotal/total rows with bold and highlighted text.
- Makes a solid first pass at accessibility.

There ARE still some funky tables; namely ones with double-decker headings. I think we can simplify those further, but that should be a much easier task to design and implement now.


web/src/containers/ApdPreviousActivityTable.js
```diff
@@ -71,8 +69,8 @@ const ApdPreviousActivityTable = ({
                       name={`hithie-approved-total-${year}`}
                       label={`total approved funding for HIT and HIE for FFY ${year}, state plus federal`}
                       hideLabel
-                      wrapperClass="m0"
-                      className="m0 input input-condensed mono right-align"
+                      wrapperClass="budget-table--input-holder"
+                      className="budget-table--input__number"
```
hbillings: semantic css!! it's a thing!!

web/src/containers/ApdPreviousActivityTableMMIS.js
```diff
@@ -26,23 +26,21 @@ const ApdPreviousActivityTableMMIS = ({
     <Fragment>
       <h4 className="ds-h4">MMIS</h4>
       {[90, 75, 50].map((level) => (
-        <table key={level} className="table-cms centered-headers table-fixed">
+        <table key={level} className="budget-table">
           <thead>
             <tr>
-              <th id="prev_act_mmis_null2" />
-              <th className="pre-line" id={`prev_act_mmis${level}_total`}>
+              <th id="prev_act_mmis_null2" rowSpan="2" />
```
hbillings: No real reason to include two null `th` cells when `rowspan` will do the trick with one!

web/src/containers/BudgetSummary.js
```diff
@@ -12,25 +12,24 @@ const categoryLookup = {
   combined: 'Subtotal'
 };
 
-function DataRow({ data, title, year }) {
+function DataRow({ data, title }) {
   return (
-    <tr>
-      <th headers="summary-budget-null1 summary-budget-null2">{title}</th>
+    <tr
+      className={ title===categoryLookup.combined ? 'budget-table--subtotal budget-table--row__highlight' : '' }
```
hbillings: I'm sure there's a better way to do this than to generate a blank class, but I wasn't coming up with it.

web/src/containers/BudgetSummary.js
```diff
@@ -12,25 +12,24 @@ const categoryLookup = {
...
+    >
+      <th scope="row">{title}</th>
```
hbillings: For these simpler tables without double decker headers, we can forgo the complicated `headers` business in favor of simpler `scope`.

web/src/containers/IncentivePayments.js
```diff
@@ -36,18 +34,31 @@ class IncentivePayments extends Component {
...
+              <caption className="ds-u-visibility--screen-reader">
+                {t('ffy', { year })}
+              </caption>
               <thead>
                 <tr>
-                  <th />
+                  <th>
+                    <span aria-hidden="true">
+                      {t('ffy', { year })}
+                    </span>
```
hbillings: In lines 38-46, we're creating a table caption that will be read by screen readers once, prior to reading the table contents, rather than creating a situation where the screen reader will say `FFY 2019` before _every single cell_.

web/src/styles/app/base.css
```diff
@@ -4,10 +4,6 @@ html {
   font-size: var(--base-font-size);
 }
 
-body {
-  font-family: var(--base-font-family);
-}
-
```
hbillings: @jenniferthibault SOURCE SANS PRO HAS BEEN SMITED FROM THE CODEBASE
miss you comrade

hbillings:
@austinhernandez One of the funkier tables is the very last set in the exec summaries:
Screen Shot 2019-06-17 at 9 54 58 PM

I broke out the "Medicaid totals" row (it was previously part of the bigger table) because it was just too squishy. @jeromeleecms should probably take a peek to see if this change makes sense, or if there's something else we should do.

I wanted to make all the columns the same size, but if I make them all small enough to match MMIS, it looks super weird. I have a feeling this is a table that should really be multiple tables, but I'm not sure how to correctly break them apart.

web/src/containers/IncentivePayments.js
```diff
@@ -36,18 +34,31 @@ class IncentivePayments extends Component {
       <Fragment>
         {years.map(year => (
           <Fragment key={year}>
-            <h3 id={thId(year)}>{t('ffy', { year })}</h3>
-            <table className="table-cms table-fixed">
+            <table className="budget-table">
+              <caption className="ds-u-visibility--screen-reader">
+                {t('ffy', { year })}
```
mgwalker: Should we have an issue for moving this table header out of this cell and into an `<h#>` like the other tables?

web/src/containers/activity/CostAllocateFFPQuarterly.js
```diff
@@ -67,37 +77,29 @@ class CostAllocateFFPQuarterly extends Component {
...
                           name={`ffp-${aKey}-${year}-${q}-${name}`}
                           label={`federal share for ffy ${year}, quarter ${q}, ${name}`}
-                          wrapperClass="m0"
-                          className="m0 input input-condensed mono right-align"
+                          wrapperClass="budget-table--input-holder"
+                          className="budget-table--input__number"
                           onChange={this.handleChange(year, q, name)}
                           value={quarterlyFFP[year][q][name].percent * 100}
                           hideLabel
+                          aria-controls={`ffp-${aKey}-${year}-${q}-${name}-dollar-equivalent`}
```
mgwalker: oooooh

web/src/styles/app/base.css
mgwalker: thumbs up sheep

web/src/styles/scss/forms.scss
```diff
@@ -43,3 +44,30 @@ div.form--computed-value__input-aligned {
 .rdw-editor-toolbar {
   border-radius: $border-radius;
 }
+
+/* dolla dolla inputs */
```
mgwalker: dolla dolla bills

web/src/styles/scss/tables.scss
```diff
@@ -1,33 +1,70 @@
...
+      font-weight: $font-bold;
+      th, td {
+        font-weight: $font-bold;
+      }
+    }
   }
-  + .table-cms:not(.table-frozen-data){
-    margin-top: $spacer-2;
+
+  thead {
```
mgwalker: I ***ALWAYS*** misread this as `<thread>` and I have to stop and think what the heck the `thread` element is...  Probably something @edmullen invented.

web/src/styles/scss/tables.scss
```diff
@@ -1,33 +1,70 @@
...
+    th {
+      background-color: $color-gray-lighter;
+      white-space: pre-line;
+      /* used for highlights in a column */
+      &.budget-table--col__highlight {
+        background-color: $color-warn-lightest;
+      }
+    }
+    border-top: 1px solid $color-gray-light;
+    border-bottom: 1px solid $color-gray-light;
```
mgwalker: Do all these colors come from the design system?  I'm just a little confused because I know ace-css brings in some colors too, but I guess those are classes, not variables.

web/src/styles/scss/type.scss
```diff
@@ -1,17 +1,22 @@
 @import '@cmsgov/design-system-support/src/settings/index';
+@import './variables';
+
+body {
+  font-family: $font-sans;
```
mgwalker: fixed it

austinhernandez:
For the tables with two headers I [used the pattern from MailChimp](https://ux.mailchimp.com/patterns/tables) where there is a simple bottom border separating the two with the headers aligned left.

image

The labels would be aligned left with the content, so the red line is imaginary to show alignment.

hbillings:
@jeromeleecms 
1. They're semibold; the only bold rows should be the subtotal and total rows.
2. I defer to @austinhernandez, but we can always iterate on this if it turns out to be too light a weight!
3. Good catch; I right-aligned those quarter headers.
4. I'm going to defer to Austin again on this one, but I think this is one of those things we can iterate on.
5. This is taken care of in #1597, right? If so, this should look A-OK once that one is also merged.
6. Yes... I wasn't entirely sure when to call out subtotals and when to not. For instance, in that screenshot, would both the subtotal column and the Total Enhanced FFP row get highlighting? Maybe another @austinhernandez question. 😬 
7. 🤷‍♀ The inputs shouldn't really have changed size in this PR, so if that's something we need to consider, let's open a new issue to think about it.
8. I re-added the totals column to the main tables. My main concern here is readability, especially with the MMIS table; it gets pretty squished even with zeroes in some fields. I think if we could break those last three total columns into their own table, it would help, but this might also be something to address when we assess the printability of things.

web/src/styles/scss/tables.scss
hbillings: Yesssss, these are all the design system colors! Good gut check!

web/src/styles/scss/tables.scss
hbillings: If only there were a single, defined, semantic way to hierarchically converse....

web/src/containers/activity/CostAllocateFFPQuarterly.js
hbillings: run awaaaaaayyyyy

hbillings: Actually, I meant to put a comment in this part... there's some _weird_ interactive shit going on here and I'm really not entirely sure what an accessible workflow should be. I opened a ticket to assess it further. 😬

web/src/containers/IncentivePayments.js
hbillings: I'm a little confused, tbh. It seems like most of the mocks that @austinhernandez has going actually move the year _into_ the corner cell as a standardization. But then there are also tables with multiple years so... I don't even know any more.

web/src/containers/ExecutiveSummaryBudget.js
```diff
@@ -34,185 +34,229 @@ const ExecutiveSummaryBudget = ({ budget }) => {
...
+              key={year}
+              className={display==='Total' ? 'budget-table--total budget-table--row__highlight' : ''}
+            >
+              <th
+                headers="program-budget-table-null1 program-budget-table-null2"
+                scope="row"
+              >
+                {display}
               </th>
+              <DollarCell
```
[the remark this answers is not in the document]
hbillings: WHERE DID YOU FIND THIS this is incredible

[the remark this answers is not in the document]
mgwalker: I cannot reveal my sources

mgwalker:
Re: point 5, that is correct. Once this PR is merged, I'll go handle any merge conflicts in #1597 and we'll be good to go.

austinhernandez:
Things are lookin up! Just one small line issue:

Can the line stretch the entire width like the mock above?
Screen Shot 2019-06-20 at 1 04 49 PM

Like this one does? 
Screen Shot 2019-06-20 at 1 05 52 PM

jeromeleecms:
@austinhernandez 

1. Does the extended line pattern above also apply to these tables? 

image

2. Should the right hand subtotal for the cost allocation be in a different background color (looks like.. #fff1d2)?

image

image

3. @hbillings Awesome. Can we pair with @austinhernandez sometime (maybe working session?) to get wrap this up? I want to make sure we are applying the consistent styling throughout the tables. For the other items you responded to 👍 - always open to iteration!

hbillings:
The line should be taken care of; I think at this point all I need to know is which, if any, subtotal rows should be highlighted that aren't.

austinhernandez:
Here are all the places without a subtotal highlight:

Federal Financial Participation (FFP) and Cost Allocation
Screen Shot 2019-06-21 at 11 04 06 AM

quarterly federal share HIT, HIE, and MMIS
Screen Shot 2019-06-21 at 11 04 34 AM
Screen Shot 2019-06-21 at 11 04 40 AM

austinhernandez:
Per @jeromeleecms 's comment "...while both are important, i would make the total row MORE important. " So, the general rule we should follow in that case is if there is a total and subtotal, highlight and bold the total, and just keep the subtitle row bold, but not highlighted. Correct me if I misunderstood, Jerome.

For example, Subtotals bolded in this one
Screen Shot 2019-06-24 at 5 04 40 PM

But not in other ones (under Quarterly Federal Share).
Screen Shot 2019-06-24 at 5 04 20 PM

But I'm not sure what to do in the case of Medicaid Total - @jeromeleecms should that be highlighted?
Screen Shot 2019-06-24 at 5 04 40 PM

hbillings:
@austinhernandez Is that last screenshot supposed to be of this table?
Screen Shot 2019-06-25 at 9 50 29 AM

PR #1620, opened 2019-06-20
Adds cacheing and security headers to the API

### LINKED ISSUE #1520: Cache in IE causes changes made in the APD not to appear when exiting the opening the APD
In IE only (works in Firefox), when making changes in the APD building, saving, exiting the APD and reopening the APD does not show the changes. Verified this behavior against multiple PR (#1517 and #1518) and main.

From Slack convo
1) if I save in PR1517, exit the APD, and go back in, the update doesn't appear.
2) if I save in PR1517, go to PR1518 in a different tab, view the APD, the change appears. No save. Then if I go back to PR1517, got the dashboard, and then reopen the APD, the change appears.

Per @mgwalker, we will spin up an issue to investigate if this if this is related to caching of API data (which signs point to yes).

### This task is done when...
 - [ ] Changes in IE can be seen whenever exiting and re-entering an APD

mgwalker:
- disables the `X-Powered-By` header
   - This is a non-standard header that some server applications will add to a response to identify themselves. But what good is identifying the server application? 😱 Not a super huge security concern, but it can be helpful with throwing malicious automated scans off the trail a bit.
- adds a `Cache-Control` header
   - sets `private` - This instructs proxies that they should never cache the API responses because they may contain sensitive or authenticated data. Since that's true for literally everything in our API aside from the OpenAPI documentation, it makes sense to tell proxies no.
   - sets `no-cache` - This instructs browsers that before using cached data, they should go ahead and make the request to the server and send the `If-None-Match` header set to the cache's etag value with their request (the etag value would have been sent by the server in a previous response). If the server builds a response and computes the same eTag as the browser sent, it will respond with a 304 status ("Not Modified") to let the browser know to use its cache. (This sequence is basically the browser saying, "Hey the last time I asked for this, you said it was abc123. Is that still true?" And the server responding with a 304 if it's still abc123, but sending the new data if it has changed.)  This should fix issues where the browser is using its cache by default, and it should also reduce network traffic in some cases.  (But not Chrome, because they ignore the hell out of this and reload every time. 🤷‍♂)
- adds a `Strict-Transport-Security` header
   - This instructs the browser that it should always use HTTPS for this site. The header is configured so that the browser will remember this instruction for one year, and also apply it to all subdomains. (In an emergency, browsers will let you go into settings and reset this, but it's a good practice to enable it in production.)
- adds a `X-Content-Type-Options` header
   - This tells the browser to always obey the `Content-Type` header sent by the server. Without this, some browsers (***coughcough***chrome***cough***ie***cough***) will ignore what the server says and instead inspect the response content to try to infer what it is. Which is kinda dumb, but it's an artifact of a decade or two ago, when `Content-Type` was set far less reliably than it is now.  Also, IE would sometimes guess that a response was executable and then it'd...  well, *execute* it. And nothing has *ever* gone wrong by executing random crap you find on the internet, right?
- adds a `X-Frame-Options` header
   - This tells the browser that this site cannot be put into a frame. There's no reason this should ever be in a frame, so disable it. Frames are a pretty good vector for cross-site-scripting (XSS) attacks.
- adds a `X-XSS-Protection` header
   - This header tells the browser to turn on the XSS protection that it already has built-in and probably turned on anyway, but might as well be explicit. Besides, security scanners will likely to be looking for this.
- closes #1520 


api/main.js
```diff
@@ -35,6 +41,35 @@ server.get('/heartbeat', (_, res) => {
...
+  // the next year.
+  res.header(
+    'Strict-Transport-Security',
+    'max-age=31536000; includeSubDomains'
+  );
+
+  // Instruct browsers to use the content-type indicated by the server rather
+  // than ignore that and try to guess the content-type from the response body
+  // (IE used to do this and would sometimes decide that things were executable
+  // and then run them, with I'm sure HILARIOUS results)
```
hbillings: uh-huh

PR #1623, opened 2019-06-21
fix summary budget table 'total' header

### LINKED ISSUE #1607: FFY total for Summary Budget Tables needs to be capitalized
This builds off #1599 and making this issue lets me close out the last to-do item from that issue.

The tables under the Summary Budget Tables subsection are the only ones that have been using sentence case for the various labels (rows, columns, etc.). I changed most of them to title case on a branch and am going to submit a PR for it shortly. But there's one label — `FFY total` — that is generated from data elsewhere; it's not a text string that I can just go edit. Someone who knows their way around the code needs to make it display as `FFY Total`; note the capital `T`. This will bring all the tables into consistent capitalization styling.

mgwalker:
The table header described in #1607 is the total across FFYs in the "Summary Budget Table" in the "Proposed Budget" section.  The header was being computed based on the year, but for the total across FFYs, the year field is `total`.  Thus, the header was displaying as `FFY total`.

The fix is to make the header text conditional so that if the year field is `total`, it will display something else.  Because `FFY Total` doesn't really make sense (it isn't a total of a single FFY and there is no FFY called "total"), I changed it to just `Total`.

Since @quarterback is back (:tada:), I'd love feedback on the decision to change `FFY total` to just `total`.  It makes sense to *me*, but who am I but a lowly, humble dev? 😄

#### Before

screenshot showing the original header that says FFY total with improper capitalization

#### After

screenshot showing the revised header that just says Total

- closes #1607 

### DIFF
--- a/web/src/containers/BudgetSummary.js
+++ b/web/src/containers/BudgetSummary.js
@@ -73,7 +73,7 @@ const HeaderRow = ({ yr }) => {
         className="ds-u-font-weight--bold"
         id={`summary-budget-fy-${yr}`}
       >
-        FFY {yr}
+        {yr === 'total' ? 'Total' : `FFY ${yr}`}
       </th>
       <th
         className="ds-u-text-align--center"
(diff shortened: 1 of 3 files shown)


jeromeleecms:
@quarterback @mgwalker How about making it consistent with the Quarterly Federal Share? 

1) Proposed = "Total"
image

2) Quarterly Federal Share = "Total MMIS" or "Total HIT + HIE"
image

Though looking at this, does "HIT Total" or "Total HIT" make more sense?

Issue #1501, opened 2019-05-02
Implement review pattern for Activity List


hbillings:
There are two review sections necessary to replace the current Activity List/Activity Overview sections. They both use the CMS review pattern. #1452 is the Overview issue.

![image](https://user-images.githubusercontent.com/23194720/58493520-1cf22780-8141-11e9-8b0a-0ed2a92a3f69.png)

### This task is done when...
 - [ ] There is one expanded section by default
 - [ ] Expanded activity sections have two data entry points, an activity name text box and a type radio button
- [ ] Expanded activities may be completed or cancelled, with the exception of the first one
- [ ] "Add another activity" is visible
- [ ] When collapsed, the list follows the [CMS review pattern](https://design.cms.gov/patterns/review/)

austinhernandez:
When we reviewed this section together as a team 4/22 we decided we were going to keep the current functionality of the `Activity List` the same as it is currently, so the above mockup is incorrect. We will be creating the list of activities first, then populating the sections below.

Attached is how the activity list should look with the new styles.

![image](https://user-images.githubusercontent.com/23194720/58493520-1cf22780-8141-11e9-8b0a-0ed2a92a3f69.png)

hbillings:
@austinhernandez Can you update the main issue with this, rather than having it in a comment? (You should be able to click the three dots in the bar with my username, and then select "edit."

PR #1631, opened 2019-06-25
Update activity bodies to use review/expand pattern

### LINKED ISSUE #1561: Change out Collapsible elements for review pattern in Program Activities
We need to build out the review pattern to work with activities (screenshot in #1450).

### This task is done when...
- [ ] Instead of the current `Collapsible` element, the list of activity overviews 
- [ ] The first activity is expanded by default.
- [ ] The close button collapses the section.
- [ ] When collapsed, only the title of the activity is visible.
- [ ] The title of the activity is `h3.ds-h3`.
- [ ] Once collapsed, the text would change to edit and it would open up!

mgwalker:
Might want to hold off on finalizing this one until #1501 is merged because the two are ***right next*** to each other, and I'm not 100% sure whether 1501 adds a bottom border or not, so the boundary between the two is weird right now.  I figure it'll sort out once that one is merged.  It's probably fine to go ahead and review this, just be aware about that one boundary being wonky.

- gets rid of the `Collapsible` component, which served us well, but now its time is passed
- brings in the review pattern from other places and crams it in here too (I mean, it fits pretty nicely so not really "crammed in")
- closes #1561


web/src/containers/activity/EntryDetails.js
```diff
@@ -28,72 +24,73 @@ const makeTitle = ({ name, fundingSource }, i) => {
...
-      </Collapsible>
-    );
-  }
-}
+  return (
+    <div
+      id={`activity-${aKey}`}
+      className={`activity--body activity--body__${
+        collapsed ? 'collapsed' : 'expanded'
+      } activity--body__${num === 1 ? 'first' : 'notfirst'}`}
```
hbillings: lulz

We might be able to nix this by using the `:first-child` pseudoclass, but it'd require changing the markup, and I think I want to be able to look at the whole thing in the design system before I can really suggest optimal approaches.

PR #1634, opened 2019-06-26
Fix an issue where changing APD years broke everything

### LINKED ISSUE #1633: [Pilot Critical] Fix Dirty State Handling with the App
Per @mgwalker:

The APD defaults to the current fiscal year and the next fiscal year. It looks like you added the next fiscal year after that as well, but internally, the data didn't update to account for the third year so it's blowing up.  It's _supposed_ to add data when you add a year, though. We've got a special section in the code just for that purpose!

It's a bug in the dirty state handling. If you change the APD years and save, it doesn't save the placeholder data. 

### This task is done when...
 - [ ] Selecting New FFY's in the app doesn't cause the system panic
 - [ ] Saving and Reopening the APD is done without issue
 
mgwalker:
When you change the years on an APD, the app also updates several other fields that rely on the APD years:
  - activity contractor resources
  - activity cost allocation
  - activity non-personnel expenses
  - activity state personnel expenses
  - APD incentive payments
  - APD key personnel

However, our "dirty state" tracking wasn't marking those fields as dirty when the APD years changed, so a subsequent APD save could lose the data for newly-selected years.  Then when the page was reloaded, the data from the API was missing key pieces and the app would crash because it can't work without all of that data.

This PR fixes the "dirty state" tracking so that it marks those year-based fields as dirty whenever the APD years change.

I propose that we create another issue to address this on the backend as well: the API can (should!) provide default data for years when that data is missing from the database.

- fixes #1633


api/seeds/development/apds.js
```diff
@@ -67,26 +67,6 @@ exports.seed = async knex => {
...
-      q4_eh_payment: 0,
-      q1_ep_count: 0,
-      q2_ep_count: 0,
-      q3_ep_count: 0,
-      q4_ep_count: 0,
-      q1_ep_payment: 0,
-      q2_ep_payment: 0,
-      q3_ep_payment: 0,
-      q4_ep_payment: 0
-    },
```
mgwalker: It hasn't been FFY 2018 for a while so there's no good reason to keep this data in here... 😛

web/src/containers/admin/CreateAccount.js
```diff
@@ -4,7 +4,7 @@ import React, { Component, Fragment } from 'react';
 import { connect } from 'react-redux';
 import CardForm from '../../components/CardForm';
 import Password from '../../components/PasswordWithMeter';
-import { STATES } from '../../util';
+import { STATES, toSentenceCase } from '../../util';
```
mgwalker: I did a rebase to fix a commit message and I think I rebased over _two_ commits, so this old one got caught in here as well.  This stuff is already on master but it's a different commit hash so 🤷‍♂ I guess.

hbillings:
@mgwalker If you created the issue you mentioned in the PR description, could you link to it here?

web/src/reducers/dirty.js
```diff
@@ -52,9 +52,26 @@ const up = updates =>
...
+    case SELECT_APD:
+      // When we select an APD, stash off all the activity keys so we'll
+      // have them to iterate over later, if the APD years change (see
+      // below for more explanation of why).
+      return u(
+        {
+          data: {
+            activities: {
+              byKey: action.apd.activities.reduce(
+                (byKey, activity) => ({ ...byKey, [activity.key]: false }),
```
hbillings: What does the `false` indicate?

EDIT: Never mind, I figured it out (more or less) by reading the tests, which means they're good tests, Brent.

mgwalker:
@hbillings It's #1633

PR #1654, opened 2019-07-01
Fix border and spacing on rich text fields

mgwalker:
- removes border around rich text element
- removes min-height on rich text

### DIFF
--- a/web/src/styles/scss/print.scss
+++ b/web/src/styles/scss/print.scss
@@ -38,11 +38,10 @@
     display: none;
   }
 
-
   .ds-l-container,
   .ds-l-col--9 {
-    width: 100%!important;
-    min-width: 100%!important;
+    width: 100% !important;
+    min-width: 100% !important;
   }
 
   .accordian {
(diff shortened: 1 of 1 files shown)


web/src/styles/scss/print.scss
```diff
@@ -38,11 +38,10 @@
     display: none;
   }
 
-
   .ds-l-container,
   .ds-l-col--9 {
-    width: 100%!important;
-    min-width: 100%!important;
+    width: 100% !important;
+    min-width: 100% !important;
```
hbillings: Nitpicking the CSS, I see. I approve. 😆

web/src/styles/scss/print.scss
```diff
@@ -80,6 +79,15 @@
...
     border: none !important;
   }
+
+  .public-DraftEditor-content {
+    border: 0;
+  }
+
+  .rte-textarea-l .public-DraftEditor-content,
+  .textarea-l {
+    min-height: 0;
```
hbillings: Was there a reason this had a `min-height` at all?

mgwalker: No idea.  Maybe it's to mimic the textareas?  Without a `min-height`, it's just as tall as the content, so it kind of looks like a one-line textfield.

web/src/styles/scss/print.scss
mgwalker: Blame `prettier`!

PR #1667, opened 2019-07-01
Make numbered sections readable in print

hbillings:
Does this:
Screen Shot 2019-07-01 at 5 04 26 PM

### DIFF
--- a/web/src/styles/scss/print.scss
+++ b/web/src/styles/scss/print.scss
@@ -76,13 +76,13 @@
     padding: 0 !important;
     border: none !important;
   }
-}
 
-.numbered-section > h2:before {
-  background-color: transparent !important;
-  font-weight: bold !important;
-  color: black !important;
-  border: 1px solid black;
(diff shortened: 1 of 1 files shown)


web/src/styles/scss/print.scss
```diff
@@ -30,7 +30,7 @@
   .btn-collapse svg,
   .alert,
   .rdw-editor-toolbar,
-  .table-frozen-left-pane {
+  .ds-c-review .nowrap {
```
mgwalker: Neat! I was expecting to put a `visibility--screen` class on the edit bar, but I like this approach better. Would it make sense to make a special class for the edit bar instead of targeting the `nowrap` though?  🤷‍♂

hbillings: Thanks to the weird way I daisy-chained these, it's now tough for me to go back and edit this for this PR. But I'll fix it in the next one!

PR #1670, opened 2019-07-01
Tweaks for legends and fieldsets in print

hbillings:
Don't break inside fieldsets.

Legends should be 100% width.

### DIFF
--- a/web/src/styles/scss/print.scss
+++ b/web/src/styles/scss/print.scss
@@ -20,7 +20,8 @@
 
   table,
   figure,
-  textarea {
+  textarea,
+  fieldset {
     page-break-inside: avoid;
   }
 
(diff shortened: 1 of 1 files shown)


web/src/styles/scss/print.scss
```diff
@@ -20,7 +20,8 @@
 
   table,
   figure,
-  textarea {
+  textarea,
+  fieldset {
```
mgwalker: oh

web/src/styles/scss/print.scss
```diff
@@ -41,6 +42,12 @@
     min-width: 100%!important;
   }
 
+  fieldset,
+  .ds-c-label {
+    width: 100% !important;
```
mgwalker: wide

hbillings: This is the perfect gif, good job, we can all go home now

PR #1682, opened 2019-07-08
Handle empty/invalid dollar and date values

### LINKED ISSUE #1658: Determine the best way to handle blank inputs
There are several places in the app that have different ways of handling blank inputs. For example:

`Date not specified` and `$0`
Screen Shot 2019-07-01 at 1 38 43 PM

`Undefined`
Screen Shot 2019-07-01 at 1 38 51 PM

`Blank`
Screen Shot 2019-07-01 at 1 38 57 PM

`N/A`
Screen Shot 2019-07-01 at 1 39 15 PM

Some fields leave it blank, some use 0, some use NA, some use undefined, and maybe more. Since we aren't doing validation, we should figure out the best way to handle blank values and apply that to all fields.

### This task is done when...
 - [ ] There is an established way of handling blank inputs

mgwalker:

- empty/invalid dollar values are displayed as `$0` instead of `--`
- empty/invalid single dates are displayed as `Date not specified`
- empty/invalid date ranges are displayed as `Dates not specified`
- several components updated to use util date methods, so we can make sure everything is consistent
- closes #1658

### DIFF
--- a/web/src/containers/ScheduleSummary.js
+++ b/web/src/containers/ScheduleSummary.js
@@ -16,42 +16,38 @@ const ScheduleSummary = ({ activities }) => (
             {t('scheduleSummary.noDataMessage')}
           </div>
         ) : (
-          activities.map(
-            ({ end, name: activityName, milestones, start }, i) => (
-              <table key={activityName} className="budget-table">
-                <caption className="ds-u-visibility--screen-reader">
-                  Activity {i + 1}: {activityName}
-                </caption>
-                <thead>
-                  <tr>
-                    <th
-                      className="ds-u-font-weight--bold ds-u-border-right--0"
-                      style={{ width: '70%' }}
-                    >
-                      Activity {i + 1}: {activityName}
-                    </th>
-                    <th className="ds-u-font-weight--bold ds-u-padding-right--3 ds-u-text-align--left ds-u-border-left--0 budget-table--cell__nowrap">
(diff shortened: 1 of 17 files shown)


web/src/containers/ScheduleSummary.js
```diff
@@ -16,42 +16,38 @@ const ScheduleSummary = ({ activities }) => (
...
-                          {milestoneEnd}
-                        </td>
-                      </tr>
-                    )
-                  )}
-                </tbody>
-              </table>
-            )
-          )
+          activities.map(({ name: activityName, dateRange, milestones }, i) => (
```
mgwalker: One of the downsides of prettier is that it reformats whole big chunks of code sometimes and ends up changing indentions but nothing else, so it looks like more changed than really did. :sigh: Maybe GH's "hide whitespace" option helps.

web/src/containers/activity/ContractorResource/ContractorResourceReview.js
```diff
@@ -1,18 +1,8 @@
 import PropTypes from 'prop-types';
 import React, { useMemo } from 'react';
-import Review from '../../../components/Review';
-
 import Dollars from '../../../components/Dollars';
-
-const getHumanReadableDateRange = (startDate, endDate) => {
```
mgwalker: Use the util method instead of doing its own thing. Yay consistency?

PR #1687, opened 2019-07-09
Remove unused code

mgwalker:
- removes the old material-ui "snackbar" notifications
- removes the wrappers around the `react-dates` date pickers
- removes code around uploading contractor files
   > This code contained references to the old notifications and wasn't being used anyway. Rather than try to "fix" unused code, I just deleted it. It's still in our git history, so when it comes time to add file uploads back, we'll still have it for reference if we need it, though I suspect we'll reconsider how we do uploads then.
- removes code around submitting and withdrawing APDs
  > Same as above.
- Removed a bunch of components we aren't using anymore
  - Btn
  - ConfirmAlert
  - DeleteConfirm
  - Label
  - MiniHeader
  - NoDataMsg
  - Select
  - StateDashboardCurrentDocuments
  - StateDashboardSidebar
- Removed validation utility functions because we weren't using them and I'm pretty sure they're wrong anyway; we'll do validation properly when we do it
- moved the `ProposedBudget` component from the `components` directory into the `containers` directory, so it's in the same place as all the other APD-level components instead of being off by itself in some weird way
- no user-visible changes


web/src/actions/activities.js
```diff
@@ -76,79 +70,6 @@ export const addActivityContractor = (key, { save = saveApd } = {}) => async (
   dispatch(updateActivity(key, updates));
 };
 
-export const uploadActivityContractorFile = (
```
hbillings: Guessing these aren't actions we need to take any longer?

web/src/actions/apd.test.js
```diff
@@ -475,13 +474,7 @@ describe('apd actions', () => {
       const store = mockStore(state);
       fetchMock.onPut('/apds/id-to-update').reply(403, [{ foo: 'bar' }]);
 
-      const expectedActions = [
-        { type: actions.SAVE_APD_REQUEST },
-        {
-          type: notificationActions.ADD_NOTIFICATION,
-          message: 'Save successful!'
```
hbillings: We're still displaying the `Save successful!` message, though, aren't we?

mgwalker: Yep. It's handled by the `SaveButton` component now.

web/src/actions/activities.js
mgwalker: Indeed, at least not for now. We used to support uploading arbitrary attachments for contractor resources, but we took it out because it didn't work right. Rather than just leave it all in there, I yoinked it. "YAGNI" as @abrouilette says, and when we do eventually need this functionality again, we can look back at this history to decide if we want to pull it up or do it differently.

PR #1697, opened 2019-07-16
Notify users before logging them out due to inactivity

### LINKED ISSUE #1666: Implement a timeout warning
Timeout happens after 1 hour of signed in activity. If so, consider:

- Letting users know there is an inactivity timeout
- Giving them a warning/chance to save before logging out
- Extending the time limit or allowing it to be adjusted.

mgwalker:

- when a user's session will be logged out due to inactivity within the next two minutes...
   - flashes their browser tab title
      screenshot of tab title flashing
   - issues a browser notification that they can click (**Chrome, Firefox, Safari, and Edge only - not IE**)
      screenshot showing browser notification
      - we can customize the title/heading text, but not the layout
      - we cannot change the URL, that is set by the browser
      - we can customize the body text, but not the layout
      - we cannot change the buttons, but we can make things happen when you click it
         - right now clicking it resets the idle timeout and focuses the tab
- activity in any tab containing eAPD will reset the idle timer in all eAPD tabs (**Chrome and Firefox only; Edge support is coming in a future browser update**)
- closes #1666


web/src/components/Broadcast.js
```diff
@@ -0,0 +1,52 @@
+import { useEffect } from 'react';
```
mgwalker: It seems a little weird to put this into a React component, but it gives us two advantages that I can see:
1. It makes initialization explicit, rather than happening when the module loads, which makes it easier to test. We could accomplish this with an `init` method or something similar as well, but...
2. It allows us to hook when the DOM begins to unload so we can also close the broadcast channel. I'm not sure if it's necessary in the case of the page shutting down, but because it's a shared resource across tabs, I was worried about memory leaks and the browser never being able to release the channel once created.

web/src/components/Broadcast.js
```diff
@@ -0,0 +1,52 @@
...
+  }
+};
+
+const Broadcast = () => {
+  useEffect(() => {
+    try {
+      if (window.BroadcastChannel) {
+        channel = new BroadcastChannel('eapd_activity');
+        channel.onmessage = ({ isTrusted, data }) => {
+          if (isTrusted) {
```
mgwalker: The DOM [`Event.isTrusted` property](https://developer.mozilla.org/en-US/docs/Web/API/Event/isTrusted) is cool because it indicates a user action rather than a scripted action. I'm not sure that's what we want for a *generic* broadcast handler, but it's what we want for this particular PR.

It might be worth revisiting at some point, though - how neat would it be if you accidentally open multiple tabs of eAPD and then they all stay in sync because when you update one, it broadcasts the update to the other tabs?

pretty cool

mgwalker:
Question for @jeromeleecms and/or @austinhernandez: when should the notification trigger? Right now it's 2 minutes before you are logged out. Is that too short? I was trying to shoot for something long enough a user would have a chance to see it and react, but not so long that they get notified too often.

web/src/containers/IdleLogout.js
```diff
@@ -16,26 +17,97 @@ const events = [
...
+      const idleWarning = () => {
+        stopFlashing = flashWindowTitle();
+
+        if (window.Notification) {
+          notification = new Notification('Session expiring', {
+            body: 'You will be logged out soon',
+            requireInteraction: true
+          });
+          notification.onclick = () => {
+            resetTimer(); // eslint-disable-line no-use-before-define
```
mgwalker: Normally I like this lint rule that you can't use things that the reader hasn't seen before, but it's tough to figure out how to break that in a circular dependency situation.

mgwalker:
1. Yes, that's probably right. We can't really differentiate between "clicking the tab" and "clicking the page," we have to accept whatever the browser tells us. It'll also reset the timer if your browser is unfocused and you just casually move your mouse across it - you're not really interacting with it, but the browser sends us a single "mousemove" event anyway.  Ultimately I think this is fine - the spirit of the security control is to handle the case where someone gets up and walks away from their computer, leaving the site exposed. As long as we're detecting that they're probably still at the computer, that seems okay.
2. The browser notification won't show up at all unless you grant it permission. It _*should*_ have asked after you logged into the preview site. If you say no or just close the request box, the browser won't allow the notifications. With permission, the notification should show up wherever system notifications normally show up for you (I'm not sure where they show up in Windows). We should dig into this one some more - notifications depend on operating system settings, browser settings, and user interaction before they can show up at all, so there's a lot that could prevent them.
3. A precise amount of time makes more sense to me than a percentage time.  If two minutes is ever sufficient warning, then it seems like it should always be sufficient.  If it sometimes isn't, then it probably never is. Definitely interested in hearing from design, though - this seems like a very clear UX issue. 😄

web/src/components/Broadcast.js
hbillings: MAGIC

Issue #1434, opened 2019-04-11
Inconsistent Stickiness Sidebar Behavior with Internet Explorer


jeromeleecms:
#1430 appears to work fine in Chrome, but has some interesting behavior on Internet Explorer.

1) Sidebar resizes on scroll

APD Load: 

![image.png](https://images.zenhubusercontent.com/5c547884d11206601037f0c7/e7da6dd1-52f7-4d1f-bd00-a4086efcdcb0)

APD on scroll:

![image.png](https://images.zenhubusercontent.com/5c547884d11206601037f0c7/0f4e2a5f-70dd-4eb2-b186-2c2cba1dc940)

2) ~~Page content is not centered in IE~~ Not sure when this was fixed, but IE content is centered.

### This task is done when...
 - [ ] Left side navigation remains consistent in size during scrolling in IE
 - [ ] The sidebar has appropriate bottom whitespace when the page is scrolled to the bottom (if possible - see [the first image](https://github.com/18F/cms-hitech-apd/pull/1721#issuecomment-524382855) to see the problem
 - [ ] ~~Page content is centered in IE~~

mgwalker:
The sidebar vertical spacing issue may be unsolvable, unless we want to write some javascript to track where the sidebar is on the page. I think we could "solve" the sidebar by making it fixed at `top: 0` once its relative top goes below 0, but we'd have to have scroll handling Javascript to enforce that, and we definitely don't want that in browsers that support sticky.

PR #1721, opened 2019-07-23
Keep the save button out of the footer

### LINKED ISSUE #1524: Save Button in the APD builder of the app appears in the footer
When scrolling through the APD Builder portion of the app, the Save button should stop at the bottom of the APD (not float through to the end footer).

### This task is done when...
 - [ ] Save button does not appear in the footer of the app
 
mgwalker:

- makes the save button sticky instead of fixed, so it is constrained to the APD form body and cannot extend into the footer
- updates the save button prop type to allow a boolean as well as a string for the error message, so we'll stop getting React warnings about `false` values
- closes #1524


web/src/containers/SaveButton.js
```diff
@@ -41,6 +42,10 @@ class SaveButton extends PureComponent {
     return null;
   }
 
+  componentDidMount() {
+    stickybits('#apd-save-button', { verticalPosition: 'bottom' });
```
hbillings: yay

jeromeleecms:
@mgwalker IE is funny (looks great in Chrome). 

Save Button location in IE...

image

But as you get close to the end... poof! Save button disappears completely!

image

mgwalker:
🤦‍♂

hbillings:
@mgwalker I'll see if I can get an IE VM up and running to troubleshoot this, if you like!

mgwalker:
@hbillings That would be great, if you can! I'm stuck with a busted VM installation and still fighting it.

mgwalker:
The disappearing save button in IE feels like a dealbreaker. I ***STILL*** can't get Virtualbox installed on this machine (there's an issue installing the kernel extension, and GSA IT is going to have to enable something... but I dunno if that's ever going to happen... 🤷‍♂)

mgwalker:
Okay, we figured out the problem, but the solution is unclear. So we don't forget, here's the basic problem.

On IE, stickybits uses `fixed` positioning to simulate `sticky`, until the parent element scrolls out - then it switches to `absolute` positioning. When that switch happens, the save button jumps to the top of its parent, which is not the page itself but rather the APD container.

mgwalker:
Made some progress! The reason the button jumps to the top is that there is no positioned element in the button's ancestry, so it gets positioned relative to the window, with a `top: 0;`. We can fix that pretty easily by setting `position: relative;` on the button container's parent and setting `bottom: 0px` on the button container itself. Need to double-check that it doesn't mess anything up in other browsers, though.

mgwalker:
@jeromeleecms I'm pretty sure it works in IE now! Pretty small change in the end, it turns out.

@hbillings My only concern is putting a `position: relative;` on the APD container, but it doesn't seem to have affected anything. The `@support` stuff is ***COOL***.

jeromeleecms:
It sort of works now... in IE (this is from scrolling through the builder):

image

But when you resize the window, it jumps back down (though the nav bar doesn't disappear anymore 🎉):  

image

There is also an odd spacing issue in the nav bar and the footer. In the first example, there is no space, but upon a resize, the space appears (which is what I expected from Chrome - see below).

image

mgwalker:
@jeromeleecms I can't reproduce the weird button behavior in IE. No matter how much I resize the window, it stays in the right place. Can we plan some screen sharing time tomorrow or later in the week?

What do you think about moving the sidebar spacing part over to #1434? It's related to how we emulate stickiness in IE.

jeromeleecms:
Want to see something really weird?

image

Scroll...scroll... poof!

image

...ah so you just popped to the bottom of the screen...

image

jeromeleecms:
@mgwalker Just giving you maybe another data point. It seems like you fixed the jumping save button somehow, but 👍 screen share - during parking lot time? If you think its more clear to merge the the weird stickiness over in #1434, that works for me.. you OK with updating the issue?

mgwalker:
The spacing issue with the sidebar is because of the code in this PR, so it might be worth holding this open to fix that? I _thought_ it was a preexisting bug, but no, I did it. 😬

jeromeleecms:
Maybe it will fix the magically jumping sidebar too? 🤞

mgwalker:
Weirdly, I'm not seeing the issue with the vertical spacing on the sidebar anymore. I don't understand computers.

Issue #1423, opened 2019-04-04
Not all territories have images of their location.


jeromeleecms:
See the screenshot below. 

![image.png](https://images.zenhubusercontent.com/5c547884d11206601037f0c7/2449d7ad-0f5d-4a12-9da1-c6fed3b0f388)

Not necessary to fix now unless one of those states/territories is part of the pilot.

### This task is done when...
 - [ ] All States and Territories have their associated map icon
- [ ] Verified by someone else

mgwalker:
In #1730, we made it so that territories without images wouldn't try to display anything, so the broken image behavior is fixed. Do we want to keep this issue/open a new one to track getting images for the 4 remaining territories?

jeromeleecms:
Let's close this out since the documented bug is complete. I don't think it's a deal breaker if we can't locate images for each territories (nice to have).

PR #353, opened 2018-03-27
add state icons

brendansudol:
Adds script to generate individual state svg files from a USA geojson file; this should allow us to quickly iterate on creating these with update-able sizes / colors / etc.

(It seems like future UI will require us having some state icons of some sort)

Preview:
states

### This pull request is ready to merge when...
- ~Tests have been updated (and all tests are passing)~
- [x] This code has been reviewed by someone other than the original author
- [ ] The experience passes a basic manual accessibility audit (keyboard nav, screenreader, text scaling) OR an exemption is documented
- ~The change has been documented~
  - ~Associated OpenAPI documentation has been updated~

### This feature is done when...
- ~Design has approved the experience~
- ~Product has approved the experience~

PR #1730, opened 2019-07-25
For the 4 territories we don't have images for, don't try to display images

mgwalker:

- for American Samoa, Guam, Northern Mariana Islands, and U.S. Virgin Islands, don't display the state/territory outline in the sidebar because we don't have those outlines (thanks for the idea @hbillings!)
- addresses #1423
  - up to product whether this closes the issue and open a new one later about adding those outline images, but...

The outlines we have now are SVGs generated from TopoJSON files. The original source of the TopoJSON files is murky but [Bren used a minor modification](https://github.com/18F/cms-hitech-apd/pull/353#discussion_r177516462) of [a script by Mike Bostock](https://gist.github.com/mbostock/9821217#file-makefile) to pull data from the USGS and do some magic massaging on it to turn it into nice little blobs that fit into a 64x64 square, which works great for the 50 states plus DC and Puerto Rico, which are included in the original data. The big win here is converting from the USGS geographic projection (whatever that is) into an appropriate state plane projection and computing the right set of scales and translations to make it fit in a box that originates at 0,0.

However, adding the four remaining "state equivalents" (this was a new term to me, but it's fascinating) is difficult because the original USGS dataset doesn't include them, so 1) we'd have to find them and 2) we'd have to figure out how to convert whatever we find into the same thing that the Bostock script does, which is no easy task. I'm not even sure if there are appropriate state plane projections for, e.g., Guam (but probably).

My recommendation is to punt for now until we can find an original dataset that includes the 56 states and equivalents, then we can figure out how to convert those into what we want. Alternatively, if we can find a set of images of all 56 - we don't actually need GeoJSON or TopoJSON or SVGs, just plains PNGs or JPEGs will do fine.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -13,6 +13,7 @@ Anticipated release: TBD
 - Fixed semantic heading levels ([#1695])
 - Fixed a keyboard focus order problem when adding new items to a list ([#1712])
 - Fixed a keyboard focus order problem with the system use banner ([#1715])
+- For users in American Samoa, Guam, Northern Mariana Islands, or U.S. Virgin Islands, do not attempt to display a territory outline at the top of the sidebar because we don't have those outlines ([#1423]; see [#1730] for more information)
 
 #### ⚙️ Behind the scenes
 
@@ -33,6 +34,7 @@ Released: July 9, 2019
 
 Pilot release to select state partners
 
+[#1423]: https://github.com/18F/cms-hitech-apd/issues/1423
 [#1647]: https://github.com/18F/cms-hitech-apd/pull/1647
 [#1651]: https://github.com/18F/cms-hitech-apd/pull/1651
 [#1688]: https://github.com/18F/cms-hitech-apd/pull/1688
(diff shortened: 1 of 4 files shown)


CHANGELOG.md
```diff
@@ -13,6 +13,7 @@ Anticipated release: TBD
 - Fixed semantic heading levels ([#1695])
 - Fixed a keyboard focus order problem when adding new items to a list ([#1712])
 - Fixed a keyboard focus order problem with the system use banner ([#1715])
+- For users in American Samoa, Guam, Northern Mariana Islands, or U.S. Virgin Islands, do not attempt to display a territory outline at the top of the sidebar because we don't have those outlines ([#1423]; see [#1730] for more information)
```
hbillings: This is the kinda stuff I really love changelogs for!

web/src/containers/Sidebar.js
```diff
@@ -197,18 +197,22 @@ class Sidebar extends Component {
...
-                alt={place.name}
-                className="align-middle mr2"
-                width="40"
-                height="40"
-              />
+              {hasImage && (
+                <img
+                  src={`/static/img/states/${place.id}.svg`}
+                  alt={place.name}
+                  className="align-middle mr2"
```
hbillings: While we're in here, can we rip this out? I think this should do the trick (after fiddling with it, it appears that the `align-middle` class doesn't actually have an effect on the way it's centered):
```suggestion
                  className="ds-u-margin-right--2"
```

PR #1731, opened 2019-07-26
Create a component to announce ARIA regions

hbillings:
Because some screenreaders shout at you when non-empty regions are mounted in the DOM, we should have one component that manages regions and their accompanying announcements. This also lays the groundwork necessary for any JSX-controlled ARIA components we might need in the future.

- Creates an `AriaAnnounce` component that loads empty on page load
- Triggers a region announcement on item change 


web/src/app.dev.js
```diff
@@ -22,7 +22,15 @@ const history = createHistory();
 
 const middleware = [thunk, routerMiddleware(history), createLogger()];
 
-const store = createStore(reducer(history), applyMiddleware(...middleware));
+/* eslint-disable no-underscore-dangle */
+const composeEnhancers = window.__REDUX_DEVTOOLS_EXTENSION_COMPOSE__ || compose;
+/* eslint-enable */
```
hbillings: If you don't want to commit this to `app.dev.js`, that's fine by me -- but, man, Redux devtools make things easier!!!

web/src/containers/AriaAnnounce.js
```diff
@@ -0,0 +1,25 @@
+import React, { useState } from 'react';
+import { connect } from 'react-redux';
+
+import { getAriaAnnouncement } from '../reducers/aria';
+
+const AriaAnnounce = (props) =>{
+    // const [ariaMessage, getAriaAnnouncement] = useState(null);
```
hbillings: I was tryin' to be clever but confused myself. Figured I could go back to this after I figured it out the old way. 😅

web/src/containers/activity/CostAllocateFFPQuarterly.js
```diff
@@ -27,8 +28,10 @@ class CostAllocateFFPQuarterly extends Component {
         [year]: { [q]: { [name]: +e.target.value } }
       }
     };
-    const { aKey, update } = this.props;
+    const { aKey, update, announce, quarterlyFFP } = this.props;
     update(aKey, change, true);
+    // ARIA region update goes in here somewhere
+    announce(quarterlyFFP[year][q][name].dollars);
```
hbillings: Either I'm calling the old props, or I'm expecting this to be synchronous when it's async and I need to figure out how not to do that.

web/src/app.dev.js
hbillings: You can thank Mr. Heather for that one -- dinner conversation was me complaining about not being able to get my head around what the state store looks like. This is lifechanging for me:
Screen Shot 2019-07-26 at 9 39 10 AM

web/src/containers/activity/CostAllocateFFPQuarterly.js
hbillings: Thanks so much for the thoughtful response! I actually started down the `Promise` path and felt icky about it, so I'm glad that that was a reasonable gut instinct. 😄 

I like the approach of #2, but I'm also not the hugest fan of making a specific action to deal with individual components' structure because ultimately the ARIA region only cares about a string. I _think_ I figured out how to deal with this in `componentDidUpdate` with generic actions that can be reused anywhere we might need them. Will push it up so we can chat Monday!

web/src/containers/activity/CostAllocateFFPQuarterly.js
```diff
@@ -190,12 +199,14 @@ CostAllocateFFPQuarterly.propTypes = {
 const makeMapStateToProps = () => {
   const selectCostAllocateFFPBudget = makeSelectCostAllocateFFPBudget();
   const mapStateToProps = (state, props) =>
+    ariaQueuedMessage: getAriaQueuedAnnouncement(state),
```
hbillings: This approach seems like it'll work, but I can't test because this fails -- I can't figure out how to properly merge it with the object that `selectCostAllocateFFPBudget` returns.

web/src/actions/aria.js
```diff
@@ -1,5 +1,13 @@
+export const ARIA_QUEUE_ANNOUNCEMENT = 'ARIA_QUEUE_ANNOUNCEMENT';
 export const ARIA_ANNOUNCE_CHANGE = 'ARIA_ANNOUNCE_CHANGE';
 
+export const ariaQueueAnnouncement = change => dispatch => {
+    dispatch({
+        type: ARIA_QUEUE_ANNOUNCEMENT,
+        change
+    });
+}
```
hbillings: @mgwalker What do you think about something like this? This would let us queue whatever ARIA announcement we might need, and do the logic of converting the announcement in the component instead of in a bunch of separate actions. (Although now that I've written that out, it occurs to me that maybe separate actions is a more React-y way of doing this than using `componentDidUpdate`?)

mgwalker: It feels a little odd to have an action that is basically a component telling Redux to remind the component to do something later.  My gut tells me that distinct actions would be cleaner. 😕

hbillings: That's a fair criticism; it's messier this way than I was anticipating it might be. It seems like it can be DRY or it can be clean but it can't be both, which kinda sucks. (Or, at least, I'm not clever enough to figure out how to get it to be both.) Thanks for indulging me!

web/src/containers/activity/CostAllocateFFPQuarterly.js
```diff
@@ -27,8 +45,10 @@ class CostAllocateFFPQuarterly extends Component {
         [year]: { [q]: { [name]: +e.target.value } }
       }
     };
-    const { aKey, update } = this.props;
+    const { aKey, update, queue } = this.props;
     update(aKey, change, true);
+    // queue the change for AriaAnnounce
+    queue(change);
```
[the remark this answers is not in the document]
hbillings: I should have prefaced that bit with a big "DON'T LOOK AT THIS I HAVEN'T MADE IT PRETTY" warning 😆

hbillings:
@mgwalker All right, sir, I think this is ready for final review!

web/src/containers/AriaAnnounce.test.js
```diff
@@ -0,0 +1,18 @@
...
+
+describe('aria announce component', () => {
+    it('renders correctly', () => {
+        const component = shallow(<AriaAnnounce />);
+        expect(component).toMatchSnapshot();
+    });
+
+    it('renders a message', () => {
+        const props = {
+            ariaMessage: 'No. No. No. No.'
```
mgwalker: laughing

web/src/actions/aria.js
```diff
@@ -0,0 +1,8 @@
+export const ARIA_ANNOUNCE_CHANGE = 'ARIA_ANNOUNCE_CHANGE';
+
+export const ariaAnnounceFFPQuarterly = (aKey, year, q, name) => (dispatch, getState) => {
+    dispatch({
+        type: ARIA_ANNOUNCE_CHANGE,
+        message: getState().budget.activities[aKey].quarterlyFFP[year][q][name].dollars
```
mgwalker: Should we open an issue for figuring out exactly what this message should be? Right now it reads the new quarterly FFP total but doesn't tell you that's what it is. I feel like that satisfies the original issue but now it's about figuring out what information is useful to the user?

hbillings: Ooh, good call. Yeah, without the `Dollars` component, this doesn't even tell you it's a dollar amount.

web/src/containers/AriaAnnounce.test.js
hbillings: thanku

PR #1733, opened 2019-07-26
Scroll activities and form/review components into view when they're collapsed

### LINKED ISSUE #1732: Scroll the activity review into view when collapsing an activity
Currently when you collapse an activity, the page scroll stays in the same place, but the total height of the page shrinks significantly, meaning that you are effectively scrolled way down the page. The practical upshot is that when you collapse an activity with the "Done" button at the bottom, you're often left staring at the submission instructions and you have to scroll way back up the page to get to the next activity.

Instead, we should scroll the page so that the collapsed activity component is in view. That seems like it'll be less jarring than dropping the user in the middle of nowhere with no clear indication of what just happened or how to get back on track.

### This task is done when...
 - [ ] activity review component is visible after collapsing an activity form

mgwalker:

- if an activity review component is not in view when the activity is collapsed, scroll it into view
- same logic for form/review components
- closes #1732 


web/src/components/FormAndReviewList.js
```diff
@@ -8,20 +8,27 @@ const FormAndReviewItem = ({
   initialCollapsed,
   ...rest
 }) => {
+  const container = useRef(null);
   const [collapsed, setCollapsed] = useState(initialCollapsed);
-  const collapse = useCallback(() => setCollapsed(true), []);
+  const collapse = useCallback(() => {
+    const { top } = container.current.getBoundingClientRect();
+    if (top < 0 || top > window.innerHeight) {
```
mgwalker: This is just to make sure the target isn't already in view. The browser doesn't check for us, and will scroll the element to the top of the viewport. I thought we probably didn't want to scroll the page to the top if the target was already in view - that'd be especially jarring.

mgwalker:
If you scroll to the bottom of an activity and click the "Done" button, when it collapses, the review pane for the activity you just closed should be at the top of the screen.  If you collapse the activity using the "Close" button in the activity header, then nothing will happen because the activity review is already on the screen.

Happy to hop on a screen share sometime if that'd be more helpful!

lauraponce:
Hm, I must be missing something still because I don't think I'm seeing this. When I click done, it jumps down to the Executive Summary section which I don't think is what you're explaining. Maybe during parking lot time tomorrow after standup, we could steal a couple minutes to walk through this?

hbillings:
What do we need to do to nudge this one over the finish line?

mgwalker:
@hbillings I wanted to look into the focus thing to see what it'd take, but maybe we should create an issue for that instead? Though if it's an a11y issue, I think we'd agreed that fixing those would be part of the definition-of-done for a PR, so maybe we can't merge this in that case?

PR #1734, opened 2019-07-26
extend user sessions on API activity

### LINKED ISSUE #1728: Backend sessions are never extended
When a user logs in, the backend creates a session for them. That session is given an expiration time based on the `SESSION_LIFETIME_MINUTES` environment variable. When the expiration time comes, the session is invalidated and the user must log in again.

When the user interacts with the backend, their session should be extended back to the full lifetime. However, that is not happening, so users are forcibly logged out 48 hours after logging in. Instead, user sessions should be extended back to a full lifetime.

### This task is done when...
 - [ ] user backend sessions are extended on API activity

mgwalker:

- extends a users's backend session on every API access
- closes #1728

  This one is tricky to test because a backend session is valid for 48 hours at a time. I guess if you log in on Monday, then open the site on Tuesday and Wednesday, if you're still logged in on Thursday, it worked!


jeromeleecms:
Logged in on 07/30. Will try to login on 07/31 and 08/01. Will try and report back on 08/02.

jeromeleecms:
@mgwalker Logged in on 07/30 - returned on 08/01 to site, but was logged out of the system. That seems to be the expected behavior based on our time out settings. I'm not sure how to verify that's the right behavior that I'm actually extending the API backend?

mgwalker:
@jeromeleecms The way to test it is to login and note the time. Then in less than 48 hours, open it up - that _should_ extend your session.  Finally, just over 48 hours after you initially logged in (but less than 48 hours after you visited the second time!), open it again. If you're still logged in, it worked! If you're not... then it didn't... :sadface:

hbillings:
What do we need to do to close this one?

mgwalker:
@hbillings I think we couldn't test it because of #1756. I'm _pretty_ confident this does what it's supposed to, but we can't be sure yet. I'm content to merge it now and double-check when #1756 goes in, though. What do you think, @jeromeleecms?

PR #1744, opened 2019-08-01
ARIA region announcement shouldn't be visible

hbillings:
@mgwalker discovered I forgot to actually hide the ARIA region from #1731 from sighted users. This does that.


mgwalker:
Ooh. Can you update the changelog before merging? It doesn't have the original bug in it yet.

PR #1746, opened 2019-08-02
Add a link to skip navigation and go straight to the main content

### LINKED ISSUE #1304: Using a screen reader, there is no option to "Skip to main content" when tabbing through pages
When tabbing through the content, there is no option to "Skip to main content." Instead user must navigate through the sidebar before getting to main content.

### This task is done when...
 - [ ] "Skip to main" content is available in each portion of the app (login, dashboard, app builder)

mgwalker:

- adds a link at the top of the page, only visible when focused (e.g., if the user tabs to it), that lets the user skip the navigation and go directly to the main content
- closes #1304

### DIFF
--- a/web/src/components/Wrapper.js
+++ b/web/src/components/Wrapper.js
@@ -22,7 +22,7 @@ class Wrapper extends Component {
     const showSiteTitle = document.location.pathname === '/';
     return (
       <div className="site">
-        <a href="#apd-title" className="skip-nav">
+        <a href="#start-main-content" className="skip-nav">
           Skip to main content
         </a>
         <Header showSiteTitle={showSiteTitle} />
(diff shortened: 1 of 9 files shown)


web/src/components/Wrapper.js
```diff
@@ -22,11 +22,14 @@ class Wrapper extends Component {
     const showSiteTitle = document.location.pathname === '/';
     return (
       <div className="site">
+        <a href="#apd-title" className="skip-nav">
```
hbillings: This ID looks like it only exists on the APD page itself, but the "Skip to main content" link is on every page with a header (so login, admin pages, etc.) Does it make sense to instead give the main `h1` on every page the same ID? Or maybe the container that wraps our main content on every page?

mgwalker: smart

PR #1750, opened 2019-08-05
Show a more meaningful save error if the user is logged out

### LINKED ISSUE #1729: If save fails because the user is logged out, tell them so instead of using the generic save failed message
If the user is logged out, the returned status code is 403. Any other failure would be a 400, 401, or 404. If a save fails and the response code is 403, we can display a different message to the user so they have a better understanding of how to fix the problem.

(Related to #1728, in that once #1728 is fixed, the user should only run into this issue if the frontend idle timeout is longer than the backend timeout, which would be a misconfiguration. Still, worth adding in, I think.)

### This task is done when...
 - [ ] attempting to save while logged out shows a different error message

mgwalker:
We can detect if the save fails due to the user being logged out, so we should give them a more useful error message.

- give them a more useful error message
   Screen Shot 2019-08-05 at 12 09 03 PM
- closes #1729 

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -22,6 +22,7 @@ Anticipated release: TBD
 - Removed the emoji button from the rich text editor ([#1649])
 - Adds spacing around the Medicaid director and Medicaid office address form headers ([#1646])
 - Adds margin to the bottom of the state dashboard ([#1601])
+- Adds an ARIA region component to prevent screen readers from prematurely announcing quarterly budget numbers ([#1731])
 
 #### ⚙️ Behind the scenes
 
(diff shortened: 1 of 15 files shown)


mgwalker:
@jeromeleecms Yes, that'd be a good long-term change! There are some internal architectural changes we'd need to make to support it, but the user experience would definitely be better.

quarterback:
@jeromeleecms @mgwalker I'll propose another mod of the text for brevity and consistency with other text. _"**Save failed**: You are logged out. Use another tab to login, and try saving again."_

mgwalker:
Text changed to "Save failed: You are logged out. Use another tab to login, and try saving again." and moved into the yaml (`errors.yaml`, `save-apd.not-logged-in`)

mgwalker:
That's the expected behavior. If there aren't any changes in the first tab, it won't actually try to save anything to the API and instead just says it was successful.

PR #1755, opened 2019-08-06
Select numeric field content on focus if the current value is 0

### LINKED ISSUE #1736: Placeholder values should disappear when clicked into
As pointed out in one of our shadowing sessions, when clicking into the input field with a placeholder, it should clear out when the field is in focus. 

For example:

### This task is done when...
 - [ ] The percent commitment to the project field placeholder value disappears when in focus

### LINKED ISSUE #1754: Cost allocation form "other funding" input is not a dollar field
On the activity cost allocation form, we ask for "other funding amount" using a design system `TextField` instead of our `DollarField` wrapper. That means we could end up with invalid values, and our fix for #1739 won't carry into this field.

### This task is done when...
 - [ ] the activity cost allocation form is updated so that the "other funding amount" is a `DollarField`

mgwalker:
Number fields whose current value is zero will have their contents selected when the form field is focused. That way when a user clicks in that field, they can just begin typing and overwrite the zero that's already there.

- added `NumberField` component
- updated `DollarField` to be a `NumberField`
- updated APD key personnel, incentive payments, state personnel, contractor resources, and cost allocation forms to use the `NumberField` component
- updated cost allocation form to use `DollarField` component for "other funding" (see #1754); incidental - in order to fix #1736 across the app, #1754 needed to be fixed too... so I went ahead and did it
- closes #1736
- closes #1754 

### DIFF
--- a/web/src/components/NumberField.js
+++ b/web/src/components/NumberField.js
@@ -1,0 +1,24 @@
+import { TextField } from '@cmsgov/design-system-core';
+import React, { useEffect, useRef } from 'react';
+
+const selectTextIfZero = ({ target }) => {
+  const value = +target.value.replace(/[^0-9]/g);
+  if (value === 0) {
+    target.select();
+  }
+};
+
+const NumberField = ({ ...props }) => {
+  const textField = useRef(null);
+
+  useEffect(() => {
+    textField.current.addEventListener('focus', selectTextIfZero);
+    return () => {
+      textField.current.removeEventListener('focus', selectTextIfZero);
+    };
+  }, []);
+
+  return <TextField {...props} fieldRef={textField} />;
+};
+
+export default NumberField;
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -26,6 +26,7 @@ Anticipated release: TBD
 - Adds an ARIA region component to prevent screen readers from prematurely announcing quarterly budget numbers ([#1731])
 - Fixed alignment of the message if there are no APDs on the state dashboard ([#1602])
 - Adds spacing between the login form and the "forgotten password" help link ([#1600])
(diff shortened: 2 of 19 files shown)


web/src/components/NumberField.js
```diff
@@ -0,0 +1,24 @@
+import { TextField } from '@cmsgov/design-system-core';
+import React, { useEffect, useRef } from 'react';
+
+const selectTextIfZero = ({ target }) => {
+  const value = +target.value.replace(/[^0-9]/g);
+  if (value === 0) {
+    target.select();
```
hbillings: ooh

web/src/components/NumberField.js
```diff
@@ -0,0 +1,24 @@
...
+  }
+};
+
+const NumberField = ({ ...props }) => {
+  const textField = useRef(null);
+
+  useEffect(() => {
```
hbillings: Even though I get Hooks better now, I can't really get my head around `useEffect` or when it fires. (Which will probably change soon, because I think it's what needs to happen with the `DateField` error.) So... very nice magic you have here!

mgwalker: `useEffect` is run ***after*** every render of the component. But hooks can be memoized to prevent them running _every_ time. The signature for `useEffect` is:

```js
useEffect(functionToRun, [list, of, variables, to, memoize, on]);
```

The hook will only run if one of those variables has changed. If you leave out the list of variables, it will run every time. If you pass it an empty array, it will only run the first time the component is rendered and then never again. This way, you can mimic the behavior of `componentDidMount`. The major difference between `useEffect` and `componentDidMount` is when they run - `componentDidMount` runs before render, so the `ref` here would be unset; `useEffect` runs *after* render, so `ref` is ready to go.

PR #1760, opened 2019-08-08
Make a copy of the Choice component to remove the aria-live attribute

hbillings:
In the final ARIA announcement saga episode, we create a wrapper around the design system's Choice component that brute-force removes the `aria-live`, `aria-atomic`, and `aria-relevant` attributes.

**TODO** Open a ticket with the design system folks asking them to make the aria region optional via a component flag.

- ☝️ 

### DIFF
--- a/web/src/components/Choice.test.js
+++ b/web/src/components/Choice.test.js
@@ -1,0 +1,84 @@
+import { mount, shallow } from 'enzyme';
+import React from 'react';
+
+import Choice from '../components/Choice';
+
+
+
+describe('Choice wrapper component', () => {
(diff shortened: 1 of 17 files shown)


web/src/components/Choice.test.js
```diff
@@ -0,0 +1,82 @@
...
+            <p>I am the child element</p>
+          </div>
+        }
+      >
+        I have checkedChildren
+      </Choice>
+    );
+   
+    const child = component.find('.ds-c-choice__checkedChild').render();
+    expect(child[0].attribs['aria-live']).toBeFalsy();
```
mgwalker: that's good

PR #1764, opened 2019-08-09
[HOTFIX] Update contractor hourly data structure when APD years change

### LINKED ISSUE #1762: Cannot edit default contractor resource after changing APD years
Clicking the "edit" link on a contractor resource causes the page to go blank. Users cannot proceed until they reload the page.

### Reproduce
1. Open an APD
2. Change the fiscal years for the APD
3. Attempt to edit a contractor resource

### Workaround
1. Create a new contractor resource
2. Delete the offending contractor resource

### This task is done when...
 - [ ] users can edit contractor resources after changing APD fiscal years

### LINKED ISSUE #1765: Emptying date fields prevents saving
If the user enters a date and then backs it out so that the date is empty, the APD no longer saves.

### This task is done when...
 - [ ] users can save if the date is empty

mgwalker:

- When you change the federal fiscal years that an APD applies to, we need to also update the state personnel, contractor resource, and non-personnel expense data structure to remove or add the years that were removed/added from the APD. Contractor resources includes _two_ cost structures - yearly total costs, and hourly costs by year. We were not updating the hourly costs by year, resulting in the app trying to access a data structure that did not exist. This PR updates the code to correctly update the contractor resource hourly costs data structure.
- When a date field is given a value and then you empty the date back out, the APD can no longer save. This is because of how we format data from the design system date field - an initial empty value stays empty, but once a user enters something, the date is _always_ formatted as `yyyy-mm-dd`. If you back out the year, month, and date, the result is `--`, which is not valid. This PR updates the code to produce an empty string if the year, month, and day are empty.
- fixes #1762
- fixes #1765


web/src/containers/activity/ContractorResource/ContractorResourceForm.js
```diff
@@ -32,29 +32,17 @@ const ContractorResourceForm = ({
 
   const handle = {
     changeDesc: useCallback(handleChange(index, 'desc'), []),
-    changeHourlyHours: apdFFYs.reduce(
-      (acc, ffy) => ({
-        ...acc,
-        [ffy]: useCallback(handleHourlyChange(index, ffy, 'hours'), [apdFFYs])
-      }),
```
mgwalker: This is a bug. The number and order of hooks needs to be deterministic - it can't change between renders, or else React loses track of what it's supposed to return. Here, the number of callbacks is based on how many federal fiscal years the APD has. If you add a year, that increases the number of hook calls, and React throws an exception for that.

hbillings: whoa

PR #1770, opened 2019-08-20
Remove old Ace/Bass styles

### LINKED ISSUE #1703: Site doesn't respect browser text size settings
Setting text size at the browser level doesn't change app text size.

Tested in Chrome.

hbillings:
Closes #1703 by removing all of the old styles (which hardcoded the HTML font size to 16px, sidestepping a user's specific browser settings).

- Removes all the old CSS
- Edit some of the styles that were still in use to use design system instead
- Copy over the few styles still in use


docs/frontend-build.md
```diff
@@ -30,7 +30,6 @@ files copied from static sources.
 | index.html | Entry page, copied directly from `src/index.html`, unchanged                    |
 | app.js     | The React app and all necessary libraries (entrypoint is `src/app.js`)          |
 | app.css    | Our primary styles, built from Sass (entrypoint is `src/styles/index.scss`)     |
-| legacy.css | Our legacy styles (entrypoint is `src/styles/legacy.css`)                       |
```
mgwalker: shoo

web/src/containers/Sidebar.js
```diff
@@ -202,8 +202,8 @@ class Sidebar extends Component {
     return (
       <div className="ds-l-col--3">
         <aside className="site-sidebar">
-          <div className="xs-hide sm-hide site-sidebar__sticky">
-            <div className="flex items-center ds-u-border-bottom--1 ds-u-padding-y--2 ds-u-margin-bottom--4">
+          <div className="site-sidebar__sticky">
```
mgwalker: Is it okay that the sidebar no longer disappears on narrow screens? My guess is this is a usability improvement because previously it went away entirely and there was no longer any way to quickly navigate. Maybe there should be a new issue to figure out how to handle the sidebar on small screens?

Question for @lauraponce maybe?

hbillings: Yeah, we should probably have issues to figure out how we want to deal with small screens. Right now, nothing really works well on a small screen (and I think there's an argument to be made that most of this doesn't need to work on a small screen).

PR #1771, opened 2019-08-21
Fix vertical spacing in activity overview section

### LINKED ISSUE #1648: Fix vertical spacing in activity details, activity overview
There is too much spacing after **Activity Overview** and the beginning of the form. This screenshot is from before the activities were switched to the form/review pattern, but it accurately depicts the problem:

screenshot showing excess vertical spacing

### This task is done when...
 - [ ] spacing is reduced
 - [ ] design approves

(from #1638)

mgwalker:

- makes the spacing around the activity overview subsection heading match the spacing in the activity schedule subsection
- closes #1648 


web/src/containers/activity/Overview.js
```diff
@@ -88,7 +88,11 @@ const Description = props => {
   );
 
   return (
-    <Subsection resource="activities.overview" nested>
+    <Subsection
+      headerClassName="header-with-top-margin"
```
mgwalker: I like using code words so it's hard to understand what things are... 😬

hbillings:
I looked at this but didn't have time to completely wrap my head around the changes I think we should make here. My gut feeling is that there's a way to do this that might be a little bit more seamless, such as targeting by position or targeting first-children. That would be in line with the design system utility classes and BEM naming in general. Regardless, I applaud @mgwalker for plunging feet-first into this!

PR #1808, opened 2019-09-10
Prevent sidebar from growing and shrinking in IE

### LINKED ISSUE #1434: Inconsistent Stickiness Sidebar Behavior with Internet Explorer

mgwalker:

- sets the sidebar to a fixed width on IE, so it doesn't get wider and narrower and weird stuff
- closes #1434


mgwalker:
This creates an `isIE()` utility function, and then calls that in the sidebar code to set a conditional class when we're running in IE. @hbillings I'm wondering if you think it'd make sense to set a conditional class on the `<body>` when we're running in IE instead, and we could separate out IE-specific hacks in our Sass?

mgwalker:
slapping forehead

Yep, this happens when the browser is narrow enough that the sidebar's native width is _less than_ the fixed width. This is so frustrating! I'm not sure what to do about that - my knee-jerk reaction is to add some Javascript to compute the proper width, but that's so gross. I'm going to wait until next week and ask Heather because she's smart.

hbillings:
MEDIA QUERIES!

But, also, I'm wondering if instead of setting the fixed width on the sidebar itself, if we could set it on the interior items. Or maybe on both. What's happening is that when `position: fixed` is applied, the width calculation changes (which is why @jeromeleecms has been seeing the width change -- `position: fixed` isn't applied in the initial scroll position, but it is once the page is scrolled down). If we set the width on a child of the container that gets `position: fixed` applied to it, I think we'd maybe see more predictable behavior. I'll try to get my IE setup working tomorrow and try a few things out.

web/src/containers/Sidebar.js
```diff
@@ -3,18 +3,14 @@ import React, { Component } from 'react';
 import { connect } from 'react-redux';
 
 import VerticalNav from '@cmsgov/design-system-core/dist/components/VerticalNav/VerticalNav';
-import stickybits from 'stickybits';
 
 import { t } from '../i18n';
 import { jumpTo } from '../actions/navigation';
+import { isIE } from '../util';
```
hbillings: We don't need this here any more, right?
```suggestion
```

web/src/containers/Sidebar.js
```diff
@@ -200,28 +196,24 @@ class Sidebar extends Component {
...
-                  height="40"
-                />
-              )}
-              <h1>{place.name}</h1>
-            </div>
-            <VerticalNav
-              selectedId={activeSection || 'apd-state-profile-overview'}
-              items={links}
-            />
+      <aside className={`site-sidebar${isIE() ? ' site-sidebar__ie' : ''}`}>
```
hbillings: I think we can just do...
```suggestion
      <aside className="site-sidebar">
```

web/src/util/index.js
```diff
@@ -267,3 +267,14 @@ export const replaceNulls = (obj, newValue = '') => {
   replace(objNew);
   return objNew;
 };
+
+/**
+ * Discover whether the current browser is a version of Internet Explorer
+ * @returns {Boolean} True if the browser is IE, false otherwise. Returns false
+ *                    for Edge.
```
hbillings: Can we get rid of this also, or does the save button need this?

PR #1810, opened 2019-09-11
Add ARIA attributes to the account management button to make it more accessible

### LINKED ISSUE #1681: Navigation dropdown is vague for screen readers
There's nothing announced when the screen reader hits the user email that would suggest that it's a) a dropdown, b) the parent item for account-related actions. 

Screen Shot 2019-07-05 at 3 47 23 PM

A good resource for thinking about how to attack this: https://www.smashingmagazine.com/2017/11/building-accessible-menu-systems/

### This task is done when...
 - [ ] We've identified how the nav should act for screen readers
 - [ ] We've implemented the same

mgwalker:

- sets a more descriptive ARIA label for screen readers: "You are logged in as [account]. Click to manage account."
  - Do we need the "click" part? The `aria-haspopup` attribute will indicate that it has a popup, but it doesn't give any clue as to what's hidden in the popup.
- sets `aria-haspopup="true"` so screen readers know the button controls a popup
- moves `aria-expanded` to the button instead of the containing UL, so the reader will announce the expand/collapse state when the button gets focus
- closes #1681 

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -36,6 +36,7 @@ Anticipated release: TBD
 - Round off dollar input fields when they lose focus ([#1739])
 - Fixed a bug where multiline plain text fields (such as the activity overview) gets exported as a text field, and the content is truncated within it. ([#1767])
 - Fixed a bug where the session authentication cookie would expire at the end of the browser session instead of at the scheduled time. ([#1756])
+- Improved ARIA metadata on the account management dropdown button for screen readers ([#1681])
 
 #### ⚙️ Behind the scenes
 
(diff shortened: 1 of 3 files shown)


jeromeleecms:
@hbillings Were you able to verify that the screen reader picks up the notification that the item contains a dropdown? When I highlight it with Windows Narrator, it reads the text to me and that's it. If it works for you, I'm OK to approve this.

mgwalker:
I gave up trying to get Windows Narrator to work at all for me - maybe we could pair sometime so you can show me how to get it working? 😛

I know IE11 will never announce that it's a dropdown because that attribute was introduced in ARIA 1.1 after IE11 was frozen, but Chrome should work. It worked for me using macOS Voiceover, but I'd want to hear from someone else. I'd also like to see if we can figure out what's going on with Narrator. The interplay between browsers and screen readers is weird, so I guess it's possible Narrator doesn't support the `haspopup` attribute?

jeromeleecms:
Parking lot time? My last computer read some of the text (default settings), but my replacement computer now reads way more - so besides getting it working, maybe we can figure out what is going on.

hbillings:
So I _think_ it's okay if it doesn't mention there's a dropdown if the instruction is clear from the text that's read. What I do find a little weird is that, at least on Safari/VoiceOver, you have to use the arrow keys to select the menu options rather than the tab key. But at this point, we're beyond my ability to guess the right interaction. 😬

Issue #1408, opened 2019-04-01
Decide on long-term replacement for our rich text editor


NAretakis:
Per #1127, this issue is for us to investigate and determine a long-term solution for our rich text editor. 
and make an informed switch to a different rich text editor library:
   - **pros:** can pick something that is actively maintained
   - **cons:** might not find something with all the same features
   - might be complicated to retrofit

This issue is for a research spike of the above to see what's out there, what features we'd lose/gain, and the effort to replace it.  The result of that spike could inform a decision.

### This task is done when...
 - [ ] Various rich text editor options are researched
 - [ ] A rich text editor is selected to implement based on our research

jeromeleecms:
As part of our research, have we considered COTS and open-source options?

mgwalker:
Most of the libraries I can find rely on using a `contenteditable` node rather than an `<input>` or `<textarea>`, so they will all have the same IE spellcheck problem.

- #### [Draft.js](https://draftjs.org/)

  This is the library underlying the editor we use. It seems to be the most maintained (it's a Facebook product, and it's been updated in the past week) and also the most used. One advantage of using a Draft-based editor is that the rich text fields we've already stored in the database would still be valid. If we switch to another library, we might have to figure out how to transform the text we've stored into the new library's format. The downside is that Draft.js is not itself a complete rich text editor - it's a framework for building one. So we'd still have to pick an actual editor:

  - [react-draft-wysiwyg](https://github.com/jpuri/react-draft-wysiwyg) - This is the one we use now. We know there is a performance problem with multiple editors on the screen at once (that's why we have our own fork), and we also know it isn't being actively maintained anymore.

  - [Megadraft](https://www.npmjs.com/package/megadraft) - Not widely used, but updated within the past month or so, but there are a bunch of very old issues still sitting out there open.

  - Custom implementation - Given the number of articles on the web about building a rich text editor using Draft.js, I have to wonder if maybe it's easy enough that we _should_ just do our own. The trouble is that it's not a small amount of code that would have to be maintained.

- #### [Quill](https://quilljs.com/)

  Quill also seems to be well-maintained and widely used, but it doesn't claim to support IE at all. It might work, but if anything goes wrong, we won't get any help. Similar to Draft.js, it isn't an editor in a box, but it has a bunch of plugins that you can compose together to create a full editor. That is, we wouldn't need a 3rd-party built on top of Quill, we'd just use different Quill pieces.

[...]

PR #1811, opened 2019-09-11
Enable spellcheck for rich text editor

### LINKED ISSUE #1769: Add spellcheck to rich text editor
The rich text editor doesn't support spellcheck. Fix that.

### This task is done when...
 - [ ] the rich text editor supports spell check

mgwalker:

- just what it says on the tin
- closes #1769

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,7 @@ Anticipated release: TBD
 - Attempt to alert users before automatically logging them out due to inactivity. The app will try to use built-in browser notifications as well as flashing the tab title. In browsers that support it, the app treats activity in any eAPD tab as valid, so all eAPD tabs will remain valid as long as at least one of them is getting activity. ([#1697])
 - Scroll the collapsed activity review panel into view after collapsing an activity form. This way, when you collapse an activity, you end up essentially looking at the list of activities again instead of being pushed way down the page. ([#1732])
 - Automatically select numeric form field contents when the field is focused if the current value is 0 ([#1736])
+- Add spellcheck to rich text fields ([#1769])
 
 #### 🐛 Bugs fixed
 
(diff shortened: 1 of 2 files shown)


mgwalker:
Hmm. It _should_ work in IE. The `spellcheck` attribute that this relies on is supported in IE 11. Let me dig into it a bit, but I can definitely replicate it _not_ working in IE.

mgwalker:
It looks like [Draft.js explicitly disables spellcheck in IE 11 and below](https://github.com/facebook/draft-js/issues/640) because of a bug in IE. In a nutshell, IE doesn't provide an event if you right-click the misspelled word and select a correction, so Draft.js doesn't know that anything changed. Then there's a mismatch between what's displayed on the screen and what Draft.js thinks the text is.

For example, let's say you've got this:

> This is text asdf

Conveniently, `asdf` is a typo, so if we enable spellcheck, it'll get a squiggly line. You can right-click on that and choose a corrected spelling. Let's say you choose `ads`. Now this is what you see:

>This is text ads

But IE doesn't send an event, so internally, Draft.js still thinks the text is `This is text asdf`. If you click after `ads` and start typing, IE will start sending events again. That event tells Draft.js _where_ in the text to add letters. So for example, if you type ` abc` at the end, this is what you have visually:

> This is text ads abc

But internally, this is what Draft.js has:

> This is text asd abcf

The ` abc` got inserted at the _same position_ in Draft.js's internal text as it did in the visual text, but the two have different text. Oh no! The visual text will be updated to match Draft.js's internal text whenever the text box is unfocused, and then you could go in and manually fix it, but then what was the point of spellcheck?

Unfortunately, there's simply no solution to this problem. Draft.js made the decision to completely disable spellcheck on IE. The only alternative is to beg users not to use the correction feature in IE and instead manually fix the things that spellcheck identifies.

For usability purposes, I think I have to suggest that we leave it disabled on IE, and I genuinely don't think there's a viable workaround. (There are deeper DOM events that seem to be triggered when the user corrects a spelling with IE's built in right-click thingy and if it's critical, we could investigate whether we could hook those and simulate events into Draft.js to get it to work properly, but I have no idea what that entails offhand.)

jeromeleecms:
OK - I propose we just merge this and then look to our longer term rich text editor (#1408) to incorporate spell check for multiple browsers. That work?

mgwalker:
[It looks like Edge also doesn't fire the right events](https://github.com/facebook/draft-js/issues/990).

PR #1814, opened 2019-09-12
Fix Sass warnings

### LINKED ISSUE #1774: Sass build warnings
When the app builds, the Sass step produces some warnings about missing variables without fallbacks. For example:

screenshot showing warnings

The build does complete and everything seems to be working fine, but we should fix the warnings.

### This task is done when...
 - [ ] warnings are fixed

mgwalker:
We had rounded borders applied to some parts of the rich text editor, but those borders were based on a radius variable that was set in the legacy CSS index file. When that file went away, the variables were no longer defined, so Sass would give us a warning. Since nobody seemed to notice or mind that the borders were no longer rounded, I opted to just remove them rather than put the variable back.

- removes references to `--border-radius` variable in scss
- closes #1774

- ~Changelog is updated as appropriate~ Not necessary. This is part of removing the old styles, which is already in the changelog

### DIFF
--- a/web/src/styles/scss/editor.scss
+++ b/web/src/styles/scss/editor.scss
@@ -5,7 +5,6 @@
 .rdw-editor-toolbar {
   background-color: #f5f5f5;
   border: 0;
-  border-radius: var(--border-radius);
 
   & * {
     box-sizing: content-box;
@@ -15,7 +14,6 @@
 .rdw-dropdown-wrapper,
 .rdw-option-wrapper {
   border: 1px solid #eee;
-  border-radius: var(--border-radius);
 }
 
 .rdw-emoji-modal {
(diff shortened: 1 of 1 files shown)


hbillings:
I wonder if, instead of just removing the old variable, it would make sense to replace it with the `$border-radius` variable that the other form fields use? That way we'll maintain consistent radii throughout.

mgwalker:
Perhaps! We could ask Laura to weigh in? I chose to just take them out because that wouldn't have any affect on the visuals, but maybe it was a regression when the rounded borders stopped working in the first place. 😬

hbillings:
@mgwalker We do still have rounded borders! They're just 3px rounded instead of IIRC 6px. Oddly, when I pulled this branch down and recompiled the Sass, it still showed the correct 3px, but I couldn't for the life of me figure out where it was coming from, since there's no actual input box in the RTE. Maybe we can merge this as-is, and if the RTE ends up showing square corners, we can fix it then?

mgwalker:
How very odd... Maybe it's built into the RTE's stylesheets? Anyway, I like that plan.

PR #1815, opened 2019-09-12
Add Percent field, switch APD key personnel and quarterly FFP share inputs to Percent field

### LINKED ISSUE #1753: APD key personnel asks for percentage of time, but entry field is a plain text input
We should mimic the percent input field from the quarterly FFP share table, where we include a percent sign at the end. It'll require either duplicating some CSS or refactoring the classes a bit (the latter makes more sense to me). We could also consider adding a `<PercentField>` component, but I think there are only these two instances of it so maybe better to just leave it for now.

### This task is done when...
 - [ ] the APD key personnel "percent commitment to program" input field has a percent sign mask

mgwalker:
Adds a `<PercentField>` component that does DOM jujitsu to insert a percent sign into number input fields. I had to use gymnastics here because we're using design system components, and they render the `<input>` deep inside some other stuff. That doesn't seem like a problem though, right? We can target that input with CSS!

Well, yes we can, but did you know that input elements don't have `:before` or `:after` pseudoelements? So you can't put a percent sign next to them that way, which is the most obvious way.

Okay, but we could target the container and put the percent sign on that! Sure, that's possible. But the container is the full content width and the input element is not. If we want the percent sign to be nestled against the right side of the input element...  well, tough luck, there's no way in CSS to position the wrapper's `:before` pseudoelement relative to one of the wrapper's children.

So the somersaulting solution: grab the input element after it has been rendered, programmatically create our own wrapper div, put our custom wrapper in place of the input element, and add the input element into our wrapper.  Then we relatively position the wrapper and, the key step, we set its display to `inline-block`. Now the wrapper has the same size as the input element, and we can position the `:before` pseudoelement relative to the wrapper itself.

- adds `<PercentField>` to APD key personnel and updates quarterly FFP share table
  image
- closes #1753

### DIFF
--- a/web/src/styles/scss/forms.scss
+++ b/web/src/styles/scss/forms.scss
@@ -77,7 +77,7 @@ div.form--with-reviews {
     font-weight: $font-normal;
     font-size: $lead-font-size;
     position: absolute;
-    right: 8px;
+    right: $spacer-1;
     top: 12px;
   }
 
--- a/web/src/styles/scss/forms.scss
+++ b/web/src/styles/scss/forms.scss
@@ -74,7 +74,7 @@ div.form--with-reviews {
   position: relative;
   &:after {
     content: '%';
-    font-weight: 700;
+    font-weight: 300;
     font-size: 18px;
     position: absolute;
     right: 8px;
--- a/web/src/styles/scss/forms.scss
+++ b/web/src/styles/scss/forms.scss
@@ -74,7 +74,7 @@ div.form--with-reviews {
   position: relative;
   &:after {
     content: '%';
-    font-weight: 300;
+    font-weight: $font-normal;
     font-size: 18px;
     position: absolute;
     right: 8px;
--- a/web/src/styles/scss/forms.scss
+++ b/web/src/styles/scss/forms.scss
@@ -75,7 +75,7 @@ div.form--with-reviews {
   &:after {
     content: '%';
     font-weight: $font-normal;
-    font-size: 18px;
+    font-size: $lead-font-size;
     position: absolute;
     right: 8px;
     top: 12px;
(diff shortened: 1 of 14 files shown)


lauraponce:
It looks like this new percent sign is rendering pretty dark — do you think reducing the font-weight of the % could help with that?

Screen Shot 2019-09-13 at 11 15 06 AM

mgwalker:
@lauraponce Oh, good catch! I borrowed the font-weight that the design system users for the `$` mask in text fields, but the fields in those tables don't look like fields everywhere else so it makes sense to tweak them!

mgwalker:
@lauraponce I double-checked and the percent sign in that table has the same size and weight as the dollar sign in the inputs in other tables (like the previous activities costs table). Should we pair up real quick next week to figure out exactly what we want to do? Fixing it live might be easier/faster. 😄

@jeromeleecms I tried to make sure that the percent sign thingy would be correct on every render, but it turns out (and it should have been obvious if I'd stopped and thought about it for a few minutes) that making that change every time breaks focus because the input box is removed from the DOM and then added back. So I undid that and it seems good now.

I don't know what to do with my hands

lauraponce:
👍 that sounds good! We could use the parking lot time?

mgwalker:
Changed the `font-weight` to 300.
image

web/src/styles/scss/forms.scss
```diff
@@ -69,6 +69,23 @@ div.form--with-reviews {
   }
 }
 
+.input-holder__percent {
+  display: inline-block;
+  position: relative;
+  &:after {
+    content: '%';
+    font-weight: 300;
+    font-size: 18px;
```
hbillings:
```suggestion
    font-size: $lead-font-size;
```

web/src/styles/scss/forms.scss
```diff
@@ -69,6 +69,23 @@ div.form--with-reviews {
...
+    position: absolute;
+    right: 8px;
```
hbillings:
```suggestion
    right: $spacer-1;
```

Issue #1567, opened 2019-05-31
Activities and Goals Section Review


jeromeleecms:
From our conversation in the #1565, the goals and activities section is asking for three things:

1) Name of the Goal
2) Explanation of the Goal
3) Associated Benchmarks on how the state will measure progress.

However, we only have two input fields: name of the goal + benchmarks. 

Per @jameshupp - "a single input should generally have a single ask. Otherwise we get a lot of different structures: benchmark-by-benchmark descriptions, a description and then a list of benchmarks, a description with benchmarks mixed in at various points, etc.

We could do something with the instructions themselves to make that clearer, though? Maybe just Description as a title, and then an instruction like:

Briefly describe how this goal relates to the vision identified in the program summary. Then provide any benchmarks the state will use to measure progress toward the identified goal.

(That would involve moving the first sentence from the section instructions to the ones for this specific input field.)

If this feels like too big a set of things to discuss, go ahead and approve the merge and we can figure this out as a separate need."

(copied from comment in #1565)

### This task is done when...
 - [ ] Instructions of Goals and Activities Align with our ask / input fields
 - [ ] Appropriate tasks/issues are created to support any implementation needs

lauraponce:
@jeromeleecms is this one outdated/obsolete since we have https://github.com/18F/cms-hitech-apd/issues/1819 now?

jeromeleecms:
Closed out because of #1819

PR #1816, opened 2019-09-13
Help text culling

quarterback:

- Truncated and removed excess text to improve navigation and clarity across the app user experience.
- Related to Issue #1567 

### DIFF
--- a/web/src/i18n/locales/en/activities/activities.yaml
+++ b/web/src/i18n/locales/en/activities/activities.yaml
@@ -1,8 +1,6 @@
 title: Program Activities
 helpText: >-
-  Explain the goals, timelines, and costs associated with each activity. Activities
-  are narrowly scoped bodies of work that comprise the APD as described within
-  the Program Overview section.
+  Explain goals, timelines, and costs associated with each activity.
 list:
   title: Activity List
   instruction:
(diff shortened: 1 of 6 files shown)


mgwalker:
There's a bunch of stuff cut out here, and I'm curious if we might eventually want to add it back in a less-intrusive way (maybe a little `?` button or something users can click to expand longer help text). If so, would it make sense to create a new issue about that, reference this one (so we'll have a link back to the longer text), and then throw the new issue into the icebox?

quarterback:
@mgwalker Yes, good call. I created #1817 but it could be worded better or more accurately.

quarterback:
The rationale for this PR:

- Culls the content from the blue additional help sections of the application to make it easier to flow through the app.
- If users are already trained in doing APDs and are using the real template to get instructions that are more current, do we need to repeat that content inside the application?
- By removing the content and testing it with users, will we discover that they're not paying attention to it being gone at all? If so, that might make it easier to design an experience that's even easier to navigate down the road? 
- A/B testing will give us an opportunity for real data around whether it's _actually_ helpful or not to have this information or whether we ought to implement new data/information into the application. 
- For A/B purposes, just using the deploy URLs will be easier than spinning a separate instance of the app. 

cc: @jeromeleecms @lauraponce

lauraponce:
It looks like some of the labels got taken out with the help text in the Standards and Conditions section:

Screen Shot 2019-10-16 at 8 44 33 PM

@jeromeleecms I think we should hold off on this PR for now since we're making the other big changes to the activities section. What do you think?

jeromeleecms:
Closing this one - we will revisit the content later.

Issue #1798, opened 2019-09-09
Update the frontend to use the PATCH API endpoint


mgwalker:
Once the `PATCH /apds/:apdID` endpoint is available in public and the frontend is computing JSON Patch diffs, the frontend can make the switch.

### This task is done when...
- [ ] the frontend updates APDs with the `PATCH /apds/:apdID` API endpoint
- [ ] the frontend and backend data models are the same instead of having to massage back and forth between them

PR #1818, opened 2019-09-13
Compute JSON patchset of APD edits in the state Medicaid director and office section

### LINKED ISSUE #1831: JSON patch for state director and office section
### This task is done when...
 - [ ] changes to the state Medicaid director or office are captured as JSON patchsets
 - [ ] new redux actions are defined for these edits

mgwalker:
To prepare for the glorious future in which saves are done via JSON patches, this PR does the work of capturing user edits as a JSON patchset.

My first thought was to just make a new `patch` reducer that captures all the existing edit actions. Good news! The edit actions are sending along nested objects which we'd need to extract the keys from in order to build a property path. E.g.:

```js
{
  stateProfile: {
    medicaidDirector: {
      name: "Bob"
    }
  }
}
```

...would get converted into this path:

```js
"stateProfile/medicaidDirector/name"
```

And it turns out, that's actually kinda tricky, mainly around knowing when to stop traversing the object for keys. For example, what if the value itself is an object? How do you know you've reached the actual value?

So... I decided our current actions are kind of messy anyway, so I would experiment with adding a whole new set of actions. Enter `editApd`, which has such actions as `setMedicaidDirectorName` and `setMedicaidOfficeCity`. These are really simple functions that return redux-dispatchable objects with the correct type, path, and value.  E.g.:

```js
setMedicaidDirectorName("Bob")
```

...returns:

```js
{ type: Symbol('edit apd'), path: 'stateProfile/medicaidDirector/name', value: 'Bob' }
```

One of the huge upsides I see here is that components no longer need to know about the shape of the state in order to dispatch changes. They can just say "I want to update the Medicaid director's name," and the action creator abstracts away the path for them. That feels like it falls in line with using selectors for pulling stuff from the state, so the components just don't have to know what's under the hood.

With those in place, I jumped into the state profile/director/office component (`containers/ApdStateProfileMedicaidOffice.js`) to update it to use these new actions in addition to the old ones. (We can't take out the old ones yet, since they're still used to do the actual saves.) While I was in there, I also switched it over to using hooks, including redux hooks.  Yay, we can get rid of the `connect()` method, and the component can directly dispatch actions instead of relying on that weird wrapper thing.

**If we decide to go this route**, I think the right path forward is to do multiple pull requests over time as we refactor each piece, rather than trying to do it all in one ginormous PR.

- no visible changes
- closes #1831

### DIFF
--- a/web/src/containers/ApdStateProfileMedicaidOffice.test.js
+++ b/web/src/containers/ApdStateProfileMedicaidOffice.test.js
@@ -4,7 +4,6 @@ import { Provider } from 'react-redux';
 import configureStore from 'redux-mock-store';
 import thunk from 'redux-thunk';
 
-import { UPDATE_APD } from '../actions/apd';
 import { EDIT_APD } from '../actions/editApd';
 
 import StateProfile from './ApdStateProfileMedicaidOffice';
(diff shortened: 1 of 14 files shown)


web/src/containers/ApdStateProfileMedicaidOffice.test.js
```diff
@@ -1,83 +1,148 @@
...
-        email: 'email',
-        phone: 'phone'
-      },
-      medicaidOffice: {
-        address1: 'address 1',
-        address2: 'address 2',
-        city: 'city',
-        state: 'state',
-        zip: 'zip'
+  const store = mockStore({
```
mgwalker: In the `connect()` components, we export the un-connected component as `plain` and can bypass the state-to-props process altogether. In that case, we inject the props directly into the component. With redux hooks, though, there's no way to separate the state-to-props process. As a result, in order to test the component, we need to actually give it a redux store. So, here we create .a story for the component.

web/src/containers/ApdStateProfileMedicaidOffice.test.js
```diff
@@ -1,83 +1,148 @@
...
     expect(
-      shallow(<StateProfile {...localProps} updateApd={props.updateApd} />)
-    ).toMatchSnapshot();
-
-    // name length validation
-    localProps.stateProfile.medicaidDirector.name = 'a';
-    expect(
-      shallow(<StateProfile {...localProps} updateApd={props.updateApd} />)
+      mount(
+        <Provider store={store}>
```
mgwalker: We also have to wrap the component in a `<Provider>` in order to give it the store that it needs access to. This thing is the reason we need to do a full mount - a shallow render would give us a `<Connected>` thing and inside it, which is what we really care about. A full mount gets us that whole thing.

web/src/reducers/patch.js
```diff
@@ -0,0 +1,33 @@
...
+const reducer = (state = initialState, action) => {
+  switch (action.type) {
+    case SAVE_APD_SUCCESS:
+      return initialState;
+
+    case EDIT_APD: {
+      const existingIndex = state.findIndex(
+        ({ op, path }) => op === 'replace' && path === action.path
+      );
+      if (existingIndex < 0) {
```
mgwalker: I should put comments in the code about this, but the gist is if there's already a patch to update this specific path, we don't need another one, we can just update the existing one.

But now that I think about it, I don't think this is a fully correct implementation, because if I update the 5th element in an array, delete the fifth element in an array, and then update the new 5th element in the array, that is different than updating the 5th element twice and then deleting it. So, probably need to go back and rethink this. Maybe update the existing patch if no add or remove operations occur later in the patch set. 🤔

mgwalker:
This will merge into the `feature/document-model` branch because it has breaking changes. Until #1798 is done, changes to the medicaid director/office won't be saved.

PR #1829, opened 2019-09-19
Add an API method to update APD documents via JSONPatch

### LINKED ISSUE #1795: Add a new API method to update APDs using a JSON Patch
### This task is done when...
- [ ] API `PATCH /apds/:apdID` endpoint exists and works correctly
  - update the document model only, not the relational model
- [ ] the API validates the incoming APD against a schema, to ensure arbitrary data isn't added to the database
- [ ] the new `PATCH` endpoint is disabled in production
  - if this endpoint is used while the old `PUT` endpoints still exist, it could create corrupted data

mgwalker:
Adds a `PATCH /apds/:id` API endpoint to update APD documents via a JSONPatch set.

- adds migration to fix up some of the data in the existing documents
- adds an APD document schema
- adds an endpoint that accepts an array of JSONPatch objects and updates the document
  - removes patches that target properties that can't be changed
  - validates the resulting document, removing fields that aren't part of the schema
  - saves the document
  - if the state profile was updated for the document, also update the state table with that information (that way when someone creates a new APD for that state, they'll get the latest state profile information by default)
- closes #1795


mgwalker:
Rather than disable this in production, we're going to use the `feature/document-model` branch to collect pieces of work that tightly depend on each other before they're ready for production. This PR will merge into `feature/document-model`. Future document model PRs should do the same.

PR #1849, opened 2019-09-23
Show year range at APD title

### LINKED ISSUE #1820: Add date range to top of APD in the app
From our federal reviewer research in the pilot, it came up that having the full range of years that the APD is covering would be more helpful than just the first year. For example instead of Texas 2019 APD, it would be Texas 2019-2021 APD.

### This task is done when...
 - [ ] Update the year at the top of the APD to be the range of years for which they are requesting funding based on what they select in the app.

mgwalker:

- updates the APD title to show the year range instead of just the first year
  screenshot
- closes #1820

### DIFF
--- a/web/src/containers/ApdApplication.test.js
+++ b/web/src/containers/ApdApplication.test.js
@@ -67,7 +67,7 @@ describe('apd (application) component', () => {
         data: {
           id: 'bloop',
           name: 'florp',
-          years: ['dinkus', 'dorkus']
+          years: ['dinkus', 'dorkus', 'durkus']
         }
       },
       dirty: {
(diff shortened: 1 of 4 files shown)


web/src/containers/ApdApplication.test.js
```diff
@@ -86,7 +86,7 @@ describe('apd (application) component', () => {
       dirty: 'moop moop',
       isAdmin: false,
       place: 'place',
-      year: 'dinkus'
+      year: 'dinkus-durkus'
```
hbillings: ah yes, the span between when my brother and I were born

Issue #1819, opened 2019-09-16
Convert Goals and Objectives Section to align with Objectives and Key Results


jeromeleecms:
Per leadership request (and it just makes sense!), content should do a review of the language in the Goals and Objective Section and revise it to align with CMS' move towards utilizing Objectives and Key Results. 

Primer on OKR: https://felipecastro.com/en/okr/what-is-okr/
 
### This task is done when...
 - [ ] @NAretakis has provided some feedback to team about what is expected from OKRs
 - [ ] Content has had a chance to review current state and see what needs to be done
- [ ] Design (user interaction) and Dev are consulted as needed if changes outside of content are required
- [ ] Appropriate issues/pull requests are added to appropriate sections

quarterback:
Pull this to start working on this afternoon. Will pull down the content and push something to review.

jeromeleecms:
@quarterback Does this have any overlap with #1407? Is seems like this is solving for a similar idea -- what are we try to do? How will me measure?

quarterback:
I'd surely be up for merging those issues or closing #1407 in favor of this one.

quarterback:
@NAretakis @jeromeleecms I'm going to rewrite this section contentwise, but if we feel like it needs to do _more_ than just that to be line with the OKR framework, then we'd need to implement some kind of design change. Right now, my changes will focus on try to use OKR language to better help people understand what's being asked of them.

quarterback:
<img width="661" alt="Screen Shot 2019-09-24 at 9 52 12 AM" src="https://user-images.githubusercontent.com/4461062/65533654-ea6bf780-deb2-11e9-92ca-7a9b7d87c871.png">
<img width="575" alt="Screen Shot 2019-09-24 at 9 52 19 AM" src="https://user-images.githubusercontent.com/4461062/65533655-ea6bf780-deb2-11e9-85ec-5a423f39f82e.png">

These aren't accurate anymore because I just pushed a new change that removes that goal content to the goals section (previously, there was no content there at all before...) which should make it easier to navigate. But it's more direct (consistent with our style guide)

jeromeleecms:
@NAretakis Will provide some language around OKRs

jeromeleecms:
@NAretakis Any thoughts/updates?

NAretakis:
Yes! Thoughts:

[...]

PR #1850, opened 2019-09-24
Rewritten goals and objectives content

quarterback:

- First revision at #1819 + #1567
- Remastered content to reflect OKR language.


quarterback:
I did a minimally invasive change to the text, but I think if we're really talking about changing the structure of content to build around OKRs, it'd require changing the whole app to do that, including the format of questions and sections so that it's consistent. 

Also I think it'd really require CMS to provide some baseline around language, because we can make something up, but it won't necessarily reflect the way that an analyst might interpret it, since we're not the folks who fill these things out. @jeromeleecms 

cc: @lauraponce

lauraponce:
@jeromeleecms this seems good to me to merge in and then we'll work off of this for the new OKR content. What do you think?

Issue #1826, opened 2019-09-18
Update APD GET endpoints to use the document model


mgwalker:
### This task is done when...
 - [ ] Update `GET /apds` to pull name, status, and years information from the `document` field instead of the individual columns (the `id` and `updated` fields will remain distinct columns)
 - [ ] Update `GET /apds/:id` to fetch the document model instead of the relational model

Issue #1800, opened 2019-09-09
Update API endpoint for creating APDs to stop creating relational APDs


mgwalker:
### This task is done when...
 - [ ] API `POST /apds` endpoint is updated to only create the document model instance of an APD, and references to the relational model are removed

Issue #1799, opened 2019-09-09
Remove the old API endpoint for updating APDs


mgwalker:
### This task is done when...
 - [ ] the `PUT /apds/:apdID` is removed

Issue #1796, opened 2019-09-09
Update the frontend to capture APD changes as JSON Patch


mgwalker:
Currently, we track changes to an APD in the frontend via the `dirty` reducer. Perhaps we need a new `patch` reducer?

### This task is done when...
 - [ ] APD changes are stored locally as JSON Patch
  - locally-stored changes should be cleared out when an APD is saved successfully
 - ~the dirty state tracker is still present and working~ - we opted to do this in a feature branch so we could ditch the dirty state tracking. Doing both dirty state and JSON Patch was going to be complicated and create a lot of technical debt to clean up later. It was easier, faster, and cleaner to just cut the dirty state out.

PR #1861, opened 2019-09-26
[WIP] Switch frontend to document model

mgwalker:
This PR will merge the `feature/document-model` branch into master, and will include the following work:

- #1795 - add an API method to update the APD using JSON Patch sets
- #1796 - compute JSON patch of APD changes on the frontend
- #1826 - update the API `GET /apd/*` methods to fetch from the document model instead of the relational model
- #1800 - update the API `POST` method to only create document model APDs and stop creating relational model APDs
- #1798 - update the frontend to use the new API JSON Patch update method
- #1799 - remove the old methods for updating APDs

**NOTE:** until #1798 is finished, any frontend sections that have been converted to JSON Patch will no longer save. As we convert sections to JSON Patch, we are removing the old behavior rather than maintain two distinct behaviors.

- no user-visible changes


api/routes/apds/patch.endpoint.js
```diff
@@ -0,0 +1,107 @@
...
+            op: 'replace',
+            path: '/activities/1/costAllocation/2019/ffp/federal',
+            value: 80
+          }
+        ]
+      });
+
+      // The updated date is the date/time stamp of when the APD is saved, so
+      // it'll change with each test run.  Rather than figure out something
+      // fancy with the snapshots, just pull out the name and test it with a
```
mgwalker:
```suggestion
      // fancy with the snapshots, just pull out the date and test it with a
```

api/routes/apds/patch.js
```diff
@@ -0,0 +1,122 @@
...
+
+const logger = require('../../logger')('apds route put');
+const { raw: knex } = require('../../db');
+const { can, userCanEditAPD } = require('../../middleware');
+const apdSchema = require('../../schemas/apd.json');
+
+const ajv = new Ajv({
+  allErrors: true,
+  jsonPointers: true,
+  removeAdditional: true
```
mgwalker:
```suggestion
  // The validator will remove any fields that aren't in the schema
  removeAdditional: true
```

api/routes/apds/patch.js
```diff
@@ -0,0 +1,122 @@
...
+          logger.error(req, 'error patching the document');
+          logger.error(null, e);
+          return res.status(400).end();
+        }
+
+        const valid = validateApd(updatedDocument);
+        if (!valid) {
+          logger.error(
+            req,
+            validateApd.errors.map(({ dataPath, message }) => ({
```
mgwalker:
```suggestion
            // Rather than send back the full error from the validator, pull out just the relevant bits
            // and fetch the value that's causing the error.
            validateApd.errors.map(({ dataPath, message }) => ({
```

api/routes/apds/post.js
```diff
@@ -1,40 +1,82 @@
...
+        apd.stateProfile.medicaidDirector = {
+          ...apd.stateProfile.medicaidDirector,
+          ...stateProfile.medicaid_office.medicaidDirector
+        };
+
+        apd.stateProfile.medicaidOffice = {
+          ...apd.stateProfile.medicaidOffice,
+          ...stateProfile.medicaid_office.medicaidOffice
+        };
+        delete apd.stateProfile.medicaidOffice.director;
```
mgwalker:
```suggestion
        // An old version of the model had the director info contained inside the office field, so
        // just in case we're still hitting a really old source, delete the director from the office
        delete apd.stateProfile.medicaidOffice.director;
```

api/routes/apds/post.js
```diff
@@ -1,40 +1,82 @@
...
+      }
+
+      const valid = validatorFunction(apd);
+      if (!valid) {
+        logger.error(req, 'Newly-created APD fails validation');
```
mgwalker:
```suggestion
        // This is just here to protect us from the case where the APD schema changed but the
        // APD creation function was not also updated
        logger.error(req, 'Newly-created APD fails validation');
```

web/src/actions/app/apd.js
```diff
@@ -0,0 +1,36 @@
...
+export const selectApd = (
+  id,
+  { deserialize = fromAPI, global = window, pushRoute = push } = {}
+) => dispatch =>
+  axios.get(`/apds/${id}`).then(req => {
+    const apd = deserialize(req.data);
+
+    dispatch({ type: SELECT_APD, apd });
+
+    if (Object.keys(req.data.federalCitations).length === 0) {
```
mgwalker:
```suggestion
    // By default, APDs get an empty object for federal citations. The canonical list of citations is in frontend
    // code, not backend. So if we get an APD with no federal citations, set its federal citations to the initial
    // values using an EDIT_APD action. That way the initial values get saved back to the API.
    if (Object.keys(req.data.federalCitations).length === 0) {
```

web/src/actions/editApd/keyPersonnel.js
```diff
@@ -0,0 +1,105 @@
...
+    dispatch(updateBudget());
+  }
+};
+
+/**
+ * Set a key person's name
+ * @param {String} name Key person's new name
+ * @param {Number} index Index of the key person to update
+ */
+export const setKeyPersonName = (name, index) => ({
```
mgwalker: Swap the order of these so they match everywhere else.

web/src/components/DateField.js
```diff
@@ -22,7 +22,9 @@ const joinDate = ({ day, month, year }) => {
     return '';
   }
 
-  return `${year}-${month}-${day}`;
+  return `${year}-${month < 10 ? `0${month}` : month}-${
```
mgwalker:
```suggestion
  // Make sure it's an ISO-8601 date, which uses 2-digit month and day
  return `${year}-${month < 10 ? `0${month}` : month}-${
```

web/src/containers/ApdStateKeyPersonnel.js
```diff
@@ -1,79 +1,48 @@
...
-  updateApd,
-  updateBudget: dispatchBudget,
-  years
-}) => {
-  const handleChange = useCallback((index, field, value, isExpense = false) => {
-    updateApd({ keyPersonnel: { [index]: { [field]: value } } });
-    if (isExpense) {
-      dispatchBudget();
+const ApdStateKeyPersonnel = ({ add, poc, remove, years }) => {
+  const deletePerson = key => {
```
mgwalker: As we continue cleaning up the activity-byKey stuff in the state, we should update the `FormAndReivewList` component so that when it calls the delete handler, it passes an index instead of a key. Then we can update all these things so they don't have to walk through an array to figure out what to delete.

web/src/containers/activity/All.js
```diff
@@ -9,70 +9,74 @@ import {
...
+            list={other}
+            className="ds-u-border-bottom--0"
+            collapsed={NameAndFundingSourceReview}
+            expanded={NameAndFundingSourceForm}
+            noDataMessage={false}
+            onAddClick={onAdd}
+            onDeleteClick={onRemove}
+          />
+        </Subsection>
+        {keys.map((key, index) => (
```
mgwalker: I don't think we actually need these keys here. We have the list of activities, we can just iterate over that.

web/src/reducers/patch.js
```diff
@@ -0,0 +1,94 @@
...
+};
+
+const reducer = (state = initialState, action) => {
+  switch (action.type) {
+    case SAVE_APD_SUCCESS:
+      return initialState;
+
+    case ADD_APD_ITEM: {
+      const patches = getApdPatchesForAddingItem(action.state.apd, action.path);
+      return [...state, ...patches];
```
mgwalker: Why are these two different? Make these two the same!

web/src/containers/activity/StandardsAndConditions.js
```diff
@@ -1,62 +1,292 @@
...
+  setInteroperability,
+  setKeyPersonnel,
+  setLeverage,
+  setMinimizeCost,
+  setMITA,
+  setMitigationStrategy,
+  setModularity,
+  setReporting
+}) => {
+  const syncBusinessResults = html => {
```
mgwalker: In other places, we've used something like:

```js
const getHandleConditions = action => html => action(activityIndex, html)
```

Can we use that here to cut down on the code?

web/src/reducers/apd.js
```diff
@@ -77,6 +277,51 @@ const initialState = {
...
+        ])
+      };
+    }
+
+    case REMOVE_APD_YEAR: {
+      const patches = getPatchesToRemoveYear(state, action.value);
+      return { ...state, data: applyPatch(state.data, patches) };
+    }
+
+    // ------------------
```
hbillings: I guess?

mgwalker: 😂 
Above the line are reducers for the new actions, below the line are old reducers. I was (am?) operating on the assumption that a lot of the reducers below the line can be deleted.

PR #1865, opened 2019-09-27
Prepare changelog for next release

mgwalker:
Adds the latest release info to the changelog and prepares it for the next release.

A question: should we keep the complete change history in the changelog, or would we rather only track changes for *upcoming* releases in the changelog? Once we do a release, we'll have the change history on the [releases page](https://github.com/18F/cms-hitech-apd/releases), so having the complete history in the changelog seems redundant. 🤷‍♂

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,16 @@ Anticipated release: TBD
 
 #### 🚀 New features
 
+#### 🐛 Bugs fixed
+
+#### 🐛 Bugs fixed
+
+# v1.1.0
+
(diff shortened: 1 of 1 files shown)


mgwalker:
I don't think there's any standard practices about using both a changelog file and the Github releases tab, so we'd kinda be making something up.

It occurs to me now that one reason to keep a complete changelog forever is in case the project ever moves off of Github. The releases tab is tied to Github explicitly, but the changelog lives with the project itself so it'd happily move wherever the project moved.

PR #1870, opened 2019-10-01
JSON patch the activity list

### LINKED ISSUE #1835: JSON patch for activity list section
### This task is done when...
 - [ ] creating or deleting an activity; and changing an activity's funding source are captured as JSON patchsets
  - in the `activities/All` container
 - [ ] new redux actions are defined for these edits

mgwalker:

- begins migrating the application state from using key-based activities to index-based activities
- introduces a context for tracking which activity a component is tied to, rather than passing that information down over and over and over again
- JSON-patchifies adding and removing activities, as well as updating their names and funding sources
- closes #1835

### DIFF
--- a/web/src/containers/activity/All.js
+++ b/web/src/containers/activity/All.js
@@ -1,6 +1,5 @@
-import PropTypes from 'prop-types';
 import React from 'react';
-import { connect } from 'react-redux';
+import { useDispatch, useSelector } from 'react-redux';
 
 import FormAndReviewList from '../../components/FormAndReviewList';
 import EntryDetails from './EntryDetails';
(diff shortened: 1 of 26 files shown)


web/src/containers/activity/All.js
```diff
@@ -9,70 +8,59 @@ import {
...
+            expanded={NameAndFundingSourceForm}
+            noDataMessage={false}
+            onAddClick={add}
+            handleChange={update}
+            onDeleteClick={remove}
+          />
+        </Subsection>
+        {activities.map((activity, index) => (
+          <Waypoint id={activity.key} key={activity.key}>
+            <EntryDetails activity={activity} index={index} />
```
mgwalker: Instead of passing an activity key and number into `EntryDetails`, we now just pass the activity itself along with its index in the array. It's silly that we made each component down the line go fetch the activity again based on the key, but that's what we have/had.

web/src/containers/activity/EntryDetails.js
```diff
@@ -1,8 +1,8 @@
 import { Button, Review } from '@cmsgov/design-system-core';
 import PropTypes from 'prop-types';
 import React, { Fragment, useMemo, useRef, useState } from 'react';
-import { connect } from 'react-redux';
 
+import { Provider } from './ActivityContext';
```
mgwalker: Pull in an activity context provider. This is a custom thingy that does almost nothing on its own, but it wraps up some React functionality with a nice little bow to make it easier to use.

web/src/containers/activity/EntryDetails.js
```diff
@@ -67,41 +70,35 @@ const EntryDetails = ({ activity, aKey, num }) => {
...
-        <Goals aKey={aKey} />
-        <Schedule aKey={aKey} />
-        <Costs aKey={aKey} />
-        <ContractorResources aKey={aKey} />
-        <CostAllocate aKey={aKey} />
-        <StandardsAndConditions aKey={aKey} />
-        <Button variation="primary" onClick={() => setCollapsed(true)}>
-          Done
-        </Button>
+    <Provider value={{ activity, index }}>
```
mgwalker: This is a context provider. Any context consumers _down_ the tree from here will be able to access these values. I'm stashing off the activity and index. Now we don't have to pass anything down - child components can just grab the activity from the context.

One thing we should dig into is re-renders. I _think_ a component will re-render any time its context changes. We don't really want that - e.g., if an activity name changes, we don't need to activity goals to re-render. So maybe we should only put the index in the context and use that to go fetch the activity. 🤔

web/src/containers/activity/EntryDetails.js
```diff
@@ -67,41 +70,35 @@ const EntryDetails = ({ activity, aKey, num }) => {
...
+        id={`activity-${aKey}`}
+        className={`activity--body activity--body__${
+          collapsed ? 'collapsed' : 'expanded'
+        } activity--body__${num === 1 ? 'first' : 'notfirst'}`}
+        ref={container}
+      >
+        <Review heading={title} headingLevel={4} editContent={editContent} />
+        <div className={collapsed ? 'visibility--print' : ''}>
+          <Overview />
+          <Goals aKey={aKey} />
```
mgwalker: We'll be able to yank out `aKey={aKey}` from the rest of these as we migrate them over.

web/src/containers/activity/Overview.js
```diff
@@ -1,18 +1,19 @@
...
 import RichText from '../../components/RichText';
 import Instruction from '../../components/Instruction';
 import { Subsection } from '../../components/Section';
 import TextArea from '../../components/TextArea';
-import { selectActivityByKey } from '../../reducers/activities.selectors';
 
 const Description = props => {
-  const { activity, updateActivity } = props;
+  const { updateActivity } = props;
+  const { activity } = useContext(ActivityContext);
```
mgwalker: `useContext` is a fancy-pants way to consume a context without using the `<ContextConsumer>` thing. The Overview component doesn't have tests right now, so I'm not sure how we'll test this. Maybe we'll just have to wrap it with an `ActivityContext.Provider`, I dunno.

web/src/reducers/activities.selectors.js
```diff
@@ -2,15 +2,14 @@ import { createSelector } from 'reselect';
...
-export const selectActivityKeys = ({ activities: { allKeys } }) => allKeys;
-
-export const selectActivitiesByKey = ({ activities: { byKey } }) => byKey;
-
 export const selectActivityByKey = ({ activities: { byKey } }, { aKey }) =>
   byKey[aKey];
 
-export const selectAllActivities = ({ activities: { byKey } }) =>
-  Object.values(byKey);
+export const selectAllActivities = ({
```
mgwalker: For this selector, get activities from the APD instead of that weird `byKeys` thing we add.

web/src/reducers/budget.js
```diff
@@ -133,7 +133,7 @@ const buildBudget = bigState => {
 
   // Since all of our expenses are tied up in activities, we'll start
   // by looking at all of them and doing Magic Math™. (It's not magic.)
-  Object.values(bigState.activities.byKey).forEach(activity => {
+  bigState.apd.data.activities.forEach(activity => {
```
mgwalker: For the budget stuff, we also want to start using the activities from the APD rather than `byKey`.

PR #1893, opened 2019-10-11
JSON patch cost allocation FFP

hbillings:

### DIFF
--- a/web/src/containers/activity/CostAllocateFFP.test.js
+++ b/web/src/containers/activity/CostAllocateFFP.test.js
@@ -11,52 +11,45 @@ import {
   setCostAllocationFFPOtherFunding
 } from '../../actions/editActivity';
 
-jest.spyOn(React, 'useContext');
-React.useContext.mockImplementation(() => ({
-  index: 'activity index'
-}));
-
 describe('the CostAllocateFFP component', () => {
   const state = {
     apd: {
       data: {
-        activities: {
-          0: {
-            costAllocation: {
-              '1066': {
-                other: 10,
-                ffp: {
-                  federal: 90,
-                  state: 10
-                }
-              },
-              '1067': {
-                other: 0,
-                ffp: {
-                  federal: 10,
-                  state: 90
-                }
+        activities: [{
+          costAllocation: {
+            '1066': {
+              other: 10,
+              ffp: {
+                federal: 90,
+                state: 10
+              }
+            },
+            '1067': {
+              other: 0,
+              ffp: {
+                federal: 10,
+                state: 90
               }
             }
           }
-        }
-      }
-    },
-    budget: {
-      activities: {
-        key: {
-          costsByFFY: {
-            '1066': {
-              medicaidShare: 1550,
-              federal: 1430,
-              state: 120,
-              total: 1970
-            },
-            '1067': {
-              medicaidShare: 8870,
-              federal: 8770,
-              state: 100,
-              total: 9889
+        }]
+      },
+      budget: {
+        activities: {
+          key: {
+            costsByFFY: {
+              '1066': {
+                medicaidShare: 1550,
+                federal: 1430,
+                state: 120,
+                total: 1970
+              },
+              '1067': {
+                medicaidShare: 8870,
+                federal: 8770,
+                state: 100,
+                total: 9889
+              }
             }
           }
         }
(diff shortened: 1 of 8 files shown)


web/src/containers/activity/CostAllocateFFP.test.js
```diff
@@ -1,34 +1,41 @@
...
+  setCostAllocationFFPFundingSplit,
+  setCostAllocationFFPOtherFunding
+} from '../../actions/editActivity';
 
-describe('the CostAllocateFFP component', () => {
-  const sandbox = sinon.createSandbox();
+jest.spyOn(React, 'useContext');
+React.useContext.mockImplementation(() => ({
+  index: 'activity index'
+}));
```
hbillings: I am not entirely sure I need this; I copied it from the Contractor Resources tests in an attempt to get things to work.

web/src/containers/activity/CostAllocateFFP.test.js
```diff
@@ -118,19 +120,12 @@ describe('the CostAllocateFFP component', () => {
...
-      )
-    );
+    expect(props.setFundingSplit).toHaveBeenCalledWith(0, '1066', 35, 65);
   });
 
   test('maps redux state to component props', () => {
     const mapStateToProps = makeMapStateToProps();
-
-    expect(mapStateToProps(state, { aKey: 'key' })).toEqual({
+    expect(mapStateToProps(state, { activityIndex: 0 })).toEqual({
```
hbillings: This is exploding and I can't figure out why. When the test calls `makeSelectCostAllocateFFP` in `activities.selectors.js`, it bombs out on [this line](https://github.com/18F/cms-hitech-apd/blob/feature/document-model/web/src/reducers/activities.selectors.js#L35). Using `node --inspect` in Chrome, it looks like both the `budget` and `activity` are undefined at this point in the test. I've been trying to wrap my head around what might be going wrong with the selector, but I don't have enough brain cells to rub together at 5:30 on a Friday to figure it out, so documenting here.

web/src/reducers/activities.selectors.js
```diff
@@ -30,7 +30,7 @@ const selectBudgetForActivity = ({ budget }, { aKey }) =>
 
 export const makeSelectCostAllocateFFP = () =>
   createSelector(
-    [selectActivityByKey, selectBudgetForActivity],
+    [selectActivityByIndex, selectBudgetForActivity],
```
hbillings: I thought this might help with the above failing test scenario, but no dice. @mgwalker, if you can spare some time Tuesday, I would so appreciate a pair!

web/src/containers/activity/CostAllocateFFP.test.js
mgwalker: Nah, you don't need this anymore!

web/src/containers/activity/CostAllocateFFP.test.js
```diff
@@ -1,34 +1,41 @@
...
+              },
+              '1067': {
+                other: 0,
+                ffp: {
+                  federal: 10,
+                  state: 90
+                }
               }
             }
           }
```
mgwalker: This needs to be an array rather than object; e.g.:

```js
activities: [{
  costAllocation: {
    ...
  }
}],
```

The reason the test fails is because `activities` is an object here. The `selectActivityByIndex` selector does some safety checking on the requested index to make sure it's in-bounds for the array.  It does that by making sure the index is `< activities.length`, but an object doesn't have a `.length` property, so it ends up checking `index < undefined`, which is always false. So then `selectActivityByIndex` will always return null because the safety check fails.

web/src/containers/activity/CostAllocateFFP.test.js
```diff
@@ -153,13 +148,14 @@ describe('the CostAllocateFFP component', () => {
           year: '1067'
         }
       ],
-      costAllocation: state.activities.byKey.key.costAllocation
+      costAllocation: state.apd.data.activities[0].costAllocation
```
mgwalker: This works because the object has a property with key `0`. It should just keep working when the state is updated to an array, because instead of `0` being a property key, it'll be an array index.

web/src/containers/activity/CostAllocateFFP.test.js
mgwalker: See above for details. This also needs an activity key for the budget part, which is still based on activity keys (whoops!). Rather than try to pull all that out in this PR, you can add the activity key here and (with the fix describe above), that should take care of this:

```js
mapStateToProps(state, { activityIndex: 0, aKey: 'key' })
```

web/src/containers/activity/CostAllocateFFP.test.js
hbillings: thank you

PR #1904, opened 2019-10-22
Design unsupported browser banner

### LINKED ISSUE #1881: Design the cute little unsupported browser banner
This is for yellow-level browsers. From #1869

### This task is done when...
 - [ ] we know what words to use
 - [ ] we know where to put them

hbillings:
Closes #1881. One thing to note is that the library we're using isn't capable of detecting versions of any browser other than IE, so if we want to show this banner on old versions of Chrome, Firefox and Safari, we're going to need a different library. 

- Modifies the unsupported browser banner to have new styles
- Shows the banner on all versions of IE

Screen Shot 2019-10-22 at 2 20 04 PM


lauraponce:
I changed my user agent string to test the experience and I'm seeing it pop up on a blank page rather than appearing at the top of log in or on the screen before where they have to agree and continue. Should I be seeing it above one of those other pages? @hbillings 

Screen Shot 2019-10-22 at 4 09 57 PM

web/src/components/UpgradeBrowser.js
```diff
@@ -11,12 +11,6 @@ const browsers = [
     name: 'Mozilla Firefox',
     link: 'https://www.mozilla.org/en-US/firefox/new/'
   },
-  {
-    id: 'ie',
-    name: 'Internet Explorer 11',
-    link:
-      'https://support.microsoft.com/en-us/help/17621/internet-explorer-downloads'
-  },
```
mgwalker: Should we create an issue to add an Edge download link when the Chromium builds are released? I think they're supposed to be out in the first quarter of 2020, so not that long.

hbillings:
Since IE 11 is the last version of IE, for now, we can just check to see if the browser is IE at all. (When we do the red browser page, we'll need to do a more granular check, but it will be a different check than was in this code.)

This should be working now.

PR #1910, opened 2019-10-29
Cleanup Redux actions and states after the big JSON Patch refactor

mgwalker:
We created a bunch of explicit actions when we switched over to JSON Patch, and those replaced the old very general actions. Now that we're all switched over, we can get rid of all the old actions.

We can also entirely eliminate the `activities` reducer since the activities are now stored in the `apd` state rather than as a separate entity. The `reducers/activities.js` file still has some convenience functions that are used by the `apd` reducer, but we can refactor those later.

And finally, we can get rid of the `dirty` reducer, since it has been completely superseded by the `patch` reducer.

- cleans up a bunch of no-longer-used code


web/src/containers/ApdApplication.js
```diff
@@ -32,8 +32,8 @@ class ApdApplication extends Component {
     // Hook the browser's beforeunload event to catch page reloads or manual
     // changes to the URL bar while the APD is loaded.
     this.unloadListener = window.addEventListener('beforeunload', e => {
-      const { dirty } = this.props;
-      if (dirty) {
+      const { needsSave } = this.props;
```
hbillings: this is a much better name

PR #1915, opened 2019-11-04
Clean up React warnings and errors

mgwalker:
Adds keys, fixes prop types, and makes sure CSS class names aren't booleans.

There are three categories of outstanding errors/warnings:

- Prop type error passing the `fieldRef` prop into the design system `TextField` component. The component only wants a function, but React accepts both functions and objects. Objects are the newer, more preferred approach, so I've stuck with that. We can open an issue with the design system to update their prop types.
- There are warnings about `componentWillMount` and `componentWillReceiveProps`, both deprecated lifecycle methods. We're not using these in our code anymore, so those warnings are coming from our dependencies. We can't really do anything about them.
- Nearly 300 warnings due to deprecated events being subscribed by the current rich text editor we're using. It should continue to work fine, but those deprecated events could prevent us from upgrading to a more recent version of React at some point in the future (e.g., React 17 or 18).

- fixes most of the React warnings and errors caused by our app


web/src/containers/BudgetSummary.js
```diff
@@ -178,11 +159,13 @@ const BudgetSummary = ({ activities, data, years }) => (
           return (
             <tr
               key={ffy}
-              className={ffy === "total" && "budget-table--row__highlight budget-table--total"}
```
mgwalker: This results in `className` being either `"budget-table--row__highlight budget-table--total"` or `false`. It turns out that React shouts at you if your `className` isn't a string. Who knew?

web/src/containers/activity/ContractorResource/ContractorResourceReview.js
```diff
@@ -54,7 +54,7 @@ ContractorResourceReview.propTypes = {
     name: PropTypes.string,
     start: PropTypes.string,
     totalCost: PropTypes.number,
-    years: PropTypes.array
```
mgwalker: This was never an array. 🤷‍♂

web/src/containers/activity/StatePersonnel.js
```diff
@@ -40,7 +40,7 @@ const StatePersonnel = ({ activityIndex, add, personnel, remove }) => {
 };
 
 StatePersonnel.propTypes = {
-  activityIndex: PropTypes.string.isRequired,
```
mgwalker: And this was never a string. 🤷‍♂ 🤷‍♂

web/src/components/Header.js
```diff
@@ -126,11 +126,12 @@ Header.propTypes = {
   authenticated: PropTypes.bool.isRequired,
   currentUser: PropTypes.object,
   isAdmin: PropTypes.bool.isRequired,
-  ariaExpanded: PropTypes.bool.isRequired,
+  ariaExpanded: PropTypes.bool,
   showSiteTitle: PropTypes.bool.isRequired
 };
 
 Header.defaultProps = {
+  ariaExpanded: false,
```
hbillings: 👌

web/src/components/NumberField.js
```diff
@@ -19,6 +19,9 @@ const NumberField = ({ fieldRef, ...props }) => {
     };
   }, []);
 
+  // A bug in the CMS design system says the fieldRef prop must be a function,
+  // but React also accepts objects. Objects are the newer approach, so we'll
+  // stick with that and just accept the prop types warning for now.
```
hbillings: Thanks for documenting this here!

web/src/containers/BudgetSummary.js
hbillings: Are you saying that the component... has no class? ;-)

PR #1920, opened 2019-11-15
Automatically save changes

mgwalker:
Automatically saves the APD as changes are being made. Waits 300 milliseconds after the last change before saving, to prevent flooding the user's computer with a high volume of network traffic. Prevents sending a save while a save is already in progress, since we can't guarantee what order saves will be processed by the server, but queues the new save and sends it when the current one completes.

- adds a redux middleware to handle firing save actions
- updates patch reducer to be smart about not wiping out changes made while a save is ongoing
- updates save action to debounce and queue

### Needs

- review of the concept and behavior
- consideration of the accessibility of this approach - presumably we don't want to announce every save. Maybe we don't need to announce anything as long as the saves are successful because "saved" is now the default state?
- consideration of what this does for multiple simultaneous users
- updated tests


web/src/reducers/patch.test.js
```diff
@@ -14,8 +14,16 @@ describe('JSON patch reducer', () => {
     expect(reducer(undefined, {})).toEqual([]);
   });
 
-  it('resets to the initial state when an APD is saved', () => {
-    expect(reducer({}, { type: SAVE_APD_SUCCESS })).toEqual([]);
+  it('removes the already-saved entries when an APD is successfully saved', () => {
+    // Prime the reducer by trigger a save request. The state at this point is
+    // marked for saving and should be removed when the safe is cucessful
```
hbillings:
```suggestion
    // marked for saving and should be removed when the safe is successful
```

hbillings: although that was quite an impressive typo

web/src/saveMiddleware.js
```diff
@@ -0,0 +1,26 @@
+import {
+  ADD_APD_ITEM,
+  ADD_APD_YEAR,
+  EDIT_APD,
+  REMOVE_APD_ITEM,
+  REMOVE_APD_YEAR
+} from './actions/editApd';
+import { saveApd } from './actions/apd';
+
+const saveMiddleware = (store, { save = saveApd } = {}) => next => action => {
```
hbillings: This makes my head hurt.

web/src/reducers/patch.test.js
mgwalker: What the hell? 🤣

web/src/saveMiddleware.js
mgwalker: Yeah... The way this redux middleware works is funky.

PR #1929, opened 2019-11-21
Add a screen to indicate when an APD is loading

### LINKED ISSUE #1368: When opening an APD, the system will notify user that the APD is loading
For any size APDs, it would be nice to include some visual indication that an APD is loading to help users recognize that the system is not stuck or unresponsive.

History: 
-In #1339, large APDs took time to load.
-In #1340, we attempted to minimize this by enabling compression on the API requests.

### This task is done when...
 - [ ] Design has had a chance to review this and given a go-no-go on the need for this
 - [ ] Design has created an mockup (if applicable)
- [ ] New Issue created for Dev. implementation

mgwalker:

- when the user selects an existing APD or chooses to create a new one, show a loading indicator while they wait for it to become available:
  ezgif com-optimize
- closes #1368 

- [ ] ~Tests have been updated (and all tests are passing)~ N/A


web/src/containers/StateDashboard.js
```diff
@@ -28,7 +29,15 @@ const StateDashboard = (
   },
   { global = window } = {}
 ) => {
+  const [isLoading, setIsLoading] = useState(false);
```
hbillings: This is unrelated to this PR, but your use of hooks here reminds me: Should we write up something in the wiki about why we don't use hooks more extensively?

mgwalker: Oh, good idea! We use some of them pretty extensively at this point, but avoid others. Definitely worth explaining why!

mgwalker:
Hmm. That is a great question @jeromeleecms. No, it doesn't announce that the APD is loading. It also doesn't announce when the APD is finished loading. I'm curious how that should behave. If the page loaded immediately, it would probably be fine not to announce anything because clicking a link implies that the page will change, but since we have a delay, an announcement would probably be helpful... 🤔

mgwalker:
Added ARIA announcements for when an APD is loading and when it has finished loading. After asking my question above, it was pretty obvious that if we indicate something visually, we need to have a non-visual indication as well. It also made sense to me to announce when the APD was finished loading, otherwise it's not possible to tell when the app is ready to use again.

PR #1953, opened 2019-12-04
Upload images from rich text editors instead of embedding them

### LINKED ISSUE #1735: Upload images as separate files instead of embedded in text fields
Currently in our rich text editor, if the user adds an image, we encode that image in base64 and add it to the text. The result is that the text field can be enormous, and if the user adds multiple images, it can exceed our upload limit, meaning the APD couldn't be saved.

The rich text editor we use provides a mechanism for handling images. Right now we use it to base64 encode the image data, but instead, we could upload the images and then just add a link to the text. That would dramatically reduce the data size of the text field. Depending on how we implement the links, it could even allow browsers to cache the images separate from the APD data (which is not cached), speeding up data loads (maybe!). It will *definitely* make it easier to compute differences in text fields over time.

### This task is done when...
- [ ] backend provides an API path for uploading images associated with an APD
  - [ ] we know where to store files (S3 in production/staging/preview; what about dev?)
- [ ] backend provides an API path for fetching an image
  - [ ] enforces that only users with access to the APD can access the image
- [ ] frontend uses the API to upload images, and embeds URLs into the rich text instead of encoded images

mgwalker:

- adds API endpoints to upload and view files
- **not done**: updates rich text editor to upload images and inserts `` tags instead of data URIs
- closes #1735

  - [x] Updated the wiki


.circleci/config.yml
```diff
@@ -451,7 +451,10 @@ jobs:
             apt-get -qq install jq -y
             pip install --quiet awscli
             ./aws.sh \
+              --API_AWS_ACCESS_KEY_ID "${<< parameters.env >>_API_AWS_ACCESS_KEY_ID" \
+              --API_AWS_SECRET_ACCESS_KEY "${<< parameters.env >>_API_AWS_SECRET_ACCESS_KEY" \
               --API_DATABASE_URL "${<< parameters.env >>_API_DATABASE_URL}" \
+              --API_FILE_S3_BUCKET "${<< parameters.env >>_API_FILE_S3_BUCKET" \
```
mgwalker: These are the credentials and the S3 bucket to write files into. Staging/production save to S3.

.gitignore
```diff
@@ -8,3 +8,5 @@ node_modules
 
 npm-debug.log*
 pgdata
+__files
+api/test-data/files
```
mgwalker: `__files` is where files are uploaded in local dev. We don't want those going into git. (And you might want to remember to clean those out periodically 😁)

`api/test-data/files` is where files are stored for endpoint testing. There is one file there, and this rule will prevent git from capturing any new files that end up there - for example, after running the endpoint tests.

api/routes/apds/files.endpoint.js
```diff
@@ -0,0 +1,147 @@
...
+        );
+
+        expect(response.statusCode).toEqual(404);
+        expect(body).toMatchSnapshot();
+      });
+
+      it(`with an APD in a state other than the user's state`, async () => {
+        const { response, body } = await request.get(
+          url(4001, '74aa0d06-ae6f-472f-8999-6ca0487c494f'),
+          { jar: cookies }
```
hbillings: cookie jar

hbillings: Seriously, though -- should this be named as a test file?

api/store.js
```diff
@@ -1,172 +0,0 @@
-const fs = require('fs');
```
hbillings: This seems possibly bigger than just the images, so I just wanted to flag to make sure we want to delete this?

docs/database-diagram.dot
```diff
@@ -1,9 +1,21 @@
...
+ * Graphviz of '<open file '<stdin>', mode 'r' at 0x7f3fab3ed0c0>', created 2019-12-04 19:56:55.514987
  * Generated from https://github.com/rm-hull/sql_graphviz
  */
 digraph g { graph [ rankdir = "LR" ];
 
+  "apd_files" [
+    shape=none
+    label=<
+      <table border="0" cellspacing="0" cellborder="1">
+        <tr><td bgcolor="lightblue2"><font face="Times-bold" point-size="20">apd_files</font></td></tr>
```
hbillings: WOW I haven't seen cell `bgcolor` attributes or `font` tags in like fifteen years 😳

api/store.js
mgwalker: It was originally intended for uploading RFPs and contracts, but we never used it. The upload API endpoints in this PR can be used for the RFPs and contracts too, I'm pretty sure, so no need for this parallel implementation!

PR #1957, opened 2019-12-10
Add hover state to disabled buttons

hbillings:
I can't figure out how to screenshot this because cursors don't show up in screenshots. But, this adds a `cursor: not-allowed` state to disabled buttons, which you can see on the log in button.

- adds `cursor: not-allowed` to buttons with the `disabled` attribute (NOT the `.ds-c-button--disabled` state, for reasons having to do with `pointer-events`)


lauraponce:
@hbillings it looks like the preview URL didn't get generated — is there a way to see this change?

hbillings:
@lauraponce I think it's stalled because of the frontend vulnerability updates, which I fixed in #1958. Once I get that approved merged in, this should work. 🤞

PR #1960, opened 2019-12-12
Replace a bunch of standards and conditions text boxes with just a couple

### LINKED ISSUE #1947: Replace monolithic Standards and Conditions section with single question
Remove the long Standards and Conditions section at the end of the section and add in the new Standards and Conditions question right before the statement of alternative considerations.

### This task is done when...
 - [ ] The standards and conditions have been consolidated to a single question
Screen Shot 2019-11-26 at 9 26 16 PM

mgwalker:
Just what it says on the tin!

image

---

Readonly/print view if the user entered text:
image

---

Readonly/print view if the user did not enter any text:
image

- migrates existing APDs to concatenate the multiple standards and conditions into a single text field
- updates the APD schema to replace the multiple standards and conditions fields with two fields: `supports` and `doesNotSupport`
- adds new redux actions to set those two fields; removes the actions for the individual standards and conditions
- updates the UI to display and edit those two fields
- updates the readonly/print view to display those two fields
- removes YAML for instructions and whatnot of the removed standards and conditions
- closes #1947 


mgwalker:
We'll get a preview link when #1958 goes in and updates our dependencies

mgwalker:
Yep, we can do that! I went with the assumption that if the user put text in the second box ("does not support the standards and conditions"), then they probably wouldn't put anything in the first box, since the two boxes seemed (to me) to be contradictory. It felt weird to me to call out that they hadn't said how they were supporting the standards and conditions if they did say why they ***weren't***. 😛

jeromeleecms:
Just for context to help understand the "Why" of this change -- since there are multiple standards and conditions, the state may comply with 90% of them -- but there may some that may be not applicable. For example: MAGI determinations  may not make sense with a HITECH infrastructure project for public health, so they theoretically they would have responses in both boxes.

mgwalker:
Got it, that makes sense! Updated!

web/src/containers/activity/StandardsAndConditions.js
```diff
@@ -1,247 +1,75 @@
...
-          content={activity.standardsAndConditions.reporting}
-          onSync={getHandler(setReporting)}
-          editorClassName="rte-textarea-l"
+  setDoesNotSupport,
+  setSupport
+}) => (
+  <Fragment>
+    <h4 className="ds-h4">Standards & Conditions</h4>
+    <div className="ds-u-margin-bottom--6 ds-u-margin-top--3">
+      <h5 className="ds-h5 ds-u-margin-bottom--3">
```
hbillings: These probably shouldn't be headings, because they're really describing the text input instead of introducing a section. We can make them `p` tags with `ds-h5` styling, or we could also make them `label`s for their text editors. (I don't think either of the textareas here have labels.) The latter would be more accessible (and they should retain the styling).

web/src/containers/activity/StandardsAndConditions.js
```diff
@@ -1,247 +1,75 @@
...
+        content={activity.standardsAndConditions.supports}
+        onSync={html => setSupport(activityIndex, html)}
+        editorClassName="rte-textarea-1"
+      />
+
+      <div className="ds-c-choice__checkedChild ds-u-margin-top--3">
+        <h5 className="ds-h5">
+          If this activity does not support the Medicaid standards and
+          conditions, please explain.
+        </h5>
```
hbillings: We could probably just move this into the `label` attribute of the `TextArea`.

web/src/containers/viewOnly/activities/Activity.js
```diff
@@ -300,9 +250,37 @@ const Activity = ({ activity, activityIndex }) => {
         </small>
         <br /> Standards and Conditions
       </h3>
-      {Object.entries(activity.standardsAndConditions).map(
-        ([key, description]) => buildStandardsAndConditions(key, description)
+
+      <h4>
```
hbillings:
```suggestion
      <p><strong>
```

web/src/containers/viewOnly/activities/Activity.js
```diff
@@ -300,9 +250,37 @@ const Activity = ({ activity, activityIndex }) => {
...
+        This activity supports the Medicaid standards and conditions from 42 CFR
+        433.112.
+      </h4>
```
hbillings:
```suggestion
      </strong></p>
```

web/src/containers/viewOnly/activities/Activity.js
```diff
@@ -300,9 +250,37 @@ const Activity = ({ activity, activityIndex }) => {
...
+      {activity.standardsAndConditions.supports ? (
+        <div
+          dangerouslySetInnerHTML={{
+            __html: activity.standardsAndConditions.supports
+          }}
+        />
+      ) : (
+        <div>
```
hbillings:
```suggestion
        <p>
```

web/src/containers/viewOnly/activities/Activity.js
```diff
@@ -300,9 +250,37 @@ const Activity = ({ activity, activityIndex }) => {
...
+          No response was provided for how this activity will support the
+          Medicaid standards and conditions.
+        </div>
```
hbillings:
```suggestion
        </p>
```

web/src/containers/viewOnly/activities/Activity.js
```diff
@@ -300,9 +250,37 @@ const Activity = ({ activity, activityIndex }) => {
...
       )}
+
+      <h4>
```
hbillings:
```suggestion
      <p><strong>
```

web/src/containers/viewOnly/activities/Activity.js
```diff
@@ -300,9 +250,37 @@ const Activity = ({ activity, activityIndex }) => {
...
+        This activity does not support the Medicaid standards and conditions
+        from 42 CFR 433.112.
+      </h4>
```
hbillings:
```suggestion
      </strong></p>
```

web/src/containers/viewOnly/activities/Activity.js
```diff
@@ -300,9 +250,37 @@ const Activity = ({ activity, activityIndex }) => {
...
+      {activity.standardsAndConditions.doesNotSupport ? (
+        <div>{activity.standardsAndConditions.doesNotSupport}</div>
+      ) : (
+        <div>
```
hbillings:
```suggestion
        <p>
```

web/src/containers/viewOnly/activities/Activity.js
```diff
@@ -300,9 +250,37 @@ const Activity = ({ activity, activityIndex }) => {
...
+          No response was provided for how this activity will support the
+          Medicaid standards and conditions.
+        </div>
```
hbillings:
```suggestion
        </p>
```

mgwalker:
Updated per reviews and comments!

jeromeleecms:
Thanks @mgwalker -- couple more thoughts:

1) It looks like TinyMCE was applied to the first box now (I don't remember it being there before). Can we apply it to the "Does not comply" box too?

2) Not sure if this is a general issue, but I put a line break in the first box and the system recognized it as content, saved it,  and didn't provide our canned "did not provide a response." Not a deal breaker, but wanted to just call it out and see if you saw any potential issues with this behavior.

image

mgwalker:
That should be an easy fix!

mgwalker:
I lied. 😛 It's not an easy fix. Because the rich text editor creates HTML, if you put anything at all in it – even just spaces – it creates non-whitespace content.  It puts each line into a `<p>` tag. E.g., if you type:

```



```

the rich text editor creates:

```
<p></p>
<p></p>
<p></p>
```

Short of trying to parse the HTML ourselves, which *please no*, there's really no way for us to tell that content is effectively empty.

jeromeleecms:
OK - as long as this doesn't present a security or back end issue, I can handle having a blank area in the exported PDF 👍

hbillings:
@mgwalker This is why I hate WYSIWYG editors with the passion of a thousand burning suns. (Which really makes no sense, because stars have no emotions.) Do we need the RTE here at all? What content are we expecting people to enter?

PR #1971, opened 2019-12-18
Change same site property to "none"

mgwalker:
This PR changes the `samesite` property on our authorization cookies to `none`. This will allow any site to send the cookie to our API. However, if the API and the frontend are on different domains, the cookie's `secure` property must also be set and the connection must happen over SSL/TLS. This cookie sets the `secure` property.

This should fix an issue where the `samesite=strict` setting meant that our staging site no longer worked, and should also continue to make it possible for 3rd-party apps to login and work.

- fixes cookie `samesite` property
- adds an OWASP config to ignore the `samesite` cookie warning

- ~Changelog is updated as appropriate~


owasp/api-scan.conf
```diff
@@ -0,0 +1,76 @@
...
+10024	WARN	(Information Disclosure - Sensitive Information in URL - Passive/beta)
+10025	WARN	(Information Disclosure - Sensitive Information in HTTP Referrer Header - Passive/beta)
+10026	WARN	(HTTP Parameter Override - Passive/beta)
+10027	WARN	(Information Disclosure - Suspicious Comments - Passive/beta)
+10032	WARN	(Viewstate Scanner - Passive/release)
+10037	WARN	(Server Leaks Information via "X-Powered-By" HTTP Response Header Field(s) - Passive/beta)
+10040	WARN	(Secure Pages Include Mixed Content - Passive/release)
+10045	WARN	(Source Code Disclosure - /WEB-INF folder - Active/release)
+10048	WARN	(Remote Code Execution - Shell Shock - Active/beta)
+10054	IGNORE	(Cookie Without SameSite Attribute - Passive/beta)
```
mgwalker: Add an OWASP config so that we can ignore this one particular warning. We _intentionally_ set `samesite` to `none`, so we don't want our tests to fail because of it.

PR #1977, opened 2019-12-19
Fix a bug that prevents admins from changing a user's account without also changing the user's email address

### LINKED ISSUE #1973: Admins cannot edit users without changing their email address
Attempting to edit a user without changing their email address results in the error, "A user already exists with that email address." This is likely a bug introduced when we harmonized the username/email fields into just one property instead of two.

### This task is done when...
 - [ ] admins can edit a user without changing the user's email address

mgwalker:
Our user validation function optionally accepts a user ID. When it checks it the email address being validated already exists in the database, it will also check the passed-in user ID. That way, if the user being edited has the same email address, it won't fail.

The user update endpoint was not sending the user ID into the validator, so the validation always failed. Furthermore, the database user update function calls the validator and also was not passing along the user ID. Both instances have been fixed.

Also fixed a bug where the admin had to change the user's state, too. This is because the user object in the application store has a full object for `state` rather than just a two-letter ID. The EditAccount component set the user's state to the value from the application store when a user is selected, and that gets sent to the API if the state is not changed. The problem is the API expects a two-letter ID, so it fails validation. The fix is simple: when selecting a user, set their state to the `state.id` from the application store.

- admins can edit user accounts without having to edit their email addresses too
- closes #1973

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,8 @@ Anticipated release: January 6, 2019
 
 #### 🐛 Bugs fixed
 
+- Fixed a bug where an admin could not edit a user's account without changing that user's email address ([#1973])
+
 #### ⚙️ Behind the scenes
 
 # 2.2.0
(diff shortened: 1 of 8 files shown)


CHANGELOG.md
```diff
@@ -1,11 +1,13 @@
 # Next release
 
-Anticipated release: January 6, 2019
+Anticipated release: January 6, 2020
```
hbillings: boom 2020

PR #1983, opened 2019-12-30
Remove the white blob at the bottom of card-form pages

### LINKED ISSUE #1766: Get rid of the conditional white space under gray card backgrounds
On larger screens, the gray background has a white gap. The easiest fix is to set the gray to 100vh, but that'll push the footer down. 

### This task is done when...
 - [ ] figure out what to do
 - [ ] do it

mgwalker:

- adds the `site--gray` class to the app body on page routes that are card forms (login, admin, etc.)
- adds a style rule for `.site--gray` that makes the background gray so it will be gray instead of white. Gray.
- closes #1766


web/src/components/Header.js
```diff
@@ -53,11 +53,9 @@ class Header extends Component {
         <div className="ds-l-container">
           <div className="ds-l-row">
             <div className="ds-l-col--12 ds-l-md-col--4 site-title">
-              {showSiteTitle && <Link to="/">{t('titleBasic')}</Link>}
-              {!showSiteTitle && !authenticated && (
+              {showSiteTitle || !authenticated ? (
```
mgwalker: I had to hop in here when updating the tests for the `Wrapper` component to make sure I understood what it was supposed to do and... this old logic was so hard to make sense of. It covered too many convoluted cases:

- if `showSiteTitle` is true, show the site title
- if `showSiteTitle` is false AND `authenticated` is false, show the site title
- if `showSiteTitle` is false AND `authenticated` is true, show a link to the dashboard

These conditions are exhaustive, so they would be more clearly modeled as an `if/elseif/else`. However, the first two conditions have the same result! We could replace those two conditions with one, and just have an `if/else`:

- if `showSiteTitle` is true OR (`showSiteTitle` is false AND `authenticated` is false), show the site title
- otherwise, show a link to the dashboard

But that looks silly, a true and false on the same variable. That brought back memories of boolean algebra in college, but I couldn't remember all the transitive or associative or whatever else properties. I did remember truth tables, though, so I wrote one of those out, and it ends up that you can simply the above to just this:

- `showSiteTitle` is true OR `authenticated` is false, show the site title
- otherwise, show a link to the dashboard

So yeah, I did all that and turned it into a ternary and hopefully it's easier to read now.

web/src/components/Wrapper.js
```diff
@@ -1,41 +1,43 @@
...
   children: PropTypes.node.isRequired,
-  isDev: PropTypes.bool.isRequired
+  isDev: PropTypes.bool.isRequired,
+  location: PropTypes.shape({
+    pathname: PropTypes.string.isRequired
+  }).isRequired
 };
 
-export default Wrapper;
+export default withRouter(Wrapper);
```
mgwalker: Getting the location from the redux state means that the wrapper will re-render when the location changes, so the gray class will appear/disappear as appropriate. It also turns out this fixes a bug with the header not switching from the site title/dashboard link correctly.

web/src/components/Header.js
hbillings: good work

That's some good logic!! Although I did twitch when you mentioned college algebra.

web/src/components/Wrapper.js
```diff
@@ -1,41 +1,43 @@
...
-import React, { Component } from 'react';
+import React, { useEffect } from 'react';
+import { withRouter } from 'react-router';
 import Header from './Header';
 import Footer from './Footer';
 
-class Wrapper extends Component {
-  componentDidMount() {
-    const { isDev } = this.props;
+const grayPaths = ['/create-account', '/edit-account', '/login', '/me'];
```
hbillings: The wizened paths of wisdom, these are.

I get why we're doing this this way, but the hardcoding isn't my favorite. I suppose it's better than introducing contexts or something, though.

web/src/components/Wrapper.js
```diff
@@ -1,41 +1,43 @@
...
+
+const Wrapper = ({ children, isDev, location: { pathname } }) => {
+  useEffect(() => {
     if (isDev) {
-      this.addTota11y();
+      const script = document.createElement('script');
+      script.src = '/_dev/tota11y.min.js';
```
hbillings: We probably need an issue to remove this, since I don't think we've actually used it in recent memory.

web/src/components/Wrapper.js
mgwalker: I agree wholeheartedly. I was trying to think how to do it and the only things I could come up with were having an established list of "gray paths" or contexts, and I also didn't love trying to figure out how to back-propagate the context.

We might be able to do it in the `<Routes>` component, though. I'm not 100% sure where that one renders in the DOM, but it could access an `isGray` property defined on the route itself and that might be cleaner?

web/src/components/Wrapper.js
mgwalker: Agreed

web/src/components/Wrapper.js
```diff
@@ -1,41 +1,44 @@
...
+import React, { useEffect } from 'react';
+import { withRouter } from 'react-router';
 import Header from './Header';
 import Footer from './Footer';
+import routes from '../routes';
 
-class Wrapper extends Component {
-  componentDidMount() {
-    const { isDev } = this.props;
+const cardRoutes = routes.filter(r => r.isCard).map(r => r.path);
```
mgwalker: Ta-da!

Issue #1981, opened 2019-12-23
Implement Updated CMS Design System


jeromeleecms:
We should be able to implement the CMS Design System as it gets updated without losing functionality within the application.

### This task is done when...
 - [ ] New CMS Design System has been implemented
 - [ ] We are able to have a path/process for updating the Design System as it gets released (and it is documented or automated - like NASA)

mgwalker:
We currently use 1.32.1. The latest version of the CMS design system is 3.4.2.

PR #1984, opened 2019-12-31
Update dependencies

mgwalker:

- updates all in-range dependencies
- updates compatible major versions
- removes unused dependencies 🎉
- fixes an issue where the cacheing on `index.html` was not being set correctly (unrelated AWS config fix)

#### Frontend
- CMS design system - breaking changes (see #1981)
- eslint-config-airbnb - Has new rules that our code doesn't conform to. Some of them are very valid and we should update the code, but most of them we should configure to ignore, IMO. I'll create an issue for that one.
- file-loader - Switched to generating ES6 modules by default and the option to disable that behavior doesn't seem to be working, so our webpack build step fails.
- storybook - Might be able to upgrade, but since we're not using it right now, I didn't want to spend a lot of time figuring out what changed.

#### Backend
- winston - The changes in 3.x would require a pretty substantial investment in our custom wrapper code. Maybe someday? But for now this feels pretty low priority to me. 🤷‍♂


web/webpack.config.dev.js
```diff
@@ -35,13 +35,18 @@ const config = {
...
+              }
+            }
           }
         ]
       },
       {
         test: /\.(woff2?|ttf|otf|eot|svg)$/,
-        loader: 'file-loader'
+        loader: 'file-loader',
+        options: { esModule: false }
```
hbillings: I'm curious about this.

mgwalker: Oops, forgot to roll that back. I was trying to get the `file-loader` update to work, but it wouldn't. So there's no reason for this now.

PR #1986, opened 2019-12-31
Update design system to latest version

### LINKED ISSUE #1981: Implement Updated CMS Design System

mgwalker:

- updates the design system to 3.4.2
- replaces `<Select>` with `<Dropdown>`
- switches `fieldRef` to `inputRef` and swaps it with a function, since that's now required by the design system libraries
- updates `autoFocus` props to be booleans instead of strings, to stop React complaining
- updates the link colors in the header to include a `:visited` pseudoselector so they don't turn purple and disappear
- closes #1981

Basically need a quick run through to make sure things still look and behave the same!


web/src/containers/activity/CostAllocateFFP.js
```diff
@@ -1,4 +1,4 @@
-import { ChoiceList } from '@cmsgov/design-system-core';
```
hbillings: I hadn't even realized there were two ways of doing this. Interesting!

PR #1994, opened 2020-01-06
Update to latest airbnb eslint rules

### LINKED ISSUE #1985: Upgrade airbnb eslint rules on the frontend
Couldn't update the airbnb lint rules for the frontend along with the others. See #1984.

### This task is done when...
 - [ ] the code is fixed to address the issue related to initializing component state in a constructor
 - [ ] the eslint config is updated to disable the `Fragment` and props spread rules

mgwalker:
Updates to the most recent version of the airbnb eslint rules, and tweaks the code accordingly

- The rules now require that component state initialization happen in a constructor. Added constructors where necessary, and converted to functional components where possible
- The Airbnb rules require using `<>` instead of `<Fragment>`. Inverted the rule so that we require `<Fragment>` instead of `<>`
- The rules forbid using prop spreads in components like `<Component ...props>`. However, we use that pattern a lot for wrapper components so we can just blindly pass properties along. Disabled this rule.
- Added local disables for the `no-dangerous` rule that warns about using `dangerouslySetInnerHTML` since we aware of those and they're intentional. We'll still get warnings about new ones, which is desirable
- closes #1985 

### DIFF
--- a/web/src/components/ConsentBanner.js
+++ b/web/src/components/ConsentBanner.js
@@ -1,8 +1,8 @@
 import { Button } from '@cmsgov/design-system-core';
 import PropTypes from 'prop-types';
-import React, { Fragment, PureComponent } from 'react';
+import React, { Fragment, useState } from 'react';
 
-let cookie = name => {
+const cookie = name => {
   const cookieMap = (document.cookie || '').split(';').reduce((c, s) => {
     const bits = s.trim().split('=');
     if (bits.length === 2) {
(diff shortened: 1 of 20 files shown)


web/src/components/ConsentBanner.js
```diff
@@ -1,8 +1,8 @@
 import { Button } from '@cmsgov/design-system-core';
 import PropTypes from 'prop-types';
-import React, { Fragment, PureComponent } from 'react';
+import React, { Fragment, useState } from 'react';
 
-let cookie = name => {
+const cookie = name => {
```
mgwalker: The old way would read the cookies when the module was loaded, parse them, and cache that for eternity. That made the testing more difficult - we need to unload and reload the module for each test. It was meant to be an efficiency thing, but this cookie parsing will only happen on the login page, so it doesn't really seem worth optimizing. I took out this cacheing so it'd be easier to read and also easier to test.

web/src/components/ConsentBanner.test.js
```diff
@@ -2,14 +2,13 @@ import { shallow } from 'enzyme';
...
+import ConsentBanner from './ConsentBanner';
+
 describe('Consent banner component', () => {
-  let ConsentBanner;
   const onAgree = sinon.spy();
 
   beforeEach(() => {
-    jest.resetModules();
     onAgree.resetHistory();
-    ConsentBanner = require('./ConsentBanner').default; // eslint-disable-line global-require
```
mgwalker: All this module reloading stuff can go away now.

web/src/components/PasswordWithMeter.js
```diff
@@ -5,7 +5,10 @@ import zxcvbn from '../lazy/zxcvbn';
 import Choice from './Choice';
 
 class Password extends Component {
-  state = { showPassword: false, strength: 0 };
+  constructor() {
```
mgwalker: I was able to convert this to a functional component but then the tests absolutely would not pass. There's still an issue with Enzyme's support for the `useEffect` hook, and in this component, `useEffect` would alter the state, triggering a re-render. But Enzyme would bail out before the second render and then tests would fail. Rather than spend all day trying to figure that out, it seemed better to just add a constructor.

web/src/containers/SaveButton.js
```diff
@@ -25,87 +25,76 @@ const getButtonContent = (dirty, working) => {
...
-    // success state even though it wouldn't be accurate anymore.
-    if (hasFetched) {
-      const success = !working && !error;
-      return { success };
-    }
-    return null;
-  }
+  // Success has to be derived.  It can't be stored in the app state because
+  // if it was, then the next time this form was loaded, it would show the
+  // success state even though it wouldn't be accurate anymore.
```
mgwalker: One cool thing is that with functional components, you don't need pre-render lifecycle methods like `getDerivedStateFromProps`. You can just... derive the state from the props, and it's much clearer what's going on.

web/src/components/ConsentBanner.js
```diff
@@ -11,116 +11,108 @@ let cookie = name => {
...
   }, {});
 
-  cookie = n => cookieMap[n];
-
   return cookieMap[name];
 };
 
-class ConsentBanner extends PureComponent {
-  state = { showDetails: false };
+const ConsentBanner = ({ onAgree }) => {
```
hbillings: Yay for functional components!

web/src/components/ConsentBanner.test.js
hbillings: aw yeah

web/src/components/PasswordWithMeter.js
hbillings: "Upgrade this, but only if it doesn't also do that, and only during a full moon on a Tuesday."

web/src/containers/SaveButton.js
hbillings: The functional components are so much easier to follow in general -- I'm really excited about this PR!

web/src/containers/admin/EditAccount.js
```diff
@@ -55,14 +92,14 @@ class EditAccount extends Component {
             label="Name"
             name="name"
             value={name || ''}
-            onChange={this.handleEditAccount}
+            onChange={changeUserName}
```
hbillings: These are all so much clearer than the previous version! Imagine, actually being able to know what something is going to do!

web/src/stories/content/Instruction.js
```diff
@@ -4,7 +4,10 @@ import RenderViewbox from '../RenderViewbox';
 import Instruction from '../../components/Instruction';
 
 class InstructionStory extends Component {
```
hbillings: Is this still a class for the same reasons as the one above (`useEffect` and tests being funky together)?

mgwalker: It's because it's in Storybook and so in my calculus the effort to change it wasn't worth nearly as much, so I didn't do it. 😛

mgwalker: It's because in the Storybook and I didn't think it was worth fixing, tbh.

PR #2003, opened 2020-01-09
Change APD name

### LINKED ISSUE #1993: Add new unique IDs to the dashboard and the builder
Carrying over from the design issue: #1821

The current format for the unique identifiers that keys off of the created date isn't intuitive for users and is confusing for CMS because they use the submitted date. Here is the updated version for the IDs based on a discussion with Jerome:

**Dashboard**

**Builder**

### This task is done when...
 - [ ] The dashboard and builder are updated with the new formatting for IDs

mgwalker:

- changes the APD name to `HITECH IAPD` (backend)
- update APD API endpoints to include `created` property (backend)
- added data migration to set existing APD names to `HITECH IAPD`
- updates the dashboard to include the APD creation date
- updates the APD builder to display the creation date and prepend `FFY` to the APD year range
- closes #1993 


api/routes/apds/get.endpoint.js
```diff
@@ -60,7 +60,7 @@ describe('APD endpoint', () => {
       expect(body).toMatchSnapshot();
     });
 
-    describe('with authenticated as a user without a state', () => {
+    describe('with authenticated as a user with a state', () => {
```
mgwalker: 😬

lauraponce:
When states first load the APD, it'll be blank (or carried over from a prior year) so the FFY date might not be showing correctly. Is it possible to leave it as HITECH IAPD without the `| FFY` _until_ they add in the date?

This would be out of scope for this PR, but it's relevant to this — what are your thoughts on moving up the question where we ask which dates the APD covers to the very top above contact info as the first step? That way, they would see the year(s) added to the top right away, and it also kind of just seems like a logical first step of the APD. 

Screen Shot 2020-01-09 at 4 42 57 PM

mgwalker:
@lauraponce By default, a new APD is set to the current FFY and the next one, so if you create a new one now, it'll be `FFY 2020-2021` when you first open it. When they're carrying over from a prior year... 🤔 I guess we don't have that yet, but we could do the same thing, where by default it gets the current FFY plus one? But if the *most right* thing is to not have FFYs by default, then yes, we can have that just not display until they pick some years.

And I agree with the idea of moving the FFY boxes up!

lauraponce:
On the related issue about the year picker, @jeromeleecms what's your perspective on moving up where states choose the FFYs that the APD covers to the top? (see my comment from yesterday ^)

api/routes/apds/get.endpoint.js
hbillings: accuracy, who needs it?

PR #2022, opened 2020-01-23
Installation review

ondrae:
This PR adds a few small changes to the README, to make it easier for a new engineer to do a clean install. These few changes are what I was able to make on my own. Below are the rest of my notes from my installation review.

## Installation Notes
- Very well written README 🥇
- Docker process worked clean on the first try.
- Docker process needs to be running to do the following docker-compose exec migrations and seeding. Added a line about that to the README.
- The testing docs are 404. I see a PR that fixes that though by pointing to the wiki. All the tests pass.

## Manual Installation
- I had to first create the `hitech_apd` before I could run the migrations. If that is right, then it should be added as a step in the instructions. I'm not sure what user and pw it needs though, so I'll let you all make that addition.
- The web process started up on on port 8001 for me. Changed it in the README.
- The seed file doesn't seem to be giving me any users. I've tried remaking the database as the postgres user and enforcing the cms password. Nothing seems to work.
- Could not get the Storybook to work either. I suspect the same db problems as above causing it. Here is the error: `ValidationError: Invalid options object. Sass Loader has been initialised using an options object that does not match the API schema.`

### DIFF
--- a/README.md
+++ b/README.md
@@ -61,7 +61,7 @@ then run `docker-compose up`. This will do several things:
 
 This could take a few minutes. Once it's finished, everything is installed,
 configured, and running. However, the database will still be empty, so that
-needs to be created and populated with starter data. To do that, run:
+needs to be created and populated with starter data. To do that, while the docker process is still running, open a new terminal window and run:
 
 ```shell
 docker-compose exec api npm run migrate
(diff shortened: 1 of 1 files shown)


mgwalker:
This is super. Thanks @ondrae!

#2006 incorporates some changes from this PR:
- docker process needs to be running
- fixes links to other docs
- updates database configuration so username and password can be provided
- removes references to Storybook; we still have some Storybook components, but we've punted on trying to keep all of that up-to-date and working

Also updated the wiki to incorporate some other bits:
- create `hitech_apd` database
- optionally create a unique user
- configure the user in addition to the database host

Still need to dive into the issue with the seed not working locally.

mgwalker:
Closing as done. Everything was incorporated into either #2006 or wiki changes.

Thanks again @ondrae!

PR #2030, opened 2020-01-27
Have the form/review component determine the collapsed status of items itself rather than items having to keep track of that information

### LINKED ISSUE #2026: Newly-added activity pieces are open every time the activity modal re-opens
We have an `initialCollapsed` flag, which is based on the assumption that the form/review components will only be loaded once and thereafter just hidden in the DOM. We read that flag on component load and then never again, so everything works if that assumption holds. But that's not how it works now. The modal unloads from the DOM and so every time you open the activity, you will see every newly-added form/review item expanded..

### This task is done when...
 - [ ] newly-added bits are only open when they are created; subsequently opening the modal will have them collapsed

mgwalker:
The form/review list component was originally built around the assumption that it would only be created/loaded once per list. It relied on an `initialCollapsed` value living inside each element in the list to determine whether items should be expanded on that first render. The `initialCollapsed` is true for items that exist when the APD is loaded from the server, but is set to false for newly-created items. Once the item was rendered the first time, its collapsed state would be managed by the form/review component. That was fine until we switched to the modal. The modal causes the list component to be removed from the DOM, so it gets unloaded and reloaded later. The result is that any item added since the APD was originally fetched from the server show as expanded, even though they should be collapsed.

This PR removes the `initialCollapsed` value from items and instead uses the form/review list component's internal state to determine whether to expand an item initially. In a nutshell, if the user clicks to add a new item, the last item in the list will be expanded.

- manage initial collapse internally instead of having all these items try to track their own state
- closes #2026

  - N/A ~Associated OpenAPI documentation has been updated~


web/src/components/FormAndReviewList.js
```diff
@@ -82,7 +88,7 @@ const FormAndReviewList = ({
             collapsedComponent={collapsed}
             expandedComponent={expanded}
             index={index}
-            initialCollapsed={item.initialCollapsed}
+            initialCollapsed={!(hasAdded && index === list.length - 1)}
```
mgwalker: Debating whether to switch these up so it's `initialExpanded` instead, so the logic will be easier to think about. The `!` at the beginning here is easy to miss, and getting rid of it also causes a messy equation: `!hasAdded || index !== list.length - 1`

What do you think, @hbillings?

web/src/util/serialization/apd.js
```diff
@@ -41,37 +41,33 @@ export const fromAPI = apdAPI => {
...
         ...activity,
 
+        key: generateKey(),
+
         // These properties need some massaging into reducer state
         contractorResources: contractorResources.map(c => ({
-          key: generateKey(),
-          ...c
+          ...c,
+          key: generateKey()
```
mgwalker: Invert these so the local `key` will always override whatever we get from the server. It shouldn't send a `key` property at all, but this is a safeguard.

web/src/components/FormAndReviewList.js
hbillings: It's not _awful_ to reason about, but I did have to read it a couple of times to make sure I got it. I'd be in favor of switching it to `initialExpanded`.

mgwalker: Switched!

PR #2031, opened 2020-01-27
When images are uploaded, make sure the rich text content gets saved

### LINKED ISSUE #2017: When an APD with images embedded in rich text fields is loaded, the images are uploaded but the text field is not updated
Rich text fields used to support image uploads by embedding the image data directly into the text field. We switched to uploading images in #1953, and we insert an `` tag instead. One side-effect of that code is that when a rich text field is loaded, if its initial contents include an embedded image, we immediately upload that image. However, the embedded image is not removed from the rich text field and and the `` tag is not added. The result is that images are uploaded and never used, and the same image will be uploaded again every time the APD is loaded.

### This task is done when...
 - [ ] rich text fields are updated and saved after embedded images are uploaded

mgwalker:
When an image is added to the rich text editor with drag-n-drop, an `` tag gets added to the text content with the image embedded in it. A change event is triggered and the content is saved. When the image upload is finished, the `` tag is changed – the embedded image is removed and is replaced with a URL, but the editor does not trigger a change event and the content with the image URL is not saved (unless the user makes another change in the editor).

Similarly, if an APD already has an embedded image, when it is loaded, the rich text editor will upload that image. It will then replace the embedded image with the URL but it does not trigger the change event so the change is not saved.

This PR updates the image upload behavior so that it manually triggers the change event so that the embedded image is removed and is replaced with the URL.

- trigger the change event when images are uploaded either because they are embedded on load or from drag-n-drop
- closes #2017 

  - N/A ~Associated OpenAPI documentation has been updated~


web/src/components/RichText.js
```diff
@@ -53,10 +53,19 @@ const setupTinyMCE = upload => editor => {
...
   upload(blob.blob())
     .then(url => {
       success(url);
+      /* global tinyMCE */
+      // ☝️ lets eslint know that tinyMCE is defined. Once we've told Tiny the
+      // URL of the uploaded image, it'll update the content of the textbox,
+      // but it does NOT call the change event. That means that dropped images
+      // aren't saved as IMG tags unless the user makes additional changes to
+      // the textbox content. To get around that, we can get a reference to
+      // the editor, pull the content, and manually trigger the event.
```
hbillings: it's all we can do

web/src/components/RichText.js
```diff
@@ -78,15 +87,16 @@ class RichText extends Component {
   };
 
   render() {
-    const { uploadFile: upload } = this.props;
+    const { id, onSync, uploadFile: upload } = this.props;
```
hbillings: Does this `onSync` event already exist in the elements that use the RTE? I don't see where it's provided to the editor.

mgwalker: Yeah, it's a prop they pass into the `RichText` component. Without it, we'd never know the rich text changed. 😄

PR #2039, opened 2020-01-29
[Prototype] Show math

mgwalker:
A prototype branch. Putting it in a PR so we can have a preview link to look at.

## Current state
- shows APD key personnel as uneditable entries in the Program Administration activity's state personnel expense section
- adds an activity cost summary table to the cost allocation table
  - incomplete; we don't have the sum costs for each expense type
- shows math symbols in the quarterly FFP table to help show where numbers are coming from
- adds a narrative paragraph for the activity total cost
  - incomplete; doesn't use real data

### DIFF
--- a/web/src/components/Dollars.js
+++ b/web/src/components/Dollars.js
@@ -52,7 +52,7 @@ const Dollars = ({ children, long }) => {
     return <Fragment>$0</Fragment>;
   }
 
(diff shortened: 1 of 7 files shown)


mgwalker:
Closing. Implementation is happening in #2074

PR #2043, opened 2020-01-30
Validate date fields onComponentBlur instead of onChange

### LINKED ISSUE #2019: Change when/how date fields are validated
Currently the date fields are validated on every keystroke in the day, month, or year fields. As a result, when a user begins to type a new date, it is immediately invalid and stays invalid until they have finished typing the entire thing. The validation error message changes over time as the user types, resulting in repeated and unhelpful messages, which might be especially jarring on a screen reader. Autosave also fails during this time, which will result in multiple warning dialogs while the user types after #2011 is implemented. If those ALSO announce to the screen reader, that makes for a really terrible user experience.

### This task is done when...
 - [ ] users can set dates without being bombarded by validation warnings

hbillings:
Closes #2019. The Design System `DateField` component has a prop that waits until the user has clicked outside of any of the three related fields to fire. If we use that to set the error message instead of watching for the `value` to change, the error message fires all at once when the user leaves the datefield input group.

- use onComponentBlur


web/src/components/DateField.js
```diff
@@ -29,19 +29,33 @@ const joinDate = ({ day, month, year }) => {
...
+    errorMessage: [],
+    dayInvalid: false,
+    monthInvalid: false,
+    yearInvalid: false
+  });
+
   const dateParts = splitDate(value);
 
-  const errorInfo = useMemo(() => {
+  const getErrorMsg = () => {
```
hbillings: ESLint is complaining about this not returning; I'm not sure if I need to suppress the warning or rework this somehow.

mgwalker: You can change the two `return null;` (lines 44 and 54) to just `return;` and that should fix it.  The problem is that the function *sometimes* returns a value (`null`), but sometimes doesn't return anything. It needs to either never return anything or always return something.

PR #2048, opened 2020-02-05
Debounce autosave

mgwalker:
Autosave was intended to be debounced. I had even written the code. I don't know where it went, but it didn't make it into the original PR. This PR puts it back, as intended.

Debouncing autosave solves two problems:
1. If a user is typing quickly, we don't want to trigger a save on every single keypress because that could be a big drain on their computer's performance. Instead, we wait until 300 ms after the last keypress before we send the save.
2. If multiple saves are happening at once, we can't be sure what order they happen in on the server, so the final state of the APD becomes unknown. Instead, we only allow a single in-flight save at a time. Any additional save attempts while one is already in-flight get squashed into a single save enqueued to run when the previous save completes.

- debounces autosave


web/src/saveMiddleware.test.js
```diff
@@ -21,15 +21,114 @@ describe('automatic save middleware', () => {
...
+        expect(runSave).toHaveBeenCalled();
+      });
+    });
+  });
+
+  describe('debounces saves', () => {
+    const DEBOUNCE_TIME = 300;
+
+    beforeEach(() => {
+      jest.useFakeTimers();
```
hbillings: Wow, I didn't realize timer mocking was built-in to Jest! That's cool.

PR #2053, opened 2020-02-06
Remove activity list in favor of editing names/funding source in modal

hbillings:
The new thing looks like this:
Screen Shot 2020-02-06 at 4 37 27 PM

When "Add another activity" gets clicked, the activity will be appended to the list with the title `New Activity` and the default funding source (HIT). (I think maybe `Untitled` would be better, because it is possible to delete an activity name.) Currently the modal doesn't open automatically. I'd like to have a discussion about desired behavior.
Screen Shot 2020-02-06 at 4 39 29 PM

- removes activity list
- moves name and funding source into the modal
- moves remove functionality to `EntryDetails.js`


hbillings:
@mgwalker It seems like this might have broken the autosave somehow. I'm not really sure. 😬

jeromeleecms:
@hbillings I forgot to mention it today, but if memory serves, Activity 1 should not have the option to be deleted (@mgwalker does it break anything if we do delete Activity 1)?

mgwalker:
Oh yes, it breaks a bunch! 😬

mgwalker:
I think the autosave fails because the activity name is required by the schema on the backend. We can either make the name not-required or we can give activities a default name when they are created (e.g., something like `Activity #7`).

As an aside, this makes me wonder, if activity names are still guaranteed to be unique within an APD. They used to be, but there was a separate code path for adding activities. I think now the backend will allow activities to have the same name. If activity names should be unique, we should create a bug to fix that.

mgwalker:
Spoke too soon. It's because the new activity being posted doesn't have the right `standardsAndConditions` object. That thing got changed out a while back, but I guess the activity reducer that is responsible for building new activity objects didn't. Creating new activities has been broken for a while, I guess. 😬 Pushed a fix.

web/src/containers/activity/All.js
```diff
@@ -55,8 +34,7 @@ const All = ({ add, first, other, remove }) => {
 All.propTypes = {
   add: PropTypes.func.isRequired,
   first: PropTypes.object.isRequired,
-  other: PropTypes.arrayOf(PropTypes.object).isRequired,
-  remove: PropTypes.func.isRequired
+  other: PropTypes.arrayOf(PropTypes.object).isRequired
 };
 
 const mapStateToProps = state => {
```
mgwalker: We don't need the distinction between the first and other activities anymore, so it should be fine to replace `mapStateToProps` with something like:

```js
mapStateToProps = state => ({ activities: selectAllActivities(state) });
```

And update the rest of this component accordingly (e.g., replace `[first, ...other]` with just `activities`).

web/src/containers/activity/EntryDetails.js
```diff
@@ -28,13 +37,20 @@ const EntryDetails = ({ activityIndex, fundingSource, activityKey, name }) => {
...
+  };
+
   const title = useMemo(
     () => makeTitle({ name, fundingSource }, activityIndex + 1),
     [fundingSource, name, activityIndex]
   );
 
   const editContent = (
     <div className="nowrap visibility--screen">
+      <Button size="small" variation="transparent danger" onClick={onRemove}>
```
mgwalker: We don't want this delete button on the first activity

web/src/containers/activity/EntryDetails.js
```diff
@@ -28,13 +37,20 @@ const EntryDetails = ({ activityIndex, fundingSource, activityKey, name }) => {
     setShowModal(false);
   };
 
+  const onRemove = () => {
+    remove(activityIndex);
```
mgwalker: Yay, no more activity key!

web/src/containers/activity/EntryDetails.test.js
```diff
@@ -1,22 +1,37 @@
...
+    remove: jest.fn()
   };
 
+  beforeEach(() => {
+    props.remove.mockClear();
+  });
+
   test('renders correctly with the modal closed', () => {
     const component = shallow(<EntryDetails {...props} />);
     expect(component).toMatchSnapshot();
```
mgwalker: I think it'd be useful to have a separate test here where `activityIndex` is zero. Just a simple snapshot test, to make sure the delete button isn't rendered in that case.

web/src/containers/activity/EntryDetails.test.js
```diff
@@ -1,22 +1,37 @@
...
 
     const review = component.find('Review').dive();
-    review.find('Button').simulate('click');
+    // First button is the delete button; second is the edit button
+    review
+      .find('Button')
+      .at(1)
+      .simulate('click');
```
mgwalker: Another separate test for button 0, to verify that the `props.remove` mock is called after it's clicked would be good.

web/src/containers/activity/NameAndFundingSource/NameAndFundingSourceForm.js
```diff
@@ -15,16 +15,11 @@ const NameAndFundingSourceForm = ({
   setName
 }) => {
   const changeName = useCallback(({ target: { value } }) => {
-    // These forms are built from the list of editable activities, which is
-    // all activities except the first one. So the local list index starts at
-    // 0, but it corresponds to index 1 in the full list. To handle that, just
-    // increment the local index by one when updating the activity.
```
mgwalker: so long

lauraponce:
It's so nice to have only one list to worry about! 😅 It makes quite a difference. 

I agree that `Untitled` might be a better placeholder for the name of the activity until it gets edited or even leaving it blank in the activity list until a name gets added.

I know this is how it currently works in the app but I don't recall if there was a reason behind it —  why do we autoselect the funding source for states (HIT/HIE/MMIS) rather than let them select, @jeromeleecms?

@hbillings the thing that grabs my attention when looking at the list is the red delete text, I wonder if we could be more subtle about it since it's not something that users are needing to do _that_ often. And they would be editing it the majority of the time so making the most of that space could make it easier to select. What are your thoughts on something along these lines?

Screen Shot 2020-02-11 at 3 22 54 PM

Screen Shot 2020-02-11 at 3 23 44 PM

Screen Shot 2020-02-11 at 3 24 15 PM

lauraponce:
Created https://github.com/18F/cms-hitech-apd/issues/2059 as per our conversation in parking lot time for changing how the default selection works for the program type.

PR #2061, opened 2020-02-12
[Prototype] Activities in pages with tabs

mgwalker:
Opens activities in a separate page instead of a modal. Keeps the tab bar.

This is only a prototype and should never be merged.


mgwalker:
Since we're not moving forward with this design, gonna go ahead and close this one down.

PR #2062, opened 2020-02-13
[Prototype] Activities in pages with each section as a separate page

mgwalker:
Opens activities in a separate page. Each section is also a page. When looking at an activity, the sidebar expands to show all of the activity sections for navigation.

This is only a prototype and should never be merged.


mgwalker:
I don't think we need this one any more.

PR #2065, opened 2020-02-18
Internal name fix for budget state data

### LINKED ISSUE #2044: [Tech debt] In the budget state data, rename "state" to "inHouse"
In the budget data, we call the rolled-up state costs `state`, but in the APD, they are called `in-house`. We should rename the internal state to match – that will make it easier to trace the data moving through the app and easier to debug/test.

The current naming is also confusing because sometimes `state` means the state's share of a cost, and sometimes it means the in-house costs. Those aren't the same thing!

### This task is done when...
 - [ ] "in-house" data is called "in-house" rather than "state"
 - [ ] state-share costs are still called "state"

mgwalker:

- updates budget state to use the name `inHouse` for in-house costs rather than `state`, to be less confusing
- updates FFP tables that display in-house costs to use updated names
- closes #2044 


web/src/reducers/budget.js
```diff
@@ -458,18 +458,19 @@ const buildBudget = incomingBigState => {
       const quarterlyFFP = newState.federalShareByFFYQuarter[ffpSource];
 
       ['contractors', 'expenses', 'statePersonnel'].forEach(prop => {
-        const propCostType = prop === 'contractors' ? 'contractors' : 'state';
+        const ffpType = prop === 'contractors' ? 'contractors' : 'state';
```
hbillings: Does `state` in this instance refer to something different than the `state` that `inHouse` replaced?

mgwalker: Kind of yes, kind of now. It's the same type of information, but this one comes from the user input rather than the budget calculations. There's #2066 to finish out the change, but the two seemed big enough to be separate PRs.

PR #2071, opened 2020-02-18
Removes the API/client serialization module

### LINKED ISSUE #2021: [Tech debt] Remove "serialization" module from frontend
Once upon a time, the frontend and backend had very different data structures. To manage that, we created a "serializer" that would do the conversion back and forth. They are now the same data structures, but the frontend still has its serializer for appending frontend-specific properties, such as unique keys (for React) and flags indicating whether the component should be expanded by default on page load.  This isn't very complicated logic, but it's confusing that it lives in a weird folder under `utils/serialization/apd.js`. This code should be moved to either the APD reducer or the reduc action that receives APDs from the API.

### This task is done when...
 - [ ] the "serialization" code is moved to a better home
 - [ ] there are tests
 
mgwalker:
We used to have to translate between the backend and frontend data structures. We built a serialization module that handled that translation. Now that the two sides use a normalized data structure, "serialization" doesn't make sense anymore. This PR removes the last little bits of functionality from the serialization module into the APD reducer (adding unique, transient `key` properties to the pieces that need them) and then tosses that module out.

- ☝️ that
- and also adds some tests to cover a few more cases in the APD reducer
- closes #2021


web/src/actions/app/apd.js
```diff
@@ -28,7 +28,7 @@ import {
 import { getIsAdmin } from '../../reducers/user.selector';
 
 import axios from '../../util/api';
-import { fromAPI, initialAssurances } from '../../util/serialization/apd';
+import initialAssurances from '../../util/regulations';
```
hbillings: :chef_kiss:

web/src/reducers/apd.js
```diff
@@ -381,11 +384,53 @@ const reducer = (state = initialState, action) => {
...
+              schedule: schedule.map(milestone => ({
+                ...milestone,
+                key: generateKey()
+              })),
+              statePersonnel: statePersonnel.map(person => ({
+                ...person,
+                key: generateKey()
+              })),
+              key: generateKey()
+            })
```
hbillings: beautiful

PR #2080, opened 2020-02-25
Do not give new activities a funding program by default

### LINKED ISSUE #2059: Leave the default program type selection blank for new activities
Currently, we select HIT by default for new activities but this can lead to mistakes if a state user overlooks this and doesn't change it to the correct one if it's HIE or MMIS. We should update this so that no radio button is selected by default.

### This task is done when...
 - [ ] In the Activity Overview section, the default selection is blank for program type (HIT, HIE, MMIS)

mgwalker:
Previously, new activities were assumed to be HIT programs. This PR changes it so new activities are not automatically associated with any program

- new activities get `false` for their default program
- activities without a program do not contribute to program-level budget data, but they still count towards APD totals
- closes #2059


mgwalker:
Yes, one of each by default is intended.

Back in the olden times, every new activity began with I think 3 blank state personnel, 3 blank contractors, and 3 black non-personnel expenses. Then we scaled it back to just 1 of each. We can totally go down to zero if that makes more sense, though. It's an interesting thought experiment how different navigation types influence what should be displayed by default!

web/src/reducers/apd.test.js
```diff
@@ -751,7 +751,7 @@ describe('APD reducer', () => {
                     }
                   }
                 ],
-                fundingSource: 'HIT',
+                fundingSource: false,
```
hbillings: It feels a little weird to set this to false and then set it to a string later on. Would it make more sense to make this an empty string? (I realize then we'd have to do a `if (fundingSource !== '')` later, which isn't super nice.) Just thought I'd flag it!

web/src/reducers/budget.test.js
```diff
@@ -878,13 +927,159 @@ describe('budget reducer', () => {
...
+              state: 255,
+              total: 900
+            }
+          },
+          quarterlyFFP: {
+            '1931': {
+              // You might intuit that that the in-house cost per quarter
+              // should be $25 since the total cost for the year is $100 and
+              // each quarter wants 25%. That'd make sense, right? Well let
+              // me teach you about how these budgets are built up.
```
hbillings: `That'd make sense, which means it cannot possibly be how this works.`

web/src/reducers/apd.test.js
mgwalker: Thankfully we could still do a truthy check because the empty string is falsey. I waffled on this exact question for a bit before deciding that `false` felt like it more strongly indicated that no value was set. Though now that I type that out, `null` is probably the even better answer.

mgwalker: Let's talk about this again in parking lot time next week. I'm going to go ahead and merge the PR but I agree we should revisit this in particular for code consistency's sake.

Issue #1792, opened 2019-09-05
Show the math that is happening in the background


lauraponce:
Through user research, we've learned that users love automatic math (yay!), but they want to understand what calculations are happening behind the scenes.

These are the three main tables that could use some additional clarification for what math is happening.

1.  Summary Budget table:

![Screen Shot 2020-01-24 at 3 36 45 PM](https://user-images.githubusercontent.com/21222704/73111497-5c833600-3ebf-11ea-8625-088dc409cbc1.png)

![Screen Shot 2020-01-24 at 3 38 14 PM](https://user-images.githubusercontent.com/21222704/73111563-95bba600-3ebf-11ea-89b9-06612bc6b20e.png)

2. Program Budget table (in Executive Summary):

![Screen Shot 2020-01-24 at 3 39 29 PM](https://user-images.githubusercontent.com/21222704/73111607-bedc3680-3ebf-11ea-9994-36c1f754692e.png)

3. FFP Cost Allocation

![Screen Shot 2020-01-24 at 3 41 56 PM](https://user-images.githubusercontent.com/21222704/73111669-137fb180-3ec0-11ea-9c62-1f3bd14a845c.png)

### This task is done when...
 - [ ] There is a mock up of showing the math
 - [ ] The design has been shared with the team and approved by product

lauraponce:
Note from #1935

Add a short narrative to explain cost calculation in paragraph format, for example:
>The total cost of the Auditing activity is $1,000. Because of other funding of $100, the total cost to Medicaid is $900. This activity is using a 90/10 funding split, resulting in a federal share of $810 and a Delaware share of $90.

lauraponce:
I'm approaching this issue table by table since there's some variance in the types of calculations happening between the different tables. 

@jeromeleecms what are your thoughts on something like this?:

![Screen Shot 2020-01-22 at 5 12 08 PM](https://user-images.githubusercontent.com/21222704/72948557-5adc3580-3d3a-11ea-9b49-3ac5ae4fb9cc.png)

I moved the total amounts to the left since we already know those numbers, and the kept the final calculated amounts in the bottom so it was clearer what the new numbers are. This also includes the `x` and `=` signs to show what math is happening.

lauraponce:
@mgwalker is going to take a crack at some of these

mgwalker:
Some hypotheses:
- If we show a review component of the APD key personnel in the first activity's state personnel section, that will help users understand that's where those costs show up
[...]

PR #2081, opened 2020-02-27
New activity budget and FFP section

### LINKED ISSUE #2074: Build new FFP cost table
See issue #1792, or PR #2039 (prototype implementation) for more details.

screenshot of new FFP cost table

### This task is done when...
- [ ] the FFP section is updated to match the mockup above
- [ ] the tables are tightened up so that the math symbols are closer to the actual numbers, to the extent practicable
   - from @lauraponce: "I'd recommend either moving the operators closer to the number that is used in the calculation instead of centering or even just narrowing the whole table"
- [ ] remove the FTEs header from the state personnel table and add a "Total cost" header

mgwalker:

- switches to the new costs table and FFP design
- changes the print view to be one great big costs table, with other funding and FFP included
- updates internal state data tools
- adds tests
- closes #2074


mgwalker:
@hbillings Would it make sense for you to do a code review on this before taking it on to get it across the finish line?

web/src/containers/activity/CostAllocateFFP.js
```diff
@@ -1,276 +1,369 @@
...
-      aKey: activityKey
-    });
+const mapStateToProps = (
+  state,
+  { activityIndex },
+  {
+    getActivity = selectActivityByIndex,
+    getCostAllocation = selectCostAllocationForActivityByIndex,
+    getCostSummary = selectActivityCostSummary,
+    getState = getUserStateOrTerritory
```
mgwalker: Made these injectable so the tests would be easier.

web/src/reducers/activities.selectors.js
```diff
@@ -87,10 +187,31 @@ export const selectActivitySchedule = createSelector(
...
+          }
+
+          if (+dayA > +dayB) {
+            return 1;
+          }
+          if (+dayA < +dayB) {
+            return -1;
+          }
+
+          return 0;
```
mgwalker: Here's that schedule sorting fix

web/src/reducers/activities.selectors.js
```diff
@@ -30,31 +30,131 @@ export const selectAllActivities = ({
...
+    // Program Administration. For other activities, just returns empty arrays.
+    {
+      apd: {
+        data: { keyPersonnel, years }
+      }
+    },
+    { activityIndex }
+  ) =>
+    years.reduce(
+      (o, ffy) => ({
```
hbillings: What is `o`? Could we maybe add a comment?

web/src/reducers/activities.selectors.js
hbillings: nice!

web/src/reducers/activities.selectors.js
mgwalker: It's the accumulator object for the reducer. I never know what to call it... It doesn't have an inherent name, it's just the temporary thing being built up. But I guess in this case, it's the key personnel, so that seems like a good and decent name for it.

mgwalker:
Listening to it from a screen reader, this is pretty bad. I've got an idea for how to make it a better experience by hiding the table from screen readers and instead having each row of the table become a single element that gets read all at once. Will try that out tomorrow!

jeromeleecms:
Super excited about this one! Just reviewing it, it's a massive amount of work 🥇 

Some things I noticed:

1. The table label doesn't updated based on the allocated amount (the math updates based on the selection in the drop down): 
image

2. The summary also doesn't react to a change in the dropdown (currently selected at 75/25): 

image

On a related note, I totally missed your question in the original issue, for a split Federal / Share by FFY, I would suggest:

"The total cost of the Program Administration activity is $5,459,076. **Because of $338,410 in other funding**, the total cost to Medicaid is $5,120,666. This activity is using a **90/10 funding split (FFY 2019 and FFY 2020) and a 75/25 funding split (FFY 2021)** resulting in a federal share of $4,608,600 and a Missouri share of $512,066."

3. Are the column headers removed on purpose (comparing this to the mockup)? 

4. Does the alignment seem off to you with the Total column?

image

5. I know the mockup of the original table doesn't have it, but can/should there be a dollar value included for the % time employees in the table? In the example from Activity 1, it seems like the person is 30% time but $0 which is a confusing.

image

6. Maybe a super minor detail, but I noticed the indent between the various new tables seems a little off: 

image

jeromeleecms:
Separately -- this is looking GREAT. Very easy to follow the math.

mgwalker:
1 & 2: Oooops. I felt like I was forgetting to hook something up but I couldn't remember what it was. Thanks for the eagle eyes!

3: Yes, per [Laura's comment](https://github.com/18F/cms-hitech-apd/issues/1792#issuecomment-587949024) on the original design issue. Her comments didn't make it into the mockup, but they're capture in the "this task is done" checklist.

4. It's off, but it's behaving according to our standard rules for tables. Header cells are left-aligned, numeric cells are right-aligned. That column is pretty wide, it turns out, so the header is way to the left of the content. I will play around with getting the columns to shrink down to fit their contents. That should also help tighten up the whole table.

5. The APD key personnel are funky because we ask for their cost attributable to the program, not their total cost to the state, and we separately ask for their time commitment. The two numbers are not mathematically correlated. In the data used for the preview site, the second key person has an attributable cost of $0 but a 30% time commitment, so that's what the table shows. We don't have anything that would be the equivalent of a "unit cost" like we do with the activity state personnel, so I left that column blank. Maybe in the future we should change what we ask for APD key personnel to be in line with what we ask for activity state personnel?

6. It's because they're separate tables and I haven't tried to set any of the column widths. As a result, they grow or shrink however the browser thinks is most appropriate based on the content. I can give that first blank column a fixed width, though, and that'll at least align the names of things. In viewonly/print mode, this isn't an issue because it's rendered as one big table.

jeromeleecms:
Sounds good!

For your question on (5), isn't this equivalent to Unit Costs?

image

In the case the user selects "No" - can we insert "$0" as the unit cost?

mgwalker:
In the math, we have always treated the key personnel cost as the total cost attributable to the program, not the key person's total cost to the *state*. (The yes/no radio buttons kind of imply that's what we're asking for, too.) So if someone enters `$50,000` and `50%` commitment, we use $50,000 for their cost in all the math. We don't scale it based on their percent commitment.

So one of two things seems to be true:

1. We are doing it right, in which case there is no "unit cost" for APD key personnel in this table, there is only a total cost. (We could compute a unit cost as `total / commitment`, but I chose not to do that because in the table, it looks like a user input but it's calculated. It's also not a useful number to calculate since it is only for display in this one table.)

2. We are doing it wrong and the cost entered in the key personnel section is that person's total cost to the _state_, not to the program, and we should scale their cost based on their time commitment to the program.

In either case, we probably ought to update the help text to be clear which we're asking for. (And if we're going to tweak help text, can we also explain what it means for someone to have a commitment greater than 0% but also have no costs attributable to the program?)

jeromeleecms:
Makes sense - I'm inclined to leave this as is and spin up a separate issue to address this potential change since it does seem to be going out of scope of scope for this particular Issue/PR.

Does that work?

mgwalker:
@jeromeleecms 
1 and 2: Fixed up the budget summary narrative at the bottom.
3. Nothing to change I think?
4: Tightened up the columns some more so that they take up less extra space, which resulted in better alignment with the header!
5: Punting to a separate issue
6: Removed the blank cell in all those tables since that's not a proper way of getting padding, and instead added padding. Now they all have the same amount of padding, so the first columns all line up!

PR #2096, opened 2020-03-06
Reset appropriate application state when switching APDs

### LINKED ISSUE #2084: Pending saves are not cleared when switching APDs
When the user switches from one APD to another, the list of unsaved changes is not cleared. The result is that the unsaved changes are applied to the second APD the next time the app tries to save.

### This task is done when...
 - [ ] switching APDs causes the list of unsaved changes to be emptied

mgwalker:
This adds a redux action to reset the app state, adds an action for navigating to the dashboard, has the dashboard action trigger the reset action, and updates the header to use the dashboard action instead of linking directly to the dashboard.

The patch, apd, and budget reducers are all updated to respond to the reset action by reverting to an empty state.

This fixes an issue where a set of saves that failed to save to the API would persist when the user navigated back to the dashboard. If they then opened another APD, the app would attempt to persist those lingering saves to the second APD, potentially creating invalid data.

- changes header dashboard links to buttons and hooks their click events to the dashboard navigation action
- fixes #2084 


web/src/components/DashboardButton.test.js
```diff
@@ -0,0 +1,34 @@
...
+} from './DashboardButton';
+import { goToDashboard } from '../actions/app';
+
+describe('the dashboard button', () => {
+  it('renders as expected', () => {
+    const dashboard = jest.fn();
+
+    expect(
+      shallow(
+        <DashboardButton dashboard={dashboard}>clicky clicky</DashboardButton>
```
hbillings: clicky

web/src/components/Header.test.js
```diff
@@ -23,7 +22,7 @@ describe('Header component', () => {
     expect(component).toMatchSnapshot();
   });
 
-  it('opens and closes the dropdown when clicked', () => {
+  it('opens and closes the dropdown when clickedgit ', () => {
```
hbillings: clickedgit?

mgwalker: I bet I was trying to do something with `git` but I had the wrong window focused... 🤦‍♂

PR #2099, opened 2020-03-10
Give the header and sidebar fixed positions

### LINKED ISSUE #2087: Sidebar should scroll independent of the rest of the page
Currently, the sidebar is sticky and so it scrolls up on the page when the header scrolls out of view, and then scrolls up again when the footer comes into view. This can distract users. Instead, make it fixed position so that it does not scroll with the rest of the page.

### This task is done when...
 - [ ] the sidebar scrolls independently of the page

### LINKED ISSUE #2090: Implement fixed position header and autosave messaging
The page header should be fixed instead of scrolling, and should include autosave status messaging. See the design synthesis and prototype for more info.

Screenshot of what it should look like (exact autosave messaging TBD in #2094):

A2
____

### This task is done when...
 - [ ] the header is fixed
 - [ ] there's a save status message in the header

mgwalker:
Changes the header so that it stays at the top of the page at all times, and updates the sidebar so that it is sticky to both the top and bottom instead of only the top. Displays autosave status in the header.

- closes #2087
- closes #2090 

  - N/A N/A ~Associated OpenAPI documentation has been updated~

### DIFF
--- a/web/src/components/ConsentBanner.js
+++ b/web/src/components/ConsentBanner.js
@@ -33,7 +33,7 @@ const ConsentBanner = ({ onAgree }) => {
   }
 
   return (
-    <div className="card--container">
+    <div className="card--container ds-u-margin-top--7">
       <div className="ds-l-container">
         <div className="ds-l-row card">
           <div className="ds-l-col--1 ds-u-margin-left--auto" />
--- a/web/src/components/Header.js
+++ b/web/src/components/Header.js
@@ -1,18 +1,21 @@
 import PropTypes from 'prop-types';
-import React, { Component } from 'react';
+import React, { Component, Fragment } from 'react';
 import { connect } from 'react-redux';
 import { Link } from 'react-router-dom';
 
+import { selectIsSaving, selectLastSaved } from '../reducers/saving';
 import { getIsAdmin } from '../reducers/user.selector';
 import { t } from '../i18n';
 
 import DashboardButton from './DashboardButton';
 
 import Icon, {
+  Check,
   faChevronDown,
   faChevronLeft,
   faEdit,
-  faSignOutAlt
+  faSignOutAlt,
+  Spinner
 } from './Icons';
 
 class Header extends Component {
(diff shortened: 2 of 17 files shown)


mgwalker:
This won't be previewable until #2101 is merged.

mgwalker:
random X:
https://gsa-tts.slack.com/archives/C8E5EJK9V/p1583960520414400?thread_ts=1583956485.411900&cid=C8E5EJK9V

mgwalker:
For 1), it does that because otherwise the sidebar would disappear underneath the footer. I wasn't sure what else to do with it.

For 2), we don't have to worry about it because when the page-based navigation is merged, the highlighting will be fixed! (It doesn't highlight now because the export section is so far down the page that its waypoint never scrolls high enough to get highlighted. It's been that way all along.)

lauraponce:
I had a similar expectation to you, Jerome about the sidebar but the footer turns it into an interesting problem. 

Here's a quick example of how it could look if we kept the sidebar where it is and fit the footer into all of the main content area. I left some space under the sidebar at the bottom because it felt too cramped to go all the way to the edge. Once the paginated view is merged, the export section will be pretty small. 

@jeromeleecms is this what you were picturing?

footer2

jeromeleecms:
@mgwalker @lauraponce 

1) Hmm.. what about something like this - that way we can keep the sidebar header intact and the footer remains consistent with the dashboard? 

image

2) Right! I forgot that we put the Review and Export section on it's own page.

mgwalker:
@jeromeleecms I'm not sure what that screenshot is proposing. Is the footer also fixed? If so, I'd be concerned about how much vertical space that takes away from the actual form.

jeromeleecms:
Just when users scrolls to the bottom ... parking lot discussion today (03/16)?

mgwalker:
Laura with the eagle eyes! I took a look and I'm not 100% sure what's going on with it. It looks like the text has a different baseline, which doesn't make a lot of sense... 🤔 But I bet it'll turn out to be something really simple. We'll see!

Issue #2092, opened 2020-03-06
Explore options for informing users about errors before they submit


lauraponce:
Since we are allowing the app to save even when there are errors present, there is a need for an error check that lets users see any errors and fix them before submitting. The validation check won’t be on autosave or typing or anything annoying, but will be handled by inline validation as outlined in the design system. 

#2054 and the associated Mural shows the validation requirements on the fields within the app.
https://app.mural.co/t/eapd5455/m/eapd5455/1588781600846/74a95a2db5d0b3e72c73cc7a96e6b501f40dcbd1

### This task is done when...
 - [X] Design patterns for this have been reviewed and considered
 - [x] Add assurances and compliance page to the mural for posterity and info on validating the yes/no questions. 
 - [x] Determine if a field can have a generic error, or if it needs something specific customization 
 - [x] Note customization content language on the Mural 

 - [x] Don’t write validation errors for things that are bugs in the queue to be fixed (call these out on the mural by issue #) 
 - [x] Give developers direction around what the design system says about the validation we want to use. Show different areas of the design system where this guidance exists https://design.cms.gov/components/form-label/

 - [x] Mock-up has been designed (if needed)
 - [x] A prototype is created to show interactions with the help drawer and moving to sections of the app to fix errors 
 - [x] Product has provided input
 - [x] A decision has been made about whether to move forward with implementing or not

Validation check is held for a future ticket as part of the update experience - 
 - [ ] Determine and design where folks will hit a button of some kind that triggers the validation check (may be blocked by content nav issues). Should this be before summaries? Before the executive summary? When will they know they have an error and how fixing it will effect the tables they've already seen. Could end up revisiting #2167 through this task?

lauraponce:
From #2106, this issue should also cover the use case for showing an error message and then coming back to the page later. Currently, the error disappears, but it might make sense to keep it visible.

beparticular:
Mural of errors https://app.mural.co/t/eapd5455/m/eapd5455/1588781600846/74a95a2db5d0b3e72c73cc7a96e6b501f40dcbd1

jjames521:
Looking at something like this for the navigation bar.  It includes a number for amount of errors on the page.  The focus style blue bar would turn red to maintain consistency.

![NavErrorVar3.png](https://images.zenhubusercontent.com/5e7151677c78c820a2a11970/b9ae5157-51d5-40b2-a7c4-2487a2320c1b)

beparticular:
[...]

PR #2106, opened 2020-03-17
Always save the APD

pphillips-fearless:
Update the APD validation schema so that dates can be saved as any text instead of being validated as dates. Will we do informational validation as a separate step in the future. Schema validation failures will be considered client-side data failures rather than user errors.

- updated APD schema to allow dates of any string format
- added endpoint test


mgwalker:
I don't think there's anything to document in this change, and the OpenAPI should still be the same. The only thing left is to add a line to the changelog.

jeromeleecms:
Saves as intended! Clarification on the behavior -- (maybe need to spin this into a separate issue): 

1. Entered the invalid data into the date field (check)
2. System calls out invalid data

image

3. The system saved as expected (even with the invalid data)
4. Close out of the APD and come back
5. Revisit the invalid date, error message is gone.

image

Should the app be revalidated form entry on load so it can identify potential error that were saved? I can be OK with approving this as is because it's scoped as saving everything, but I would want to track how to get appropriate error messages to appear when revisiting an APD that contains invalid data.

@pphillips-fearless @mgwalker

mgwalker:
I think my opinion would be to create a new issue and remove that error message altogether. We only put it in because invalid dates would cause the save to fail, and we were trying to highlight for the user what went wrong. Now that's no longer the case and we accept anything. We can put this validation back when we do the larger info-validation work in the future.

@pphillips-fearless @lauraponce Interested what y'all think too.

lauraponce:
I agree with @jeromeleecms that it seems strange to just have the error message go away forever, it seems like this could fit into #2092 for thinking about how to handle errors as a whole. Does that seem right, @mgwalker and @jeromeleecms to call it out in 2092 for figuring out a solution to this and closing this PR?

PR #2120, opened 2020-03-20
Fix css cache busting

### LINKED ISSUE #2118: Fix cache-busting for CSS
Our generated HTML file has two requests to fetch the CSS. The first one has no cache-busting, the second uses the file's SHA hash as a query parameter. We need to get rid of the first one, because having both is causing issues with the cacheing.

### This task is done when...
 - [ ] the generated HTML file only has one request for our CSS file

mgwalker:
We used two different Webpack plugins to inject CSS tags into our HTML. One of these added a content hash query parameter to bust the cache, but the other did not. Unfortunately, because of the ordering, the non-hashed one would load first and override the hashed one. This PR removes the plugin that added the hashed version and instead configures the other to also add a hash, as it probably should have been in the first place.

- fixes CSS cache-busting
- closes #2118 

- n/a ~Tests have been updated (and all tests are passing)~
- n/a ~Changelog is updated as appropriate~

### DIFF
--- a/web/webpack.config.js
+++ b/web/webpack.config.js
@@ -2,7 +2,6 @@ const path = require('path');
 
 const webpack = require('webpack');
 const HtmlWebpackPlugin = require('html-webpack-plugin');
-const HtmlWebpackTagsPlugin = require('html-webpack-tags-plugin');
 const MiniCssExtractPlugin = require('mini-css-extract-plugin');
 
 if (!process.env.IDLE_LOGOUT_TIME_MINUTES) {
@@ -112,11 +111,6 @@ const config = {
       minify: { removeComments: true },
       template: 'src/index.html'
     })
-    // new HtmlWebpackTagsPlugin({
-    //   tags: ['app.css'],
-    //   append: true,
-    //   hash: true
-    // })
   ]
 };
 
--- a/web/webpack.config.js
+++ b/web/webpack.config.js
@@ -19,6 +19,7 @@ const config = {
   },
   output: {
     path: path.join(__dirname, 'dist'),
+    publicPath: '/',
 
     // Bust the cache with a hash!
     filename: '[name].[contenthash].js'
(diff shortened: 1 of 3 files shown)


web/webpack.config.js
```diff
@@ -19,6 +19,7 @@ const config = {
   },
   output: {
     path: path.join(__dirname, 'dist'),
+    publicPath: '/',
```
mgwalker: Adding this causes webpack to prepend `/` to the paths it injects into our HTML. This is important with the paginated view because without it, hitting reload on a page would cause an error. The page would load, but the CSS and JS would not, so you'd just get a white page with no explanation of what was going on. I think this is the fix, but we'll see...

web/webpack.config.js
```diff
@@ -92,10 +93,11 @@ const config = {
       }
     ]
   },
-  // replaces "process.env.____" with the values defined in the actual
-  // environment at build time
   plugins: [
-    new MiniCssExtractPlugin(),
+    new MiniCssExtractPlugin({ filename: '[name].[contenthash].css' }),
```
mgwalker: This plugin now supports hashing itself, which means our CSS files can have hashes in their filenames just like the JS files do.

web/webpack.config.js
```diff
@@ -109,11 +110,6 @@ const config = {
     new HtmlWebpackPlugin({
       minify: { removeComments: true },
       template: 'src/index.html'
-    }),
-    new HtmlWebpackTagsPlugin({
-      tags: ['app.css'],
-      append: true,
-      hash: true
```
mgwalker: Previously we relied on this `HtmlWebpackTagsPlugin` to add a query parameter to the CSS tags (e.g., `<link href="app.css?ca22bf35295a91c36fb0" rel="stylesheet">`), but with the `MiniCssExtractPlugin` adding hashes directly to the filename, we don't need the query parameter anymore, so we can ditch this plugin altogether.

PR #2123, opened 2020-03-20
Remove save button

pphillips-fearless:
Remove the save button

- the component is removed
- other components that exist solely to support the floating save button are removed (the only thing I could find is getSaveApdWorking)
- get rid of the old redux state/selectors that the floating save button used


web/src/reducers/apd.js
```diff
@@ -369,23 +368,6 @@ const reducer = (state = initialState, action) => {
...
-            ...state.byId[action.data.id],
-            updated: getHumanTimestamp(action.data.updated)
-          }
-        },
-        data: {
-          ...state.data,
-          created: getHumanDatestamp(action.data.created),
-          updated: getHumanTimestamp(action.data.updated)
-        }
-      };
```
mgwalker: I don't think we can get rid of this. Without it, the APD "last updated" time won't ever get updated. We use the APD state's `updated` property on the state dashboard (that's how we refer to the current landing page).

web/src/reducers/working.js
```diff
@@ -72,13 +72,11 @@ const reducer = (state = initialState, { type }) => {
...
 const getEditOwnAccountWorking = state => state.working.editOwnAccount;
-const getSaveApdWorking = state => state.working.saveApd;
 
 export default reducer;
 
 export {
   getAddAccountWorking,
   getEditAccountWorking,
-  getEditOwnAccountWorking,
-  getSaveApdWorking
```
mgwalker: Oh, nice catch! I think there's more in this file that we should remove too. There should be some maps that link action symbols to working state properties. I think the `SAVE_APD_*` actions are in those maps, and those can come out. There is a switch below that can be updated too.

web/src/reducers/apd.test.js
```diff
@@ -1049,54 +1048,6 @@ describe('APD reducer', () => {
...
-        },
-        data: {
-          created: 'July 30, 1965',
-          name: 'Timmothert',
-          updated: 'April 11, 1953, 12:00 AM GMT'
-        }
-      },
-      { type: SAVE_APD_SUCCESS, data: { id: 'apdID', updated: '' } }
-    );
-  });
```
mgwalker: Only last thing is we should also put these tests back.

pphillips-fearless: oh, good catch, I forgot that one!

PR #2128, opened 2020-03-23
Implement the message for unexpected save errors

### LINKED ISSUE #2112: Implement unexpected error dialog
That is, the dialog is for an unexpected error, not that the dialog is unexpected. Though random dialogs might be fun? (from #2093)

screenshot

>### There's been an unexpected error.
>We weren't able to save your latest changes. Try saving in a few minutes. If you continue to see this message, refresh your browser. Before this issue is resolved, new changes may be lost if you continue to make edits or if you refresh your browser.

We need to identify all the ways we might get an error in the save process (HTTP 500 returned; timed out; etc.) and make sure we cover them all. E.g., are there any of the HTTP 4xx codes that we need to account for (401 or 403 could happen, but we know what those mean).

### This task is done when...
- [ ] when an unexpected error occurs (HTTP 500 or unknown error), the dialog appears
- [ ] dialog pushes content down rather than sitting on top of it
- [ ] dialog animates down
- [ ] when save is successful, the dialog animates up and then disappears
- [ ] while attempting to save, the save button text changes to "Saving"

mgwalker:

- adds the unexpected error message
  - the dialog is hidden from screen readers until there is an error (it's never removed from the page, just visibly hidden; this extra step makes sure screen reader users aren't told about an error that doesn't exist)
  - new errors are announced to screen readers immediately, but will not be announced again until after the dialog disappears
  - unclear how we should account for non-sighted or low-vision users who get this dialog and don't know where the manual save button is
- fixes a but that would prevent autosave from continuing after a save error
- closes #2112


mgwalker:
Oops, meant to add a note about how to do that! It's not the _easiest_ thing in the world, but there is a way! Once you're on the page, you can have Chrome disable its network connection. Open the inspector (`command + option + i`), click the `Network` tab, and there will be a little dropdown at the top that says `Online`. If you click that, you can change it to `Offline`.

Then when you edit a text box, it'll get an unexpected network error and show you the dialog. You can switch back to `Online` when you're ready to get the save to succeed so you can check that behavior too.

Sorry for forgetting to write that up earlier!

mgwalker:
Oh, that's a good point. It's not as ideal as being able to manually click the save button, but it saves us from making a big mistake by trying to move focus around on the page.

jeromeleecms:
When I put Chrome in offline mode, this behavior happens with TinyMCE:



That gray box is super long too. Went back into online mode than hit save (which was successful), but TinyMCE does not load, even after navigating around on the page.



As a user, I would expect that being in offline mode would still allow the Text Editor to load or at the very least pressing the save button would force a refresh of some sort and then TinyMCE would load so I could see my content.

Thoughts?

mgwalker:
🤔 I think that should definitely be an issue for us to investigate further. It's a function of how TinyMCE loads and I'm not certain we can do anything about it, but it's worth looking into, IMO.

CHANGELOG.md
```diff
@@ -20,6 +22,7 @@ Anticipated release: April 6, 2020
 
 See our [release history](https://github.com/18F/cms-hitech-apd/releases)
 
+[2112]: https://github.com/18F/cms-hitech-apd/issues/2112
```
radavis, line 25: Don't forget to add the octothorpe!

mgwalker: eagle eyes

web/src/containers/UnexpectedError.js
```diff
@@ -0,0 +1,55 @@
...
+const mapStateToProps = state => ({
+  hasError: selectHasError(state),
+  isSaving: selectIsSaving(state)
+});
+
+const mapDispatchToProps = { save: saveApd };
+
+export default connect(mapStateToProps, mapDispatchToProps)(UnexpectedError);
+
+export { UnexpectedError as plain, mapStateToProps, mapDispatchToProps };
```
radavis, line 55: :+1: This component was a great reminder of how Redux works at the component level.

PR #2133, opened 2020-03-25
Add SaveMessage component to Header

### LINKED ISSUE #2104: Update save message
The save message in the header should be updated for each of the scenarios listed below (from [#2094](https://github.com/18F/cms-hitech-apd/issues/2094#issuecomment-598295512))

**Actively saving**: `↻ Saving`

**Less than 1 minute**: `Saved`

**Less than 24 hours, more than 1 minute since last edit**: `Last saved 10:05 a.m. PDT (1 minute ago)`

(one minute later, no new activity): `Last saved 10:05 a.m. PDT (2 minutes ago)`

**More than 24 hours**: `Last saved March 10 (1 day ago)`

**More than 30 days**: `Last saved March 10 (3 months ago)`

**More than a year**: `Last saved March 10, 2019 (1 year ago)`

### This task is done when...
- [x] the save message is updated to cover all of the above cases
- [x] the save message is updated automatically as time passes (so "2 minutes ago" becomes "3 minutes ago" as appropriate) 
- [x] all of the cases above as well as automatic updating are tested

radavis:
Resolves #2104 

- the 'Saved' message in the Header to be a dynamic string, dependent upon the amount of time that has passed since saving.

  - [x] ~Associated OpenAPI documentation has been updated~

### What's missing/incongruent with Issue #2104...

* [x] Displaying 'am' & 'pm' instead of 'a.m.' & 'p.m.' Is this acceptable?
* [x] The three-letter timezone abbreviation is missing. I can add it, but there are [caveats](https://stackoverflow.com/questions/30401787/display-three-letters-time-zone-code-using-jstimezonedetect-script).
* [x] ~Testing auto updating isn't done, yet.~ Done!
* [x] ~Fix Header Component snapshot test.~ Fixed!
* [x] ~CircleCI tasks are stuck. Does it need a kick?~ Opening a PR on this repo, and not a fork, made it work.


web/src/components/SaveMessage.test.js
```diff
@@ -0,0 +1,145 @@
...
+
+    describe("when saved 1 minute ago", () => {
+      it('displays "Last saved 11:59 am (1 minute ago)"', () => {
+        lastSaved = moment(jan1AtNoon).subtract(1, "minutes");
+        subject = shallow(
+          <SaveMessage lastSaved={lastSaved} />
+        );
+        expect(subject.text()).toMatch(/^Last saved/);
+        expect(subject.text()).toMatch(/11:59 am/);
+        expect(subject.text()).toMatch(/\(1 minute ago\)$/);
```
radavis: I might want to condense these expectations to one line. I'm curious which you think is easier to read and easier to update.

```
expect(subject.text()).toMatch("Last saved 11:59 am (1 minute ago)");
```

radavis: After that, it might be possible to format these remaining tests like on line 11

radavis: Example: https://github.com/18F/cms-hitech-apd/compare/d91c94..ce556c

pphillips-fearless: I think I like them separate

mgwalker:
Regarding "am" vs "a.m.", I'd recommend having design or product weigh in. Historically we've been pretty flexible about tweaking designs for implementation simplicity. Long-term maintainability beats short-term pixel-perfect designs. 😄

Pinging @beparticular @jjames521 @lauraponce @jeromeleecms if y'all have feedback.

jeromeleecms:
Citing our much-missed alum's (Nicole Fenton) book, Nicely Said, am / pm is a commonly accepted abbreviation used for the web (example provided without periods)... I am good to hold to that guidance...

radavis:
How do you all feel about the 3-character Time Zone being a "must-have?" The displayed time would always be in the client's browser-detected time zone, from my understanding of how the `momentjs` library works.

radavis:
Yes.

> moment(...) is local mode. Ambiguous input (without offset) is assumed to be local time. Unambiguous input (with offset) is adjusted to local time.

Input into the `moment` library will be adjusted to local time. Calls to output/format time will be in the browser-detected local time.

radavis:
It feels like getting the three-character timezone from the browser is unsupported by the `moment` library. https://github.com/moment/moment/issues/162#

We could investigate this, if we really want to spend the effort. However, it feels like _time_ could be better spent elsewhere...

jeromeleecms:
@radavis Great question - I had originally thought just having a 2 letter time zone appended to the content would be helpful since we don't all work in the same time zone... however... am I understanding that the time zone is automatically localized to client's time zone? 

If that's the case, I would be OK with dropping it off entirely -- unless someone has a compelling reason for us to keep it there.

radavis:
> I understanding that the time zone is automatically localized to client's time zone?

Yes. As long as the moment docs are telling the truth :)

jeromeleecms:
Alright I'm good with dropping it then!

web/src/components/SaveMessage.js
```diff
@@ -0,0 +1,87 @@
...
+  lastSaved: PropTypes.oneOfType([
+    PropTypes.instanceOf(Date),
+    PropTypes.instanceOf(moment),
+    PropTypes.string
+  ].isRequred),
+};
+
+SaveMessage.defaultProps = {
+  lastSaved: moment(),
+};
```
mgwalker: Since `lastSaved` is required, I don't think a `defaultProp` is necessary. Elsewhere, we only define `defaultProps` if the prop is not required. Again deferring to you and @pphillips-fearless's preferences on how you like to write the code, though.

(This is the sort of comment I can imagine leaving lots of over the next few weeks, just pointing out how we've structured things in the past so y'all can make your best decisions about how you want to do things.)

web/src/components/SaveMessage.test.js
```diff
@@ -0,0 +1,88 @@
+import React from "react";
+import { shallow } from "enzyme";
+import moment from "moment";
+import SaveMessage from "./SaveMessage";
+
+describe("<SaveMessage />", () => {
+  let lastSaved;
+  let subject;
```
mgwalker:
```suggestion
```

I'd move these closer to the point-of-use, at which point they can be `const`s. For me, at least, it's clearer that the value is new and I'm not worrying about overwriting some previously existing variable.

web/src/components/SaveMessage.test.js
```diff
@@ -0,0 +1,88 @@
...
+  describe('when saved less than 1 minute ago, it displays "Saved"', () => {
+    [
+      ["1 second ago", 1],
+      ["2 seconds ago", 2],
+      ["30 seconds ago", 30],
+      ["59 seconds", 59],
+    ].forEach(([testName, seconds]) => {
+      test(testName, () => {
+        lastSaved = moment().subtract(seconds, "seconds");
+        subject = shallow(
```
mgwalker:
```suggestion
        const lastSaved = moment().subtract(seconds, "seconds");
        const subject = shallow(
```

This is a stylistic choice, so again something for you and @pphillips-fearless to decide how you want to do it. I'm just imagining some future version of you or another dev looking at this and wondering "where did `lastSaved` come from?" and then having to search for it. 🤷‍♂

web/src/components/SaveMessage.test.js
```diff
@@ -0,0 +1,88 @@
...
+    const now = new Date(2020, 0, 1, 12, 0);
+    const oneMinuteFromNow = new Date(2020, 0, 1, 12, 1);
+    let mockDateNow;
+
+    beforeEach(() => {
+      jest.useFakeTimers();
+      mockDateNow = jest
+        .spyOn(Date, "now")
+        .mockReturnValueOnce(now)
+        .mockReturnValue(oneMinuteFromNow);
```
mgwalker: This makes me a little bit nervous because it relies on the internal behavior of `moment`.  Ideally `jest.useFakeTimers()` would let us initialize it with a timestamp and it would also fake out `Date` for us ([sinon's fake timers](https://sinonjs.org/releases/latest/fake-timers/) work that way), and then we could just fast-forward the timer by 60 seconds. Alas, Jest isn't there...

(Which is all to say, I don't love this but I also think it's probably the solution.)

web/src/components/SaveMessage.test.js
```diff
@@ -0,0 +1,88 @@
...
+      [1, "minute", "Last saved 11:59 am (1 minute ago)"],
+      [60 * 24 - 1, "minutes", "Last saved 12:01 pm (1 day ago)"],
+      [3, "hours", "Last saved 9:00 am (3 hours ago)"],
+      [1, "day", "Last saved December 31 (1 day ago)"],
+      [30, "days", "Last saved December 2 (1 month ago)"],
+      [364, "days", "Last saved January 2 (1 year ago)"],
+      [3, "years", "Last saved January 1, 2017 (3 years ago)"],
+    ].forEach(([value, timeUnit, result]) => {
+      const testName = `when saved ${value} ${timeUnit} ago, it displays "${result}"`
+      test(testName, () => {
```
mgwalker:
```suggestion
      test(`when saved ${value} ${timeUnit} ago, it displays "${result}"`, () => {
```

Another style choice. My preference is not to create variables that are only used once on the immediate next line *unless* creating them is a long or complex sequence. In the case of strings, I think making them is easy enough to read (especially with template strings!) not to ever bother. My thinking here is that when you create a variable, you're implying something about the lifetime of that value, so my brain stashes it away so I know what it is when I see it later. But then I never see it later. 😛

radavis:
Just noticing that I can't get to the deployed site via the cms-eapd-bot's link. I'll investigate this tomorrow in the AM.

jjames521:
@jeromeleecms think you had referenced the design system being somewhat built off the ideology of Material Design in some ways.   In that case am / pm would fit with the system pretty nicely, and as Nicole had said in _Nicely Said_. it's commonly accepted anyways.

web/src/components/SaveMessage.js
```diff
@@ -0,0 +1,87 @@
...
+    d: "1 day",
+    dd: "%d days",
+    M: "1 month",
+    MM: "%d months",
+    y: "1 year",
+    yy: "%d years",
+  },
+});
+
+class SaveMessage extends React.Component {
```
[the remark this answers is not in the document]
radavis: Ah, yes. I can definitely change this to a functional component since that will maintain consistency with other components.

web/src/components/SaveMessage.js
```diff
@@ -0,0 +1,87 @@
...
+    let result = "Last saved";
+
+    if (duration.asMinutes() < 1) return "Saved";
+
+    // eslint's "yoda": "Expected literal to be on the right side of <="
+    // Which is easier to visualize on a number line, Mr. Yoda?
+    // lowerBound <= object.value() && object.value() < upperBound  // or...
+    if (duration.asMinutes() >= 1 && duration.asDays() < 1) {
+      result = `${result} ${lastSavedMoment.format("h:mm a")}`;
+    }
```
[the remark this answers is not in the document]
radavis: `lowerBound <= object.value() && object.value() < upperBound` could read: is the value between the lower and upper bound

web/src/components/SaveMessage.js
```diff
@@ -0,0 +1,87 @@
...
+    return result;
+  }
+}
+
+SaveMessage.propTypes = {
+  lastSaved: PropTypes.oneOfType([
+    PropTypes.instanceOf(Date),
+    PropTypes.instanceOf(moment),
+    PropTypes.string
+  ].isRequred),
```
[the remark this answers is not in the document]
radavis: Cool, cool. I was just thinking of the possible types you could throw at this component. I'm not necessarily testing all of those types, but moment should be able to handle them...

web/src/components/SaveMessage.js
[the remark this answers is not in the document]
pphillips-fearless: I'm just reading up on hooks, not having used them before, but so far I like what I'm reading

web/src/components/SaveMessage.js
radavis: I think the linter didn't like it. I can take it out and verify.

web/src/components/SaveMessage.test.js
pphillips-fearless: I like using const whenever possible

web/src/components/SaveMessage.test.js
radavis: This is necessary because the component under test is using moment (which is using Date) under the hood. I can attempt to refactor this and mock moment's functionality, it just might take me a bit longer to figure out.... I agree, it doesn't look pretty.

web/src/components/SaveMessage.test.js
radavis: Can do. I always wrestle with the idea of "Am I doing too much on a single line?," and tend to break things up.

web/src/components/SaveMessage.test.js
radavis: As the file grows, it could be difficult to find these variable declarations. Your argument makes sense. 👍

web/src/components/SaveMessage.test.js
mgwalker: No no, please don't! It's good as-is. It's a shortcoming of the jest time mocks, not your code!

web/src/components/SaveMessage.test.js
```diff
@@ -0,0 +1,89 @@
...
+      ["30 seconds ago", 30],
+      ["59 seconds", 59],
+    ].forEach(([testName, seconds]) => {
+      test(testName, () => {
+        const lastSaved = moment().subtract(seconds, "seconds");
+        // https://reactjs.org/docs/test-renderer.html#testrendereract
+        act(() => {
+          subject = create(<SaveMessage lastSaved={lastSaved} />);
+        })
+        expect(subject.toJSON()).toEqual("Saved");
```
mgwalker: I'm going to open a tech debt issue to come back and think about these tests at some point. These particular ones are a decent candidate for snapshots, for example.

Certainly not something worth changing *now* and maybe not worth changing at all! But let's put a pin in it and figure it out later. 😄

web/src/components/SaveMessage.test.js
```diff
@@ -0,0 +1,89 @@
...
+      jest.clearAllTimers();
+    });
+
+    it('auto-updates from "Saved" to (1 minute ago)', () => {
+      act(() => {
+        subject = create(<SaveMessage lastSaved={now} />);
+      })
+      expect(subject.toJSON()).toMatch("Saved");
+      act(() => jest.advanceTimersByTime(60 * 1000));
+      expect(subject.toJSON()).toMatch(/\(1 minute ago\)$/);
```
mgwalker: Totally not a hack. I think this is how React recommends doing tests if you've got hooks, really. It's possible we should being transitioning entirely to `react-test-renderer`. Otherwise, there are ways to make it work with Enzyme.

Again, something to put a pin in and revisit later. Definitely not something we have to figure out for this PR.

radavis:
Screen Shot 2020-03-27 at 2 50 46 PM

@jeromeleecms @lauraponce You can view the new save message functionality, here: https://ec2-3-87-56-55.compute-1.amazonaws.com/apd

mgwalker:
@lauraponce Good catch! I noticed that too and remembered an issue from a while back where the APD immediately saves when it's loaded for the very first time. (The backend creates the APD and the frontend adds some initial data for the federal citations. Then the frontend tries to save it before anything gets lost.) It *should* only happen the first time an APD loads, so we need to double-check that. If it happens every time, that's probably a bug!

radavis:
Regarding accessibility of this feature, from Heather Battaglia via Slack...

> We'd had some discussions about autosave accessibility, and decided that announcing autosave every time it happened was too aggressive. We're announcing to screen readers when they land on the page that the page will be saved automatically, and then we're announcing errors when they happen, IIRC.

PR #2148, opened 2020-03-31
Remove default costs from new activities and APDs

### LINKED ISSUE #2129: Remove default set of costs for new activities
Currently a new activity comes with one state personnel, one non-personnel cost, and one contractor costs. Investigate whether we can remove those defaults.

### This task is done when...
 - [ ] we know what we want to do
 - [ ] we've create a new issue/updated this one to do it

mgwalker:
When creating a new activity, we will no longer create default state personnel, non-personnel, or contractor resource expenses with it. Same story with new APDs.

Also fixes a bug introduced in #2080 where APDs could not save after a new activity was created because the API required all activities to have a program type/funding source but new activities did not. A related bug (to be fixed in #2128) prevented any furthers saves after an initial error, so the practical upshot was that it became impossible to save an APD if you added an activity to it. Fixed in this PR because a) it was hampering my testing and b) it was super duper easy.

- new APDs and activities do not have default costs
- adding an activity doesn't break the APD for all eternity
- closes #2129

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,6 +8,9 @@ Anticipated release: April 6, 2020
 
 #### 🐛 Bugs fixed
 
+- Fixed an issue where new APDs and activities came prepopulated with cost information ([#2129]) which could cause incorrect behavior in the activity cost summary table ([#2130])
+- Fixed an issue where adding an activity could make it impossible to save the APD ([#2148])
+
 #### ⚙️ Behind the scenes
 
 - Updated 3rd-party libraries to the latest versions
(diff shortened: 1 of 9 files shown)


mgwalker:
I noticed that too! What a throwback. Those are super old, and it's been a long time since we've seen them.

I also noticed that you can't delete the first expense in a list. That's also old behavior, and I was wondering if we should revisit that as well. I'll create a couple of "let's look at this" issues so we don't forget and we can decide later how/if we want to deal with both of those.

mgwalker:
#2151, #2152

PR #2158, opened 2020-04-06
Move JWT from cookie store to Authorization header 

### LINKED ISSUE #2100: Switch authentication from cookies to another mechanism
3rd-party cookies are becoming increasingly difficult to work with, and Chrome seems intent on killing them entirely if at all possible. Rather than keep working around these issues with cookies, we should switch to a different mechanism. It's a bit more work, but it'll pay off in not having to fight these cookie fires over and over.

### This task is done when...
- [ ] we've picked a different approach
  - my initial thought is to take the token we're currently putting in the cookie and instead send it back and have subsequent requests put it in the `Authorization` header, similar to a bearer token
- [ ] implemented
- [ ] documentation is updated

radavis:
Resolves #2100 

### This pull request...

- removes JWT from cookies
- stores JWT on front end in localStorage
- sends JWT in Authorization header with each request

- [x] ~The experience passes a basic manual accessibility audit (keyboard nav, screenreader, text scaling) OR an exemption is documented~ no UI changes


api/auth/session.js
```diff
@@ -90,14 +74,7 @@ module.exports = ({ Cookies = defaultCookies } = {}) => {
...
           COOKIE_NAME,
-          jwt.sign(
-            { iss: 'CMS eAPD API', payload: session.content },
-            process.env.SESSION_SECRET,
-            {
-              algorithm: 'HS256',
-              expiresIn: `${process.env.SESSION_LIFETIME_MINUTES}m`
-            }
-          ),
+          signWebToken({ payload: session.content }),
```
radavis, line 77: Extracted common functionality in this file into functions (`signWebToken`, `verifyWebToken`)  within `api/auth/jwtUtils.js`

docker-compose.yml
```diff
@@ -42,6 +42,7 @@ services:
       - SESSION_SECRET=dadccbf1e3bcb63a4d533c3f184eecfc79fd07e76d0584f808078125a83bc40b6413829e5c690d391dd3f6cbca44406cef07b8193cd4db2f03bab355eea9cb2e
       - DISABLE_SAME_SITE=yes
       - DISABLE_SECURE_COOKIE=yes
+      - LOG_LEVEL=silly
```
radavis, line 45: Remember to change this back!

api/auth/index.js
```diff
@@ -69,7 +60,32 @@ module.exports.setup = function setup(
...
+    const jwt = req.get('Authorization').replace('Bearer ', '');
+    const payload = verifyWebToken(jwt);
+
+    if (payload) {
+      removeSession(payload.sub);
+      req.logout();
+      res.status(200).end();
+    } else {
+      res.status(400).end();
+    }
```
radavis, line 87: I could use some help updating the tests for this file. I'm so very lost...

radavis:
'when unauthenticated' tests are not returning a response. I am unsure why this is happening.

Screen Shot 2020-04-27 at 11 16 27 AM

radavis:
Sorry this is such a massive change! 😬

api/routes/auth/login.endpoint.js
```diff
@@ -1,53 +1,53 @@
...
+  xit('proper response headers', async () => {
+    // I'm unable to pick up these headers w/ axios
+    const response = await api.post(url).then(res => res);
+    // console.log(response.headers);  // -> { 'cache-control': 'private, no-cache' }
+    // https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Vary
+    expect(response.headers.vary).toEqual('Origin');
+    // https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Access-Control-Allow-Credentials
+    expect(response.headers['access-control-allow-credentials']).toEqual(
+      'true'
+    );
```
radavis, line 50: Hey @mgwalker, I switched our endpoint tests from [`node request`](https://github.com/request/request) to axios since the request library was deprecated in February. I'm unable to see these headers. 

Are they important? I can't find in the codebase where they should be set by the server. Halp! :)

mgwalker: They were, but I don't think they are anymore. I'll have to dig into it again but I think this header was necessary to allow Javascript to send authenticated requests back to the API later with cookies. Since the requests are authenticated with a token in a header ***and*** that token is sent back in a response body (instead of being sent in a cookie or header), I think we don't need this header anymore.

However, we also need to remove the `withCredentials: true` configuration from Axios on the frontend. If the `withCredentials` property of the request is true, then that header must also be set; otherwise, the browser won't deliver the results into the JS code.

docker-compose.yml
```diff
@@ -40,7 +40,6 @@ services:
       - DATABASE_URL=postgres://postgres:cms@db/hitech_apd
       - NODE_ENV=development
       - DISABLE_SAME_SITE=yes
-      - DISABLE_SECURE_COOKIE=yes
```
radavis, line 43: I'm assuming that I can remove this env var. Please correct me if I'm wrong!

api/routes/apds/files.js
```diff
@@ -41,10 +42,14 @@ module.exports = (
 
   logger.silly('setting up POST /apds/:id/files route');
 
+  // "Never add multer as a global middleware since a malicious user could
+  // upload files to a route that you didn't anticipate."
+  // -- https://github.com/expressjs/multer#any
```
mgwalker, line 47: Nice catch and fix!

api/routes/auth/logout.endpoint.js
```diff
@@ -1,43 +1,22 @@
...
-      response.headers['set-cookie']
-        .filter(cookie => cookie.startsWith('token='))
-        .every(
-          cookie =>
-            cookie ===
-            'token=; path=/; expires=Thu, 01 Jan 1970 00:00:00 GMT; samesite=none; httponly'
-        )
-    ).toBeTruthy();
+      expect(response.status).toEqual(400);
+    });
```
mgwalker, line 11: ### discussion / maybe change

This seems like a change in behavior. Would it be better for the API to always return a `200` (or probably most appropriately a `204` but 🤷‍♂️) for `/auth/logout`? I'm trying to think of the benefit of signaling to the caller that they weren't already logged in, versus the burden.

E.g., if a user has left the web app open for a while, it's possible that their auth token could expire. Then if they click the logout button in the app, what happens in the app if that results in a non-success status? (At the moment, the app wouldn't do anything because it only proceeds to the logout/re-log-in screen on a success status).

api/routes/openAPI/auth.js
```diff
@@ -135,18 +139,13 @@ module.exports = {
...
-              schema: {
-                type: 'string',
-                example: 'session=; expires=; httponly'
-              }
-            }
-          }
+          description: 'Deletes the session'
+        },
+        400: {
+          description: 'Not logged in'
```
mgwalker, line 148: Same thing with the error on logging out

api/routes/openAPI/helpers.js
```diff
@@ -1,10 +1,10 @@
 module.exports.responses = {
   unauthed: {
     401: {
-      description: 'Does not have permission to this activity'
+      description: 'Not logged in'
     },
     403: {
-      description: 'Not logged in'
+      description: 'Does not have permission to this activity'
```
mgwalker, line 7: 😂 Thank you!

api/routes/openAPI/index.js
```diff
@@ -557,10 +557,10 @@ module.exports = {
...
     },
     securitySchemes: {
-      sessionCookie: {
-        type: 'apiKey',
-        in: 'cookie',
-        name: 'token'
+      bearerAuth: {
+        type: 'http',
+        scheme: 'bearer',
+        bearerFormat: 'Bearer xxx.yyy.zzz'
```
mgwalker, line 563: Nice 👍

docker-compose.yml
mgwalker: I think that's right

web/src/actions/auth.test.js
```diff
@@ -29,27 +29,34 @@ describe('auth actions', () => {
...
-    it('creates LOGIN_SUCCESS after successful auth', () => {
+    it('creates LOGIN_SUCCESS after successful auth', async () => {
       const store = mockStore({});
       fetchMock
         .onPost('/auth/login/nonce', { username: 'name' })
         .reply(200, { nonce: '123abc' });
       fetchMock
         .onPost('/auth/login', { username: '123abc', password: 'secret' })
-        .reply(200, { moop: 'moop', activities: [] });
+        .reply(200, { user: { moop: 'moop', activities: [] }, token: 'xxx.yyy.zzz' });
```
mgwalker, line 50: Good ol' `moop moop`

api/auth/index.js
```diff
@@ -64,12 +45,50 @@ module.exports.setup = function setup(
...
 
   // Add a local authentication endpoint
   logger.silly('setting up a local login handler');
-  app.post('/auth/login', passport.authenticate('local'), (req, res) => {
-    res.send(req.user);
+  app.post(
+    '/auth/login',
+    passport.authenticate('local', { session: false }),
+    (req, res) => {
+      serializeUser(req.user, (err, sessionId) => {
```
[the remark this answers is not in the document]
radavis: Thanks for documenting that!

api/auth/index.js
```diff
@@ -64,12 +45,50 @@ module.exports.setup = function setup(
...
+        }
+      });
+    }
+  );
+
+  // Pull JWT from HTTP headers and deserialize user for all routes, except for
+  // login routes.
+  const jwtExcludedRoutes = ['/auth/login/nonce', '/auth/login'];
+  app.use((req, res, next) =>
+    jwtExcludedRoutes.includes(req.originalUrl)
```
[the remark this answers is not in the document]
radavis: So, I was thinking about this yesterday evening. If I rely on the `jwtMiddleware` to only deserialize the user if the Authorization header is present (e.g. remove the `res.send(400).end()`), and rely on `can` and `loggedIn` to manage sending `401` and `403` responses, I can get rid of this funky excluded routes thing.

api/auth/index.test.js
```diff
@@ -48,157 +43,64 @@ tap.test('authentication setup', async authTest => {
...
-
-    setupTest.ok(app.put.notCalled, 'no PUT endpoints are added to the app');
-    setupTest.ok(app.del.notCalled, 'no DELETE endpoints are added to the app');
-  });
-
-  authTest.test('setup works with defaults, too', async setupTest => {
-    authSetup(app);
+  authTest.test(
+    'setup calls everything we expect it to',
+    { skip: true },
```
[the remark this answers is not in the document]
radavis: Whoops! I was attempting to isolate some tests. Nice catch.

api/routes/apds/files.js
radavis: Thanks!

radavis:
@mgwalker I'm pretty sure I've address all your requested changes. Please, take a look...

I have a few questions for you:

* What do you think about the change to the jwtMiddleware?
* Do you think there is any issue with the way I am handling logging out? Should logging out of one session end all of that user's sessions?
* At what point in this process should I push my changes to the Wiki?

api/auth/jwtUtils.js
```diff
@@ -9,57 +9,60 @@ const HS256 = 'HS256';
...
  * @returns {(String|null)} JWT string or null
  */
 const jwtExtractor = req => {
-  let token = req.get('Authorization');
-  if (!token) return null;
-  token = token.replace('Bearer ', '');
-  return token;
+  const token = req.get('Authorization');
+  if (!token || !token.toLowerCase().match(/^bearer\s.+\..+\..+/)) return null;
+  const [temp, result] = token.split(' ');  // eslint-disable-line no-unused-vars
```
mgwalker, line 58: As a future note, one convention for throwaway destructured values would be something like:

```js
const [_, result] = ...
```

And then the linter won't complain. Only trouble is you can't do that if you have more than one throwaway variable. 😛

api/auth/jwtUtils.test.js
```diff
@@ -67,15 +77,22 @@ tap.test('jwtUtils', async t => {
...
-      t.equal(result, 'xxx.yyy.zzz', 'returns the JWT');
-    });
+    const scenarios = [
+      ['Bearer xxx.yyy.zzz', 'xxx.yyy.zzz', 'returns the JWT'],
+      ['bearer xxx.yyy.zzz', 'xxx.yyy.zzz', 'returns the JWT'],
+      ['bearer', null, 'returns null'],
+      ['bearer ', null, 'returns null'],
+      ['', null, 'returns null'],
+      ['Elephanter xxx.yyy.zzz', null, 'returns null']
+    ];
```
mgwalker, line 87: Nice!

radavis:
Notes to Design and Product: This is a back-end change, so you shouldn't see any changes in UI. Verify that this is functional by signing in and out. That's it!

PR #2163, opened 2020-04-07
Add download link to Edge

### LINKED ISSUE #2145: Add Edge (Chromium) to the list of browser downloads on the "red browser" screen
Currently on the "red browser" warning screen, we only provide downloads to Chrome and Firefox. We should add Edge now that the Chromium builds are available.

### This task is done when...
 - [ ] the "red browser" warning screen includes a link to Edge (Chromium)

mgwalker:

- adds a download link to Edge
- updates the Firefox logo
- updates the minimum versions of supported browsers per our earlier rules (Chrome to 80, Edge to 79)
- preps the changelog for the new sprint
- only shows the upgrade browser dialog on login and state dashboard pages
- closes #2145


pphillips-fearless:
I wanted to see the browser warning in action, so I modified the Root.js file to remove the browserIsYellow condition so the UpgradeBrowser control would show.  It looks like the blue header bar blocks a good chunck of it.  Is there a better way to see it (other than loading an old browser)?  Or should that be fixed?

mgwalker:
Oh, I see what you mean. I think that's a bug, yeah. Should we open a new issue for it, or do you think it'd be better for me to go ahead and try to fix it in this one?

pphillips-fearless:
My general rule of thumb is that if it's a quick fix it's good to get it while you've got the hood open.  Otherwise a new task is appropriate.  Unless there's a reason to track it for the sake of documentation, which may be another reason to make it a separate task.

mgwalker:
Makes sense to me! Let me see what I can do. 😄

mgwalker:
Now that I dig into it a bit, the yellow browser upgrade banner causes a lot of problems:

screenshot of the browser upgrade banner sitting behind the sidebar and causing the save error dialog to be visible

On the APD page, it falls behind the sidebar, and it pushes the page content down far enough that the save error dialog becomes visible (it's only visibly hidden by pushing it up off the screen – removing it from the DOM messes up the animation; it's hidden from screen readers with `aria-hidden`).

I'm wondering if we should a) provide a way to dismiss the browser upgrade alert and b) only show it on the login/dashboard pages. That is, if someone ignores it long enough to get to the APD, maybe we shouldn't continue taking up so much of their vertical space. Maybe a question for @beparticular and @jeromeleecms? (cc: @lauraponce for visibility)

beparticular:
@mgwalker I think option b) showing it only on log in and dashboard is fine as long as everything will still function in the eAPD, even if the experience is slightly degraded.

mgwalker:
- now only shows on the login and state dashboard pages
- now sits underneath the blue header in both of those

pphillips-fearless:
I see that you use dangerouslySetInnerHTML to set the HTML content in UpgradeBrowser.js.  Wasn't there a way to create it as a component or as JSX?

mgwalker:
The historical reason is that we separately export the raw HTML so that it can be used in the "red browser" banner (see `web/src/app.js`) and we'd only have that content in one place instead of having it duplicated. We opted to export the raw HTML because we couldn't be sure that "red" browsers would be able to render React at all, so we couldn't use a component for that.

For "yellow" browsers, though, we are definitely using React, so it _has_ to be a component. And since we already had the HTML for the "red" browsers, we used the `dangerouslySetInnerHTML` to reuse it.

I would be very happy to see that changed somehow, though! We just didn't want that upgrade banner content to exist in multiple places because it'd be more likely to get out of sync then.

PR #2189, opened 2020-04-22
Left side navigation should not get stuck under the header

### LINKED ISSUE #2187: Left side navigation should not get stuck under the header
The left side navigation should not scroll up under the header

### This task is done when...
 - [ ] the left side navigation does get stuck under the header
 - [ ] design has reviewed/approved the changes

Export and Submit page - 

This also happens on the Activity Schedule Summary when there is no information in the table.

pphillips-fearless:
added min-height to css for main content div because short pages cause menu to push up under the header

- css update for main content div
- closes #2187 

### DIFF
--- a/web/src/styles/scss/layout.scss
+++ b/web/src/styles/scss/layout.scss
@@ -23,7 +23,7 @@
   float: left;
   + .site-main {
     margin-left: 244px; // sidebar width plus $spacer-4
-    min-height: 800px; // min height of main content to prevent menu from pushing up under the header
+    min-height: 1100px; // min height of main content to prevent menu from pushing up under the header
   }
 }
 
--- a/web/src/styles/scss/layout.scss
+++ b/web/src/styles/scss/layout.scss
@@ -23,7 +23,14 @@
   float: left;
   + .site-main {
     margin-left: 244px; // sidebar width plus $spacer-4
-    min-height: 1100px; // min height of main content to prevent menu from pushing up under the header
+    /*
+    The state icon above the menu will push up under the header if the right side of the page is too short.
+    So we need a min-height for site-main to be at least the height of the menu (plus a little). 
+    Note: "Private Contractor costs" is a particularly short page where the effect can be seen.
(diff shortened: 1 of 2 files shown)


pphillips-fearless:
It looks like you're on a PC, I've been using a mac.  I will get my PC out and spin it up

pphillips-fearless:
hmm, I just did some testing in FF and chrome on mac and PC and couldn't reproduce your screenshot.  The only way I could get the little triangle on the scrollbar to go up under the header was to scroll all the way down to the bottom and look at the footer.  At that point the whole page scrolls up, including the nav menu, so I would expect the little triangle to go up under the head, along with everything else.  Maybe in the parking lot time tomorrow you can show my how you got it to happen.

pphillips-fearless:
@jeromeleecms after your screenshare today I was able to reproduce and fix the issue you documented.  Please re-check it when you have time.  Thanks for your help!

jeromeleecms:
Thanks @pphillips-fearless it appears to behaving as expected!

web/src/styles/scss/layout.scss
```diff
@@ -23,6 +23,7 @@
   float: left;
   + .site-main {
     margin-left: 244px; // sidebar width plus $spacer-4
+    min-height: 1100px; // min height of main content to prevent menu from pushing up under the header
```
mgwalker, line 26: I'm curious where this `1100px` comes from, and if maybe it would be better as a `calc`ed value? Something similar to the sidebar's max height, for instance: `min-height: calc(100vh - 60px);`

pphillips-fearless: I added the following comment to the CSS file to explain my reasoning:

The state icon above the menu will push up under the header if the right side of the page is too short.  So we need a min-height for site-main to be at least the height of the menu (plus a little).  Note: "Private Contractor costs" is a particularly short page where the effect can be seen.  Since there isn't an easy way to do relative heights of "unrelated" elements in CSS, just use a constant.  Remember to adjust it when adding new menu items that change the menu size.

PR #2202, opened 2020-04-30
Adjust Key Personnel Calculations

### LINKED ISSUE #2174: Adjust Key Personnel Calculations
In the math, we have always treated the key personnel cost as the total cost attributable to the program, not the key person's total cost to the state. (The yes/no radio buttons kind of imply that's what we're asking for, too.) So if someone enters $50,000 and 50% commitment, we use $50,000 for their cost in all the math. We don't scale it based on their percent commitment.

We should calculate key state staff the same as how we treat state staff in activities.

Also update the help text to be clear how it will be treated.

Refer to original task #2111 

### This task is done when...
 - [ ] the calculation is updated
 - [ ] the help text has been updated

pphillips-fearless:
Calculate key state staff costs the same as how we treat state staff in activities, as total cost = FTE (fraction) times salary

- personnel costs in the "Key Personnel and Program Management" section now use (FTE) x (cost) = (total) format in 1) review mode, 2) edit mode, and 3) print preview mode
- the tables in the "FFP and Budget" section of the Program Administration activity have been updated to use (FTE) x (cost) = (total) format
- closes #2174 


jeromeleecms:
Some thoughts:

1) There seems to be a bug with the key personnel. If "No" is selected, any data entry listed in the "Yes" still shows up. In the preview site:

image

The summary table shows this:

image

2) Instead of the label "Number of FTEs" -- what do you think about the label of "FTE Allocation (1 = 1 FTE or 100% time):" (open to other suggestions)

image

3) Let's update the collapsed view to mirror the one in the personnel activities (screenshot below):

image

Something like this:
2020 Costs: $75,000 | FTE: 0.75 | Total: $75,000
2020 Costs: $100,000 | FTE: 1 | Total: $100,000

Since we don't have the total cost in the individual activity - would be interested in seeing if @beparticular had any thoughts on (2) or (3).

web/src/reducers/budget.js
```diff
@@ -150,9 +150,12 @@ const buildBudget = incomingBigState => {
...
         ...keyPersonnel
           .filter(kp => kp.hasCosts)
-          .map(({ costs }) => ({
+          .map(({ costs, fte }) => ({
             years: Object.keys(costs).reduce(
-              (o, ffy) => ({ ...o, [ffy]: { amt: costs[ffy], perc: 1 } }),
+              (o, ffy) => ({
+                ...o,
+                [ffy]: { amt: costs[ffy], perc: fte[ffy] }
+              }),
```
mgwalker, line 158: I expected the math part of this to be more complicated, but it looks like also changing this to a by-year FTE property, it's actually a pretty straightforward change. Nice!

web/src/reducers/budget.test.js
```diff
@@ -363,10 +363,12 @@ describe('budget reducer', () => {
           keyPersonnel: [
             {
               costs: { '1931': 150, '1932': 151, '1933': 152 },
+              fte: { '1931': 1, '1932': 1, '1933': 1 },
               hasCosts: false
             },
             {
               costs: { '1931': 150, '1932': 1500, '1933': 15000 },
+              fte: { '1931': 1, '1932': 1, '1933': 1 },
```
mgwalker, line 371: I know this is still a WIP but just wanted to go ahead and plant this seed: at least one of these FTE amounts should be something other than 1. Maybe one should be zero and another should be between 0 and 1. That will verify that key personnel costs are being adjusted by their FTE percentages.

Then Jerome or I can hand-calculate whether the number are right to verify the test result. 😅

pphillips-fearless:
@jeromeleecms I made the label changes you requested.  I wanted to remind you to please check the print-preview view and comment on those labels too.

web/src/reducers/budget.test.js
```diff
@@ -363,10 +363,12 @@ describe('budget reducer', () => {
           keyPersonnel: [
             {
               costs: { '1931': 150, '1932': 151, '1933': 152 },
+              fte: { '1931': 0, '1932': 0.5, '1933': 1 },
               hasCosts: false
             },
             {
               costs: { '1931': 150, '1932': 1500, '1933': 15000 },
+              fte: { '1931': 0, '1932': 0.3, '1933': 1 },
```
mgwalker, line 371: 👍

web/src/reducers/activities.selectors.test.js
```diff
@@ -142,7 +142,7 @@ describe('activities state selectors', () => {
               costs: { '1990': 100, '1991': 200 },
               hasCosts: true,
               name: 'key person',
-              percentTime: '100'
+              fte: { '1990': 1.0, '1991': 1.0 }
```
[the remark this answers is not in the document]
pphillips-fearless: done

pphillips-fearless:
Jenn identified a bug where you can't type a decimal point into the FTE field.  I verified that the onChange event strips out a decimal point as soon as you type it.  Looking into it now

jeromeleecms:
Instead of "FTEs," can we simply make it "FTE" since it is one person? It appears in both the print preview and the builder page. 

image

With the upcoming bug fix for the decimal, this should be fine, but will recheck it once the new preview site is deployed.

beparticular:
@CSwartzHMA see Jerome's comment about FTE vs FTEs. Is this something we can add to the content consistency review if it's not there already?

CSwartzHMA:
Yes. Good catch. I will add it to my document now.

PR #2240, opened 2020-05-26
[Tech Debt] Refactor API user serialization methods

### LINKED ISSUE #2206: [Tech Debt] Refactor API user serialization methods
The API's user serialization and deserialization methods were plugins for the Passport session library prior to #2158. After #2158, the serialization methods are only used by our own code, so they doesn't have to conform to Passport's APIs anymore. It might make sense to refactor them into promise-based functions to match with the rest of our custom code.

### This task is done when...
 - [ ] the `serializeUser` and `deserializeUser` methods in `api/auth/serialization.js` return a promise instead of requiring a callback

pphillips-fearless:
Refactor API user serialization methods

- The API's user serialization and deserialization methods were refactored into promise-based functions
- closes #2206 

~~- [ ] The experience passes a basic manual accessibility audit (keyboard nav, screenreader, text scaling) OR an exemption is documented~~


api/auth/serialization.test.js
```diff
@@ -21,16 +19,14 @@ tap.test('passport serialization', async serializationTest => {
...
-    await serialization.serializeUser(user, doneCallback, { sessionStore });
-
-    serializeTest.ok(
-      doneCallback.calledWith(null, 'session-id'),
-      'serializes the user object'
-    );
+    serialization.serializeUser(user, { sessionStore }).then(sessionId => {
+      serializeTest.same(sessionId, 'session-id', 'serializes the user object');
+      serializeTest.end();
+    });
```
mgwalker, line 29: I'm curious about why this was changed from an `await`-type promise to using a `then`, just as a matter of consistency.

api/auth/index.test.js
```diff
@@ -143,7 +143,7 @@ tap.test('authentication setup', async authTest => {
       im: 'weasel',
       ir: 'baboon'
     };
-    serializeUser.yields(null, 'unique-session-id');
+    serializeUser.resolves('unique-session-id');
```
mgwalker, line 146: Curious if you think we should add a tech debt issue to add a test for the case where `serializeUser` throws an error?

api/auth/serialization.test.js
pphillips-fearless: I was on the fence about this but I did it first and was favoring the `then` type.  The second one needed the `await` syntax so you're right we should make them consistent.

api/auth/serialization.js
```diff
@@ -9,45 +9,38 @@ const { addUserSession, getUserIDFromSession } = require('./sessionStore');
...
-        done(null, null);
       }
     } else {
       logger.warn(`${sessionID} does not map to a user ID`);
-      done(null, null);
     }
   } catch (e) {
     logger.error(null, e);
-    done('Could not deserialize user');
+    logger.silly('Could not deserialize user');
```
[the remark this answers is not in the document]
pphillips-fearless: we can do a catch-log-throw here

PR #2274, opened 2020-06-03
Hide Create New APD button for users who don't have create permission

### LINKED ISSUE #2198: Hide Create New APD button for users who don't have create permission
Hide Create New APD button for users who don't have permission to create a new APD

### This task is done when...
 - [ ] "Create new" button is hidden for users who cannot perform this action

reference design task #2146 for context

pphillips-fearless:
The approach we decided on to avoid the condition is to hide the roles which don't have create permission.  Rather than removing the roles from the database and having to remember what they were when we want to add them back, we added an active column to the roles table.

- an isActive column was added to the roles table
- only active roles (admin, state coordinator) are displayed when adding or updating a user
- closes #2198 

~~- [ ] The experience passes a basic manual accessibility audit (keyboard nav, screenreader, text scaling) OR an exemption is documented~~

~~- [ ] Design has approved the experience~~

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,28 +1,17 @@
 # Next release
 
-Anticipated release: June 1, 2020
+Anticipated release: June 15, 2020
 
 #### 🚀 New features
 
-- Update Print Preview Layout ([#2182])
-- Make the "Saving" message persist for 750ms ([#2207])
-- Move the activity schedule information to the top of the Activity Overview section ([#2210])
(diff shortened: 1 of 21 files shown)


pphillips-fearless:
@jeromeleecms the migration script doesn't change the roles of any of the users.  Let me know if there are any users on staging or in production that need to be changed and I can add a check for that (or we can change them manually before this is merged).

PR #2289, opened 2020-06-10
Prevent users from entering negative FTE values

### LINKED ISSUE #2229: The FTE count accepts negative numbers
In the State Cost Categories subsection, the FTE count allows negative numbers to be entered. Though it's probably an edge case that someone would directly type in a negative number, the up/down arrows on the right side would make it easier to make this mistake if they are clicking down.

### This task is done when...
 - [ ] There is no way to add negative FTEs

radavis:
Resolves #2229

- Prevents users from entering negative FTE values


web/src/components/NumberField.test.js
```diff
@@ -20,6 +20,7 @@ describe('NumberField component', () => {
     ).toMatchSnapshot();
   });
 
+  // should text be selected if 'min' is set, and value <= min ?
```
radavis, line 23: Should it?

pphillips-fearless: I think the functionality should still work for zero, even if the min is not set.  So I would say leave it "as is" rather than change it to select min if min is set.

radavis:
> I would also set min=0 for "Number of Hours" on ContractorResourceForm

Is there already a ticket for this? If so, I would feel comfortable including that fix, too. Otherwise, we should include the rest of the team in the discussion, or probably just create ticket(s) to 1) create a list of all form fields that should not be negative, and 2) fix them.

web/src/components/NumberField.test.js
radavis: **todo**: delete this comment

jjames521:
@radavis @pphillips-fearless Looks good!

PR #2308, opened 2020-06-18
[Tech Debt] Cleanup warnings and errors

### LINKED ISSUE #2309: [Tech Debt] Snapshot Tests fail locally (because of TZ), PropType errors abound
### This task is done when...
 - [ ] snapshot tests pass locally
 - [ ] numerous PropType errors are fixed

radavis:
Resolves #2309 

Lots of PropType errors when running `npm run test`. This PR resolves many of them.

- adds `TZ=utc` to the `npm run test-watch` command, so that snapshot tests pass locally
- fixes numerous PropType errors

- [ ] ~The experience passes a basic manual accessibility audit (keyboard nav, screenreader, text scaling) OR an exemption is documented~
- [ ] ~The change has been documented~
  - [ ] ~Associated OpenAPI documentation has been updated~

- [ ] ~Design has approved the experience~
- [ ] ~Product has approved the experience~

This PR is entirely developer-facing, and does not change the UI, or how the application works.


web/src/containers/activity/All.js
```diff
@@ -17,7 +17,9 @@ const All = ({ add, activities }) => {
         {activities.length} program activities
       </h3>
       {activities.map((_, index) => (
-        <EntryDetails activityIndex={index} />
+        <EntryDetails
+          activityIndex={index}
+          key={index} /> // eslint-disable-line react/no-array-index-key
```
radavis, line 22: > When you don’t have stable IDs for rendered items, you may use the item index as a key as a last resort.

https://reactjs.org/docs/lists-and-keys.html

If there's a better option than using array index as key in this instance, please let me know.

web/src/containers/BudgetSummary.js
```diff
@@ -43,8 +43,12 @@ DataRow.propTypes = {
...
 const DataRowGroup = ({ data, year }) => (
   <Fragment>
-    {categories.map(({ category, title }) => (
-      <DataRow category={category} data={data[category][year]} title={title} />
+    {categories.map(({ category, title }, i) => (
+      <DataRow
+        category={category}
+        data={data[category][year]}
+        key={i} // eslint-disable-line react/no-array-index-key
+        title={title} />
```
radavis, line 51: > When you don’t have stable IDs for rendered items, you may use the item index as a key as a last resort.

https://reactjs.org/docs/lists-and-keys.html

If there's a better option than using array index as key in this instance, please let me know.

web/src/components/CardForm.test.js
```diff
@@ -11,10 +11,10 @@ describe('card form wrapper', () => {
     history.goBack.resetHistory();
   });
 
-  test('renders without an id on the container if id prop is false', () => {
+  test('renders without an id on the container if id prop is null', () => {
     expect(
       shallow(
-        <CardForm id={false} title="test" history={history}>
+        <CardForm id={null} title="test" history={history}>
```
radavis, line 17: Looking at the codebase, we never set the id when using the `CardForm` component. Seems like an unnecessary test.

radavis: Reading a little further into the codebase, looks like we're setting this to `'start-main-content'` when not specified. I'll leave this be.

web/src/containers/BudgetSummary.js
pphillips-fearless: I'd like to put more thought into this.  See the link here for some rationale as to why we shouldn't do this:  https://stackoverflow.com/questions/46735483/error-do-not-use-array-index-in-keys/46735689

web/src/containers/activity/ContractorResources.js
```diff
@@ -52,7 +52,7 @@ const ContractorResources = ({
 };
 
 ContractorResources.propTypes = {
-  activityIndex: PropTypes.number.isRequired,
```
pphillips-fearless, line 55: wouldn't it be better to require this to be a number?  It will always be an integer or if it's a string (not sure how that happens) it can be converted

web/src/containers/activity/Costs.js
```diff
@@ -20,7 +20,7 @@ const Costs = ({ activityIndex }) => (
 );
 
 Costs.propTypes = {
-  activityIndex: PropTypes.number.isRequired
+  activityIndex: PropTypes.string.isRequired
```
pphillips-fearless, line 23: see prior comment about preference for number

web/src/containers/BudgetSummary.js
pphillips-fearless: one thing I've done in the past is to use two variables I had handy, concatenated with a hyphen, like `${category}-${year}`

radavis: > We don’t recommend using indexes for keys if the order of items may change. 

https://reactjs.org/docs/lists-and-keys.html

In this case, according to the react docs, I think it's safe to leave it as is. However, I'm happy to change the key values to whatever you would like them to be.

web/src/containers/activity/ContractorResources.js
radavis: `activityIndex` is extracted from the URL, which is a string.

web/src/containers/BudgetSummary.js
radavis: Drilling down into the documentation for the `react/no-array-index-key` rule...

> If there is nothing unique about the items, [...] then you may want to disable this rule with an override.

https://github.com/yannickcr/eslint-plugin-react/blob/master/docs/rules/no-array-index-key.md#when-not-to-use-it

This code is iterating over a `const` defined earlier in the file: 

```
const categories = [
  { category: 'statePersonnel', title: 'State Staff' },
  { category: 'expenses', title: 'Other State Expenses' },
  { category: 'contractors', title: 'Private Contractor' },
  { category: 'combined', title: 'Subtotal' }
];
```

I think we're safe using the index as a key, in this case.

web/src/containers/activity/All.js
radavis: Turns out there was a `key` prop on the activity object.

web/src/containers/activity/ContractorResources.js
radavis: No, I'm wrong. Going up a few component levels, `<EntryPage />` coerces the activityIndex url param to a number. Will fix.

https://github.com/18F/cms-hitech-apd/blob/master/web/src/containers/activity/EntryPage.js#L9

web/src/containers/BudgetSummary.js
pphillips-fearless: agreed

web/src/containers/activity/ContractorResource/ContractorResourceReview.test.js
```diff
@@ -37,13 +37,17 @@ describe('the ContractorResourceReview component', () => {
   };
 
   it('renders correctly', () => {
-    const component = shallow(<ContractorReview {...props} />);
+    const component = shallow(
+      <ContractorReview {...props}>
+        <div/>
```
mgwalker, line 42: @radavis If you remember, why is this child element necessary here? As far as I can tell, the contractor review component doesn't require any children, but I may just be missing something.

radavis: I think it is because the Review component, which is being used by the ContractorReview component, has `props.children` marked as `isRequired`

PR #2315, opened 2020-06-22
Add an “Add another activity” button to the last page of the last activity

### LINKED ISSUE #2282: Add an “Add another activity” button to the last page of the last activity
this is based on original design ticket #2116, decisions are summarized here

Using the same patterns we have in the app, add an “Add another activity” button to the end of the last page of the last activity, on the left, as shown in this prototype.

Mockup/Prototype
https://www.figma.com/proto/XHJAkJqswl9GB7E6dqTard/ActivityListPrototypes?node-id=1%3A35&viewport=539%2C431%2C0.2764817178249359&scaling=min-zoom

Clicking the “Add another activity” button will follow the same pattern as the “Add another activity” button on the Activity List page. It will first take the user back to the Activity List page, and when they arrive there, a new Activity will have been created, with the title “Untitled”.

### This task is done when...
 - [ ] we've added an “Add another activity” button

pphillips-fearless:
Add an “Add another activity” button to the last page of the last activity.  

Clicking the “Add another activity” button will follow the same pattern as the “Add another activity” button on the Activity List page. It will first take the user back to the Activity List page, and when they arrive there, a new Activity will have been created, with the title “Untitled”.

Closes #2282 

- added an “Add another activity” button


web/src/styles/scss/secondary-nav.scss
```diff
@@ -1,11 +1,11 @@
 @import '@cmsgov/design-system-support/src/settings/index';
 
-.next-prev-buttons {
+.pre-button-section-break {
   margin: 32px 0px;
   padding-top: 32px;
```
radavis, line 5: Much of the rest of the app uses `$spacer-x` when declaring margin and padding values. Any chance we could use those, here?

`$spacer-1` is defined as `8px`, so you want `$spacer-4` in place of `32px`

pphillips-fearless: thanks for the suggestion, I hadn't noticed the $spacer-x variables

jeromeleecms:
@pphillips-fearless Can you change the page focus to the new edit link after you click "Add new activity?" With Richard's changes to the activity list styling, the screen reader would announce to the user that you can edit the new untitled activity.

pphillips-fearless:
@jeromeleecms that would not be easy.  Pages in the app don't have any idea what the previous page or the previous action was, so when we get to the activity list page we don't know that the last thing that happened was that the "add new activity" button was pressed.  In order to make that happen I would need to set a state flag somewhere (like in the redux store) and then clear it after the focus was set.  That feels very kludgy.  Can we discuss other options?

jeromeleecms:
@pphillips-fearless Of course -- let's grab a little bit of time tomorrow (Thursday)?

jeromeleecms:
Per our discussion, let's convert this button into a link and have the screen reader handle the announcing of page transition.

beparticular:
@mgwalker is it possible to have it styled the same way, but to be part of the previous/continue component? That's where I imagined it would live being that it always needs to be at the very end of an activity. Thx!

pphillips-fearless:
@mgwalker and @beparticular I like the way you're thinking.  I did that when I created the next/previous buttons, I'm not sure why I didn't do it here.

jeromeleecms:
@pphillips-fearless Also sounds good to me! I'll hold off on the review until you tell me it's ready for another look.

pphillips-fearless:
@mgwalker the change you requested is done, can you please take an other look?

web/src/components/SecondaryNav.test.js
```diff
@@ -1,9 +1,81 @@
...
+    expect(window.scrollTo).toHaveBeenCalledWith(0, 0);
+  });
+
+  it('maps state to props', () => {
+    const state = {
+      apd: {
+        data: {
+          activities: [
+            { anchor: 'activity-1', name: 'activity', key: '1' },
+            { anchor: 'activity-2', name: 'activity', key: '2' }
```
mgwalker, line 58:
```suggestion
            { name: 'activity', key: '1' },
            { name: 'activity', key: '2' }
```

The activities in the app state don't have an `anchor` property, so I think it's best if they also don't have them in a test fixture. The `anchor` property that gets returned by the selector should be computed.

web/src/components/SecondaryNav.test.js
```diff
@@ -1,9 +1,81 @@
...
+  beforeEach(() => {
+    props.add.mockClear();
+    props.jumpTo.mockClear();
+  });
+
+  it('renders correctly without add activity button', () => {
+    expect(
+      shallow(<SecondaryNav {...props} activeSection="activity-1-ffp" />).dive()
+    ).toMatchSnapshot();
+  });
```
mgwalker, line 30:
```suggestion
    expect(
      shallow(<SecondaryNav {...props} activeSection="activity-2-something-else"/>).dive()
    ).toMatchSnapshot();
  });
```

Also test the condition that we're on the last activity but *not* on the FFP section.

PR #2317, opened 2020-06-24
The $0 placeholder in dollar fields doesn't always disappear

### LINKED ISSUE #2230: The $0 placeholder in dollar fields doesn't always disappear
We've seen in user research that people were often pasting into the input fields from another place. When doing so, the 0 didn't always go away until they clicked out of the box. This confused them and was unexpected. We should update this so the 0 goes away when you click into the box.

### When nothing is selected. 

Screen Shot 2020-05-21 at 12 53 38 PM

### Clicking once into the box highlights it. 

Screen Shot 2020-05-21 at 12 53 44 PM

### However, they don't always just click once. If you click twice, it just adds to the end after the 0 placeholder.

Screen Shot 2020-05-21 at 12 53 22 PM

### This is what I'd expect - when you click into the field, the placeholder value clears away.

Screen Shot 2020-05-21 at 12 54 00 PM
 

### This task is done when...
 - [ ] The placeholder 0 values go away when you click into the box

pphillips-fearless:
We've seen in user research that people were often pasting into the input fields from another place. When doing so, the 0 didn't always go away until they clicked out of the box. This confused them and was unexpected. We should update this so the 0 goes away when you click into the box.  Closes #2230 

- The placeholder 0 values go away when you click into the box 


jeromeleecms:
@pphillips-fearless I'm seeing this weird behavior when tabbing through fields:

1) Before data entry:

image

2) If I tab to FFY 2021 costs, the "Cost with Benefits" for FFY 2020 dissappears.

image

3) If I enter 0 into FFY 2020 "Costs with Benefits" and than tab back to FFY 2021, the 0 stays...

image

4) ...Unless I tab between the fields again...

image

5) The good news is it looks like the $0 dollar value is maintained for the math once the user clicks "done."

image

pphillips-fearless:
wow, it looks like we've got our references crossed.  Taking a look.

pphillips-fearless:
1) I can observe the behavior, where the first DollarField on the page is emptied when the second one gets focus (the entanglement issue).
2) I put in a bunch of logging to see how the ref was set, to see if the two fields ended up with the same ref, but the two fields always had their own separate ref set correctly.  The id's of the underlying input controls get set automatically (they were unique) and I also tried making sure the names were unique.  
3) if I remove the mask from DollarField, the entanglement issue goes away (all fields on the page act independently)
4) if I switch from using ref and just use the React component onBlur and onFocus, leaving the mask off, it still works fine (on focus empty the field if value is zero, on blur set it back to zero, all fields on page act independently).  In retrospect, this is probably the preferred implementation.
5) if I add the mask back to the DollarField (with the React component onblur and onFocus being used instead of ref) then the onBlur doesn't reset the value to zero (it stays blank).  But on the bright side, the entanglement issue doesn't happen. 
6) my conclusion is that the entanglement issue is caused by some interaction between the ref onFocus event and the mask.  In the Mask.jsk file of the design system I see a comment about persisting the onBlur event in order to "de-bounce" it and there's a call to evt.persist().  It's conceivable that persisting the event could result in it being called for the wrong control???  I can create a small side project to demonstrate our issue to the design system team and ask them to investigate.
7) in the mean time, since we can't use the mask in either scenario, I can create a simple replacement for it (temporary until the design system team comes back with an answer).
8) let's discuss this at our co-working session on Monday and decide how to proceed.

mgwalker:
I don't think it's a design system issue. I think the issue is that the `setBlankIfZero` and `setZeroIfBlank` methods are directly changing the value of the DOM field rather than setting the local value that React is bound to. Setting the value of the DOM field causes an undesired `onChange` event (and might disconnect the binding? not 100% sure), whereas setting the local prop value does not (i.e., calling `setLocal()`). The previous variation of this worked because we weren't changing the DOM field's content, but rather its focus state.

I _think_ the proper way to handle this will be to move the `setBlankIfZero` into the React component and, instead of setting `target.value = 0`, calling `setLocal(0)`. Then for the blur effect, you shouldn't actually need to do anything. The existing blur handler should already take care of it.

The only reason we ever connected to the DOM element is because the Design System doesn't expose a `focus` event, so it was the only way to respond to the component being focused at all, unfortunately.

pphillips-fearless:
Thanks for your comment @mgwalker, I had pulled out most of the stuff that directly changed the DOM by the end of Friday, but I still had the onFocus changing e.target.value rather than using setLocal() and that was causing weirdness that I couldn't figure out.  When I changed that last thing it fixed everything.  I think you were right in your suspicion about it disconnecting the binding.  That's the only thing I can think could have caused all the weirdness I was seeing (sometimes it worked and sometimes it stopped working).  Anyway, you saved me from pulling my hair out!

PR #2334, opened 2020-07-02
Text Boxes for TinyMCE don't update when navigating between "Activity Overview" sections between Activities

pphillips-fearless:
Fixed the TinyMCE bug.  The issue is that the constructor wasn't firing when you switched between activities if you stayed on the same page (Overview) so the content wasn't being updated (constructor copies content from props to state).  I thought about updating state.content from props.content in componentDidUpdate but that is an anti-pattern.  I also thought of getting rid of state.content and just using props.content but the Editor control manipulates the content so better to separate props and state.  Then I found the article at https://reactjs.org/blog/2018/06/07/you-probably-dont-need-derived-state.html#recommendation-fully-uncontrolled-component-with-a-key which suggested using a key to force the control to re-render when switching activities.

I'm open to discussing/re-thinking the strategy.  Also, I couldn't come up with a new test for this because of the React re-rendering involved.

- added key property to the RichText control


pphillips-fearless:
@mgwalker I'm curious about your thoughts about my notes in the PR description

mgwalker:
I popped open the code in my editor to take a look at how it worked before, and I was curious if you had thoughts about why we're using component state instead of binding directly to the prop. It seems to me that if we yank out the component state and use the props directly, that might also solve the problem much more simply and it'd also make the `RichText` component less complex.

I don't remember why exactly we used component state, but I think generally with React input components, you use state when you need to manage it separately from the normal lifecycle of the contents of the input element. In this case, it _looks_ like we're not doing anything special with the state - when the input value changes, we update the state and then pass along the change with the `onSync` event. It may be that we used state to solve an earlier problem that no longer exists. (We used to use a different text editor and it was wonky; perhaps it did something that we needed to protect against.)

pphillips-fearless:
I made the change to back out the state version of content and to just use the props version.  Testing looks good and the change is much simpler.  @mgwalker and @jeromeleecms it's ready for another look.

mgwalker:
There's a weird race condition happening. Here are the steps to reproduce in dev:

1. reseed the database so everything is nice and clean (once the race condition happens in a given database, it doesn't seem to reoccur)
2. open the app
3. navigate to the 1st activity and notice that it has an overview description in the first rich text box
4. navigate to the 3rd activity and notice that it has its own overview description in the rich text box
5. navigate back to the 1st activity and notice that its overview description is now the same as activity 3's

pphillips-fearless:
@mgwalker I was struggling with one of the test cases on Friday because of a timing issue, I couldn't get it to await a promise of a promise.  An easy solution was to switch the upload call inside uploadImage to use await, but take a look (I just pushed it) and if you have an idea for a better way let me know and I'll refactor it.

pphillips-fearless:
I just did a test in PR-2343 to see if images are displaying there, and they're not.  So that tells me that this branch didn't introduce a new issue related to displaying images.  Images are stored as files, which means the environment has to be set up to handle file storage in order to display them.  At one point I asked @mgwalker which environments are set up to display images, and he couldn't remember.  I know my local dev environment isn't set up to see them.  It looks like the preview environments aren't either (which makes sense).  I don't see a test case on staging to see if images are displayed, but I'll keep trying to find one to check that.  I'm sure production is set up to display them, but I don't have visibility there.

mgwalker:
There are two things happening with the images. One I haven't fully investigated yet and one I have and could be problematic to fix. Either way, it's a bug and is probably already affecting production.

1. The URL that TinyMCE is inserting into the image `href` attribute is incorrect. For preview deployments it should be `/api/apds/.../`, but instead it's `https://api//apds/...`. I assume it's an issue with the URL being relative and TinyMCE is "helping" by adding a protocol and interpreting it as a full path. This only affects preview. The image URL is built on the client side by combining the relative image URL returned from the API with the full API URL. In staging and production, the API URL is absolute, so combining the two creates a new absolute URL. In preview deployments, the API URL is also relative, so the combined URL remains relative. Not clear to me offhand if there is a fix, but if it only affects preview deployments, it seems like a relatively low priority.

2. The API protects images with authorization just like everything else. A user can only view an image if it is associated with an APD that they have access to. The issue is that we now require authentication to happen via the `Authorization` header, but there's no (immediate) way to set that header on `` tags, which is what the TinyMCE editor is using. This worked previously because we were using cookies for authentication, and those got passed automatically by the browser – no need to set anything. There aren't a lot of good options here, but some options do exist:
    1. My current strong lean would be towards something like a [Service Worker](https://developer.mozilla.org/en-US/docs/Web/API/Service_Worker_API) to intercept all calls to the API. That service worker could then inject the authorization header (see [this StackOverflow answer](https://stackoverflow.com/a/54793004) for an example of exactly how to do it without losing anything else that may have been set on the request) on the fly. We would also remove the code that Richard added to the API wrapper since it'd be incorporated by the service worker. The downside? Service workers don't exist in IE, full stop. They work in all the modern browsers, but not IE. So if we went the service worker route, images could be *uploaded* from IE, but they couldn't be *viewed* from IE.

    2. Alternatively, we can make images publicly accessible. My instinct is that, while the images are probably mostly safe for public exposure, some won't be. Even among those that are safe, I can imagine that states (and likely CMS security folks) would be squeamish about making everything wide open like that. That said, this is an easy, universal solution. It'd just take commenting out a couple of lines of code in the API.

    3. Rewrite the contents of the RichText field on-the-fly to append the user's token to any image hrefs found in the content. This would also require tweaking the API to look for authentication tokens in the query string along with the headers. It's also not my favorite approach because it seems fragile. It might have cacheing implications, too, since caches sometimes ignore query strings, so a malicious actor could potentially retrieve an image from an intermediate cache even though the associated token had expired.

Hopefully there's another method that's also universal, but in my brief dive, I couldn't find anything. In any case, there should probably be a bug issue related to these images. This PR didn't introduce the problem, though. The bug came about when we migrated to the `Authorization` header, and we just plain didn't catch it. I did not even *think* about `` tags when Richard did all that work, and I imagine he didn't either. (It was 100% the right work to do, though.) What an unexpected edge case...

jeromeleecms:
Confirming -- it affects all versions of the app from preview site, staging, to production.

pphillips-fearless:
I was about to make a new task for this, but then we had our OKTA meeting and I realized it's all going to change again.  In a previous meeting they said that their react control uses cookies/sessions, so maybe this will fix itself (I'm an optimist, can you tell?).  I'll still make a task for it so we don't forget about it, but depending on timing we might want to hold off on doing anything until OKTA is implemented.

PR #2338, opened 2020-07-08
Update Activity Schedule Summary page

### LINKED ISSUE #2318: Update Activity Schedule Summary page
Break the existing table on the Activity Schedule Summary page into two separate tables with summary on top and a dial down by activity on the bottom.

For the style of the tables we can keep consistent with the current styling as we still haven't deployed the new styles. The hierarchy would work the same as was done with the Summary Budget by Activity table.

The three views (builder, export, print) are below.

BUILDER VIEW

EXPORT VIEW

PRINT VIEW

### This task is done when...
 - [ ] the existing table on the Activity Schedule Summary page has been broken into two separate tables as shown in the attached mockups

pphillips-fearless:
Break the existing table on the Activity Schedule Summary page into two separate tables with summary on top and a dial down by activity on the bottom.  Closes #2318 

- breaks the existing table into two tables

### DIFF
--- a/web/src/containers/ScheduleSummary.js
+++ b/web/src/containers/ScheduleSummary.js
@@ -32,7 +32,7 @@ const ScheduleSummary = ({ activities }) => (
               {activities.map(({ name: activityName, dateRange }, i) => (
                 <tr
                   key={activityName}
-                  className="summaery-table--gray_row__highlight"
+                  className="summary-table--gray_row__highlight"
                 >
                   <td
                     className="ds-u-font-weight--bold ds-u-border-right--0"
--- a/web/src/containers/ScheduleSummary.js
+++ b/web/src/containers/ScheduleSummary.js
@@ -16,21 +16,68 @@ const ScheduleSummary = ({ activities }) => (
             {t('scheduleSummary.noDataMessage')}
           </div>
         ) : (
-          activities.map(({ name: activityName, dateRange, milestones }, i) => (
+          <table className="budget-table">
+            <caption className="ds-u-visibility--screen-reader">
+              Activity List Overivew
+            </caption>
+            <thead>
+              <tr className="budget-table--row__primary-header">
+                <th scope="col">Activity List Overview</th>
+                <th scope="col" className="ds-u-text-align--left">
+                  Activity Date Range
+                </th>
+              </tr>
+            </thead>
+            <tbody>
+              {activities.map(({ name: activityName, dateRange }, i) => (
+                <tr
+                  key={activityName}
+                  className="summaery-table--gray_row__highlight"
+                >
+                  <td
+                    className="ds-u-font-weight--bold ds-u-border-right--0"
+                    style={{ width: '70%' }}
+                  >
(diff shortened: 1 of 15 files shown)


web/src/containers/ScheduleSummary.js
```diff
@@ -16,21 +16,68 @@ const ScheduleSummary = ({ activities }) => (
...
+          </div>
+        ) : (
+          activities.map(({ name: activityName, milestones }, i) => (
             <table key={activityName} className="budget-table">
               <caption className="ds-u-visibility--screen-reader">
                 Activity {i + 1}: {activityName}
               </caption>
               <thead>
+                <tr className="budget-table--row__primary-header">
+                  <th scope="col" style={{ width: '70%' }}>
```
mgwalker, line 68: Is the goal to push the date column as far to the right as possible? If so, would it be better to set the width of the name column to 100% and let the browser do the math to figure out how wide it can really be?

web/src/containers/ScheduleSummary.js
```diff
@@ -16,21 +16,68 @@ const ScheduleSummary = ({ activities }) => (
...
+                <th scope="col" className="ds-u-text-align--left">
+                  Activity Date Range
+                </th>
+              </tr>
+            </thead>
+            <tbody>
+              {activities.map(({ name: activityName, dateRange }, i) => (
+                <tr
+                  key={activityName}
+                  className="summaery-table--gray_row__highlight"
```
mgwalker, line 35:
```suggestion
                  className="summary-table--gray_row__highlight"
```

typo

web/src/styles/scss/tables.scss
```diff
@@ -113,6 +113,13 @@ td {
         background-color: $color-warn-lightest;
       }
     }
+    /* used on a tr to highlight a whole row */
+    .summaery-table--gray_row__highlight {
```
mgwalker, line 117:
```suggestion
    .summary-table--gray_row__highlight {
```

typo

web/src/containers/ScheduleSummary.js
pphillips-fearless: @mgwalker I tried taking off the 70% and the right column is all over the place -- based on the width of the left column text is where the right column text started.  So the 70% sets the starting point of the right column.  70% is the point at which it looks "nice".   You're right about 100% -- I tried it at 100% and the right column was pressed all the way over to the right.  It looked a little squished so I moved it back to 70%.

web/src/containers/ScheduleSummary.js
pphillips-fearless: oops, that's embarrasing!

web/src/containers/ScheduleSummary.js
pphillips-fearless: I forgot to mention...  70% is what it had been before.  I didn't add it, just moved it up from where it had been used before.

pphillips-fearless:
@jeromeleecms -- normally tables don't get focus when you tab through an HTML page.  I tested with VoiceOver and the screen reader reads the table OK.

PR #2351, opened 2020-07-15
[TechDebt] Change redux actions from symbols to strings

### LINKED ISSUE #2349: Redux Action Types show up as `<UNDEFINED>` in Redux DevTools
Redux Action Types show up as `<UNDEFINED>` in Redux DevTools. 

Screen Shot 2020-07-15 at 3 09 33 PM

It would be helpful to change these to string constants so that we can see what's happening with each action that is dispatched.

https://redux.js.org/faq/actions#why-should-type-be-a-string-or-at-least-serializable-why-should-my-action-types-be-constants

### This task is done when...
 - [ ] Action types are changed to string constants

radavis:
Resolves #2349 

- Changes redux actions from symbols to strings so that they are visible in ReduxDev Tools

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,8 +10,11 @@ Anticipated release: July 20, 2020
 
 #### ⚙️ Behind the scenes
 
+- Change Redux Action Types from symbols to string constants ([2349])
+
 # Previous releases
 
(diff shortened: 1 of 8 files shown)


radavis:
Screen Shot 2020-07-15 at 4 19 15 PM

I can now see what actions are being dispatched! 😀

PR #2383, opened 2020-07-23
The hourly resource and FFY costs are not overwriting as expected

### LINKED ISSUE #2236: The hourly resource and FFY costs are not overwriting as expected
In the Private Contractor Costs subsection, there is a place to enter the FFY costs and optionally select that it's an hourly resource. 

If you enter the FFY costs with "no" selected for hourly resource, that amount shows up in the final FFY costs. If instead you enter the FFY costs and then select "yes" for hourly resource, it'll apply that number to the FFY costs overwriting what you had in there before. (So far this all seems fine and as expected.) If you then go back and say "no" for hourly resource, you can change the FFY costs to something else. (Still good so far.) 

The issue is when you say "yes" again for hourly and the numbers for FFY are still using the previous numbers you put in, not the hourly ones. You have to select into the hours box and change the numbers or retype the one that was already there in order to get those to apply.

### Note that hourly resource is selected, but the FFY costs are the old numbers. It's not showing the $1 as I'd expect. 

Screen Shot 2020-05-21 at 2 19 59 PM

I'd expect that when they click into hourly, that takes over and those numbers will overwrite the FFY numbers (if they've been entered). If they change it back to "no" for hourly, then the numbers that were in hourly would get erased and would need to be added back in if that's what they need.

### Steps to reproduce:
1. Add a new contractor
2. Keep "no" selected and enter 100 into each of the FFY boxes
3. Now change hourly to "yes"
4. Add 1 hour and $1 into each of the boxes
5. Note that FFY now shows $1 for each year
6. Now go back and change hourly to "no"
7. Change FFY costs back to 100 each
8. Now change hourly back to "yes"
9. Note that the FFY costs still show 100 each even though you've changed hourly to yes. It should show $1 for each. This doesn't get updated until you change the hour amount (or just delete the 1 and put it back in)

### This task is done when...
 - [ ] When toggling between hourly and non-hourly, the FFY amount shows up accurately

pphillips-fearless:
When toggling between hourly and non-hourly, the total cost amount wasn't going back to the correct remembered amount.  Closes #2236 

- Added code to change handlers to remember non-hourly costs
- refresh total costs each time the user switches between hourly and non-hourly


mgwalker:
Since code reviews will be peer starting next week, I'm going to add @thetif as a reviewer too.

jeromeleecms:
Just a note - this seems to work as intended. I did notice that if you click back to "No" after toggling between the resource, that the new hourly rate calculation is maintained. Not a deal breaker and potentially an edge case, but this does solve the problem of the total cost not being updated when switching between the fields.

PR #2537, opened 2020-09-28
Bbrooks/2535 url testing strat

cthulhuplus:
This will auto-assign one of our available elastic IP address to a test/preview instance allowing a small set of explicit IP addresses/URLs to be added to Okta.

- Instance is provisioned with elastic IP
- Elastic IP/URL is whitelisted and working in Okta


cthulhuplus:
Current test URL: https://ec2-18-215-204-56.compute-1.amazonaws.com/

Changes to provisioning of test instances so that they receive an elastic IP and related public DNS so that we can whitelist a fixed set of URLs in Okta.

thetif (approved):

Issue #2503, opened 2020-09-09
Accessibility - Aria Labeling of Input Fields


rohitpaul:
Input fields on the "Proposed Budget" page have an aria label of "ehAmt payments for X, quarter Y. Enter amount in dollars." This is repetitive and can be an unpleasant experience for a user using assistive technologies.

### This task is done when...
 - [ ] Recommend using `aria-labelledby` tag.

Reference: [w3.org](https://www.w3.org/TR/WCAG20-TECHS/ARIA9.html#ARIA9-ex2)

PR #2551, opened 2020-10-02
Accessibility - Aria Labeling of Input Fields #2503

tbolt:
This addresses Accessibility - Aria Labeling of Input Fields 

#2503 

- Updates the "Proposed Budget" table inputs to use aria-labelledby
- Now screen readers should read "Column Label, Row Label" to provide a less-repetitive and more friendly description


web/src/containers/IncentivePayments.js
```diff
@@ -165,6 +168,7 @@ const IncentivePayments = ({
                         <NumberField
                           className="budget-table--input-holder"
                           fieldClassName="budget-table--input__number"
+                          aria-LabelledBy={'q' + q + ' ' + 'ep-count'}
```
radavis, line 171: I believe if you use a `ariaLabel=...` prop on this component, it will be passed through to the cms-gov design system `<TextField />` component, which may yield the desired results within the HTML output.

radavis: Our NumberField Component: https://github.com/CMSgov/eAPD/blob/master/web/src/components/NumberField.js

which uses the TextField component from the design system:
https://design.cms.gov/components/text-field/

tbolt: Yea, when I looked at the rendered markup initially I was puzzled how it was getting an `aria-label` (as well as a `<label>` as a sibling above it). Looks like when passing `label=...` to the cmsds it's being used for the `aria-label. Problem is I ideally _don't_ want the `aria-label` and only want the `aria-labelledby` but that's not supported
https://design.cms.gov/components/masked-field/

radavis: Ah, I see. What happens if you pass `ariaLabel={null || undefined}` and set `aria-labelledby={...}` on the component?

tbolt: I'll try that

radavis: There's also the option of filing a ticket at `https://github.com/CMSgov/design-system/issues`. It's possible there's a bug around overriding component properties, or possibly a workaround.

tbolt: Good call, I'll open an issue as this might benefit other folks and use cases. 

I looked into it some more though. We currently don't set `ariaLabel` explicitly, we are only setting the `label` which is a required prop of the component. But here's how the internals of the cmsds textfield component works:
```  ariaLabel() {
    if (this.props.ariaLabel) {
      return this.props.ariaLabel;
    } else if (this.props.mask === 'currency') {
      return `${this.props.label}. Enter amount in dollars.`;
    }
  }```

So the aria-label is set either way in the rendered output. Setting `label` to null/undefined will just set the `aria-label` to `undefined. Enter amount in dollars.`

thetif (approved):

thetif (approved):

thetif (approved):

PR #2564, opened 2020-10-09
Add a label to the rich text editor <textarea> element

### LINKED ISSUE #2546: Add a label to the rich text editor <textarea> element
This is to address the rich text editor not announcing anything to the screen reader when it is tab-navigated to.

Rohit's findings/solution:
Check 95: textarea element missing an associated label.
Repair: Add a label element immediately before or after the textarea element. Set the for attribute value of the label element to the same value as the id attribute value of the textarea element. Add label text to the label element. Or, set the title attribute value to the textarea element to the label text. Or, add a label element that surrounds the textarea element and add label text.
Describe the task here and link or add dependencies to any applicable issues...

### This task is done when...
 - [ ] Rich text editing boxes (TinyMCE) components are updated to address the above concern

radavis:
Resolves #2546 

- Adds labels to `<textarea>` elements created by tinymce

Note: this does not necessarily make the `<RichText>` component accessible. It simply fixes an issue reported by Wave. In order to label the tinymce editor iframes correctly, this tinymce-react issue will need to be addressed: https://github.com/tinymce/tinymce-react/issues/177


thetif (approved):

thetif (approved):

beparticular:
@radavis I'm testing with Chromevox and I'm hearing "edit text" 2-3 times in each field.

With VoiceOver I heard "clickable text area" and 
Screen Shot 2020-10-15 at 4 31 08 PM

cc @tbolt re: our process conversation yesterday for testing with screenreaders - is this kind of major difference expected with Chromevox vs VoiceOver?

radavis:
@beparticular, different screen readers are programmed differently, so there will be some variation in what they do and say. 

It's possible we are creating more work for ourselves by using multiple different screen readers. Perhaps we could normalize on testing with VoiceOver on Mac and NVDA on Windows, since those are two of the most popular. https://webaim.org/projects/screenreadersurvey8/#primary

Regarding the accessibility of tinymce, there's not much more I can do to improve the situation until this issue is resolved https://github.com/tinymce/tinymce-react/issues/177

beparticular (approved):
I think this is ok for now. As we look deeper into RTEs and other issues, lets continue to refine our 508 testing process as our PO/PM returns.

CSwartzHMA (approved):
Based on note above from @beparticular, sounds like this works for now.

PR #2626, opened 2020-11-02
Tbolt/2562 consistent table labels

### LINKED ISSUE #2562: Use a consistent method for labeling tables
We should re-evaluate all of the ways we are labeling tables.  Currently we use a variety of methods including:
- `<caption>`
- some use a heading like `<h4>`
- some use a styled p (ie. `<p class=“ds-h4”>`)
- in some cases there are multiple labels with the same copy, one label hidden that is announced to screen readers, and one that is visible on the page. 

As well, some tables do not use an explicit label outside of the table(ex. the read-only view), instead relying on a primary header row. We should re-evaluate if there is a better and more consistent method for these.

### This task is done when...
 - [ ] Reduce the number of labeling methodologies to a minimum set of best practices
 - [ ] Update all tables to follow the above set practices

tbolt:
Partially resolves #2562 

- Updates tables that use redundant `<caption>` for tables
- Removes hidden `<caption>`s in some cases


thetif (approved):

beparticular (approved):

jeromeleecms (approved):
Does what it says!

PR #2672, opened 2020-11-20
seeding roles, activities and their map table

thetif:
updated the way states, roles, activities, and roles-activity-mapping are seeded; added seeding to build process

- adds seeding to the prod build process
- only sees states if there are no states
- deletes all of the activities before re-adding them
- deletes all of the roles-activities before re-adding them
- adds logging to states and roles seeding

### DIFF
--- a/api/migrations/20200921100859_change-role-names.js
+++ b/api/migrations/20200921100859_change-role-names.js
@@ -16,16 +16,16 @@ exports.up = async knex => {
     .update({ name: 'eAPD Federal Leadership', isActive: false });
 
   await knex('auth_roles')
-    .where({ name: 'federal' })
-    .update({ name: 'eAPD Federal', isActive: false });
+    .where({ name: 'federal SME' })
+    .update({ name: 'eAPD Federal SME', isActive: false });
 
   await knex('auth_roles')
     .where({ name: 'state coordinator' })
     .update({ name: 'eAPD State Coordinator', isActive: false });
 
   await knex('auth_roles')
-    .where({ name: 'state' })
-    .update({ name: 'eAPD State', isActive: false });
+    .where({ name: 'state SME' })
+    .update({ name: 'eAPD State SME', isActive: false });
 };
 
 exports.down = async knex => {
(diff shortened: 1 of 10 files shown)


api/migrations/20200921100859_change-role-names.js
```diff
@@ -16,16 +16,16 @@ exports.up = async knex => {
...
 
   await knex('auth_roles')
     .where({ name: 'state coordinator' })
     .update({ name: 'eAPD State Coordinator', isActive: false });
 
   await knex('auth_roles')
-    .where({ name: 'state' })
-    .update({ name: 'eAPD State', isActive: false });
+    .where({ name: 'state SME' })
+    .update({ name: 'eAPD State SME', isActive: false });
```
radavis, line 28: If this migration filename appears in the `knex_migrations` table, this code will not run.

You probably want to generate a new migration and make the necessary changes to the data.

thetif: I already changed it in the database, the database value is "state SME" and "federal SME" and this migration originally missed it. I changed it so it would work in prod.

api/seeds/shared/states.js
```diff
@@ -1,5 +1,17 @@
...
 const { states } = require('../../util/states');
 
 exports.seed = async knex => {
-  await knex('states').insert(states);
+  logger.info(`Begging to seed the ${chalk.cyan('states')} table`);
+  const total = await knex('states')
+    .count('id')
+    .first();
+  logger.info(`${chalk.cyan('states')} table currently has ${total.count}`);
+  if (total.count.toString() === '0') {
```
radavis, line 11: Is this right?

api/seeds/shared/roles-and-activities.js
```diff
@@ -135,4 +177,5 @@ exports.seed = async knex => {
       'eAPD State SME'
     ])
     .update({ isActive: false });
+  logger.info('Updated active status for roles');
```
radavis, line 180: Running the seed command on my local database yields the appropriate associations in these tables.

api/seeds/shared/states.js
thetif: The documentation says it comes back as a bigint or a string, so I converted it to a string just in case and tested it that way.

thetif: But also when I tested it on my system it did wipe out the states and saw that it was empty and added them back without duplicating them.

radavis (approved):

PR #2678, opened 2020-11-30
removed button and link to profile page

thetif:
The profile page is not longer relevant.

- remove the "Manage account" button from the navigation dropdown
- remove the route to go to the profile page

### DIFF
--- a/web/src/routes.js
+++ b/web/src/routes.js
@@ -26,7 +26,6 @@ const routes = [
     isCard: true
   },
 
-  { path: '/me', component: MyAccount, isPublic: false, isCard: true },
   { path: '/login', component: Login, isPublic: true, isCard: true },
   { path: '/logout', component: Logout, isPublic: false },
   { component: NoMatch, isPublic: true }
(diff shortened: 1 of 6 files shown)


web/src/routes.js
```diff
@@ -4,7 +4,6 @@ import Dashboard from './containers/Dashboard';
 import ApdApplication from './containers/ApdApplication';
 import ApdViewOnly from './containers/viewOnly/Apd';
 import EditAccount from './containers/admin/EditAccount';
-import MyAccount from './containers/admin/MyAccount';
```
radavis, line 7: We probably want to delete this container component, and any associated unit test, too.

thetif: I think we will end up having a profile page, but we don't have any designs for it yet, so I wanted to leave it there for when we will need it.

jeromeleecms (approved):
Seems fine - tried with screen reader and keyboard nav. Tried to navigate to the old pages and the routes showed a 404 error, but you can't see the content of the error message. Worth spinning up as a separate issue?

image

thetif:
This ticket is for updating the Manage access page https://app.zenhub.com/workspaces/cms-eapd-5a3823ad3659e94d8ee569be/issues/cmsgov/eapd/2582. It really only lets the user request another state, so we don't want to expose that yet.

web/src/routes.js
radavis: But, isn't that what git is for?

radavis (approved):
👍

PR #2681, opened 2020-11-30
fixed migration

thetif:
Fixed issue with migrations removed route again

- update migration to set a default value for new uid field if none exists in the existing row
- removed the route to /me on the front end

~- [ ] Changelog is updated as appropriate~

### DIFF
--- a/api/migrations/20201123161901_modify-users.js
+++ b/api/migrations/20201123161901_modify-users.js
@@ -1,28 +1,9 @@
 exports.up = async knex => {
-  await knex.schema.dropTable('auth_sessions');
-  await knex.schema.dropTable('apd_versions');
-
-  // remove fields from users table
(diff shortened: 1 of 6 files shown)


jeromeleecms (approved):
Able to login on the preview site with no issue.

tbolt (approved):

api/migrations/20201123161901_modify-users.js
```diff
@@ -20,6 +20,7 @@ exports.up = async knex => {
     table
       .string('uid')
       .notNullable()
+      .defaultTo('not set')
```
radavis, line 23: I don't think this will work, since there's a unique constraint set on line 25.

thetif: should we clear out the table then

PR #2683, opened 2020-12-02
Allow requests for APD attachments to pass-through

### LINKED ISSUE #2348: Images aren't displaying in TinyMCE
There are two things happening with the images. One I haven't fully investigated yet and one I have and could be problematic to fix. Either way, it's a bug and is probably already affecting production.

The URL that TinyMCE is inserting into the image href attribute is incorrect. For preview deployments it should be /api/apds/.../, but instead it's https://api//apds/.... I assume it's an issue with the URL being relative and TinyMCE is "helping" by adding a protocol and interpreting it as a full path. This only affects preview. The image URL is built on the client side by combining the relative image URL returned from the API with the full API URL. In staging and production, the API URL is absolute, so combining the two creates a new absolute URL. In preview deployments, the API URL is also relative, so the combined URL remains relative. Not clear to me offhand if there is a fix, but if it only affects preview deployments, it seems like a relatively low priority.

The API protects images with authorization just like everything else. A user can only view an image if it is associated with an APD that they have access to. The issue is that we now require authentication to happen via the Authorization header, but there's no (immediate) way to set that header on  tags, which is what the TinyMCE editor is using. This worked previously because we were using cookies for authentication, and those got passed automatically by the browser – no need to set anything. There aren't a lot of good options here, but some options do exist:

My current strong lean would be towards something like a Service Worker to intercept all calls to the API. That service worker could then inject the authorization header (see this StackOverflow answer for an example of exactly how to do it without losing anything else that may have been set on the request) on the fly. We would also remove the code that Richard added to the API wrapper since it'd be incorporated by the service worker. The downside? Service workers don't exist in IE, full stop. They work in all the modern browsers, but not IE. So if we went the service worker route, images could be uploaded from IE, but they couldn't be viewed from IE.

Alternatively, we can make images publicly accessible. My instinct is that, while the images are probably mostly safe for public exposure, some won't be. Even among those that are safe, I can imagine that states (and likely CMS security folks) would be squeamish about making everything wide open like that. That said, this is an easy, universal solution. It'd just take commenting out a couple of lines of code in the API.

Rewrite the contents of the RichText field on-the-fly to append the user's token to any image hrefs found in the content. This would also require tweaking the API to look for authentication tokens in the query string along with the headers. It's also not my favorite approach because it seems fragile. It might have cacheing implications, too, since caches sometimes ignore query strings, so a malicious actor could potentially retrieve an image from an intermediate cache even though the associated token had expired.

Hopefully there's another method that's also universal, but in my brief dive, I couldn't find anything. In any case, there should probably be a bug issue related to these images. This PR didn't introduce the problem, though. The bug came about when we migrated to the Authorization header, and we just plain didn't catch it. I did not even think about  tags when Richard did all that work, and I imagine he didn't either. (It was 100% the right work to do, though.) What an unexpected edge case...

Because we're switching to OKTA for authentication, this may all change again once that is done, so this is a placeholder task that we probably want to wait to deal with until then

### This task is done when...
 - [ ] criteria 1
 - [ ] criteria 2
 
radavis:
Resolves #2348

Resolves https://github.com/advisories/GHSA-vrv8-v4w8-f95h

### This pull request...

- Allow requests for APD attachments to pass-through. Guessing an attachment fileId generated by AWS S3 would take somewhere around 10^69 years, by my calculations.
- Adds the tinymce media plugin to address this [GitHub security advisory](https://github.com/advisories/GHSA-vrv8-v4w8-f95h)


.circleci/config.yml
```diff
@@ -588,11 +588,11 @@ jobs:
           name: deploy to preview
           working_directory: ~/project/bin/preview-deploy
           command: |
-            if [ -n "$CI_PULL_REQUESTS" ]; then
+            if [ -n "$CIRCLE_PULL_REQUEST" ]; then
               apt-get -qq update
               apt-get -qq install jq -y
               pip install --quiet awscli
-              PRNUM=$(basename $CI_PULL_REQUESTS)
+              PRNUM=$(basename $CIRCLE_PULL_REQUEST)
```
radavis, line 595: `CI_PULL_REQUESTS` has been deprecated in favor of `CIRCLE_PULL_REQUEST` https://circleci.com/docs/2.0/env-vars/#built-in-environment-variables

api/logger/winston.js
```diff
@@ -8,7 +8,8 @@ const { LOG_CONSOLE, LOG_FILE, LOG_LEVEL, NODE_ENV } = process.env;
 const formats = [
   winston.format.timestamp(),
   winston.format.errors({ stack: true }),
-  NODE_ENV === 'development' ? winston.format.simple() : winston.format.json()
+  // NODE_ENV === 'development' ? winston.format.simple() : winston.format.json()
+  winston.format.json()
```
radavis, line 12: Debugging... Change this back

bin/preview-deploy/aws.sh
```diff
@@ -115,7 +115,7 @@ function createNewInstance() {
     --subnet-id "$AWS_SUBNET" \
     --tag-specification "ResourceType=instance,Tags=[{Key=Name,Value=eAPD PR $PR_NUM},{Key=environment,Value=preview},{Key=github-pr,Value=${PR_NUM}}]" \
     --user-data file://aws.user-data.sh \
-    --key-name eapd_bbrooks \
+    --key-name radavis \
```
radavis, line 118: Change back before merging.

bin/preview-deploy/aws.user-data.sh
```diff
@@ -324,7 +324,7 @@ git clone --single-branch -b __GIT_BRANCH__ https://github.com/CMSgov/eAPD.git
 # Build the web app and move it into place
 cd eAPD/web
 npm ci
-API_URL=/api/ OKTA_DOMAIN="__OKTA_DOMAIN__" npm run build
+API_URL=/api OKTA_DOMAIN="__OKTA_DOMAIN__" npm run build
```
radavis, line 327: An extra slash was being inserted into URLs. This was the culprit.

jeromeleecms (commented):
Are we able to limit uploading to only image files? I was able to switch the dropdown from "Image Files" to "All Files" and load a document. It allowed the upload, but the link was broken to it (which should be the correct behavior). But other than that, was able to upload, come back and forth, and the image is still present.

radavis:
@jeromeleecms, I am unsure if we are able to limit the types of files uploaded. Sounds like a feature request :)

bin/preview-deploy/aws.sh
thetif, line 118: should we remove both of these keys?

thetif (approved):

bin/preview-deploy/aws.sh
radavis: This line is going to change based on whomever needs to ssh into the preview server for a given branch.

The keys are stored on AWS. Specify which public key to copy onto the ec2 server, here.

cthulhuplus: Ideally we would remove access to the backend servers for normal operations and only deploy them out when we need access

jeromeleecms:
@radavis - Just tried to upload a file to the preview site (logged in via our em@il.com user), and I'm getting a broken link:

image

Also for baseline, was working yesterday.

radavis:
@jeromeleecms, I believe this is because the preview deploy failed. I'll let you know when it is back up.

tbolt (approved):
Looks good. Tested adding an image and worked perfectly. Persists after logging out and back in as well.

thetif (approved):

PR #2707, opened 2020-12-11
Tbolt/2583 state admin dashboard

### LINKED ISSUE #2583: State Admin dashboard
Create a state admin dashboard where the admins can see requests for access for their state and the status of their current and previous state users. Allow admin to view/approve/deny/revoke user access for their state.

### This task is done when...
 - [ ] the State Admin dashboard is created
 - [ ] the admin can select a user to view and manage
 - [ ] the admin can revoke a user's access to their state
 - [ ] the admin can approve a user's access request
 - [ ] the admin can reject a user's access request
 
tbolt:
Resolves #2583 

- Adds the state admin dashboard


beparticular (approved):
Ty and I walked through this and it looks great. 
Since I'm heading out at 4pm today I'm approving pending the question of if "federal admin" will show in the roles area and the language on the empty state changed to something like "No users on this tab at this time". Hooray!

jeromeleecms (commented):
Here's a little wierdness:
1) Click on Requests - Shows the new account:

image

2) Click on Active - Show's the pending account (is this the right behavior?)

image

3) Click on Inactive (looks right at this point)

image

4) Click back to active - but now it's empty:

image

What should I be expecting to see in this active tab?

thetif (approved):

jeromeleecms:
**Approving A User** - as before, saw the pending account, 

- clicked "Approve", selected the role and clicked "Save." 
- Account still shows as a request (instead of disappearing - is that the right behavior)? 
- Clicking on "Active" still shows the  previous behavior as the previous comment (pending account shows up)
- Clicking on "Edit Role" in the active tab, allows me to also select role and "Save" or "Cancel," but it doesn't update the account, so it still shows in "Requests"
- Clicking on "inactive," back to "active" causes the "active" tab to be empty (as previously noted)

**Revoking A User** - this works, after denying access, the pending account disappears from the "Requests" tab and moves to the inactive tab. Denied status shows properly:

image

**Trying to Restore Access** - this does **not** work, but probably due to the same issue as not being able to save users when granting them initial access(?)

jeromeleecms (commented):
Looks good.

PR #272, opened 2018-03-06
API endpoint "coverage"

mgwalker:
This PR adds some (off-by-default) code that instruments our API to collect a list of all of the endpoints that we register with it, and also to collect a list of every endpoint that is requested along with the HTTP status code that gets sent in response.

After running endpoints, we can combine these two lists plus our OpenAPI documentation to get a list of which registered endpoints are not documented, and which endpoint+status code combinations are untested.  It's kind of like coverage, but for endpoints!

screen shot 2018-03-06 at 3 03 50 pm

It'd be awesome if we were able to find a way to get this output to show up in our pull requests at some point.  Maybe if CircleCI could post a comment.

### This pull request is ready to merge when...
- ~~Tests have been updated (and all tests are passing)~~ New code is not for production
- [x] This code has been reviewed by someone other than the original author
- ~~The experience passes a basic manual accessibility audit (keyboard nav, screenreader, text scaling) OR an exemption is documented~~
- [x] The change has been documented
  - ~~Associated OpenAPI documentation has been updated~~

### This feature is done when...
- ~~Design has approved the experience~~
- ~~Product has approved the experience~~

PR #2714, opened 2020-12-16
Tforkner/2702 auth session

### LINKED ISSUE #2702: [Dev] Auth is expiring while users are using the system
The current set up for the frontend does not keep the session alive as the user uses the system. Update the frontend so that the session is kept alive. Alert the user when their session has ended.

### This task is done when…
This could include:
 - [ ] check if the user has an existing session
 - [ ] keep the session alive when the auto save is running
 - [ ] display an error message to the user when they have lost their session

thetif:
Users were having trouble with sessions not being extended automatically and not alerting them to when their session times out. closes #2702 

- If a user is navigating to a different page or APD session or if the user is typing in an APD there session should automatically be renewed
- If the users is inactive for 10 minutes, an alert will pop up to ask them if they want to continue
- If the user chooses to continue, then the alert will close and they can continue using the app
- If the user does not choose to continues, after another 5 minutes the session will time and and they will be automatically logged out
- If the user refreshes, they should maintain their session


CHANGELOG.md
```diff
@@ -5,6 +5,7 @@ Anticipated release: December 21, 2020
 #### 🚀 New features
 
 - Update endpoint for affiliations to filter by status ([#2682])
+- Update session management to warn users that there session is about to expire ([#2702])
```
cthulhuplus, line 8: their

jeromeleecms:
Reporting some findings:
1) When navigating between pages or to different parts of the same page, the session timer is not resetting
2) When the message box at the 10 min mark appears, if I navigate to the dashboard and then try to open the APD through the dashboard link, it seems like the app gets stuck at the loading screen. Refreshing the screen seems to reset the token which allows me to load the APD again.
3) Also when the message box at the 10 min mark appears, the system can no longer save:

image

4) Continuing with (3) After clicking continue in the 10 minute message box, the "There's been an unexpected error" message box doesn't go away. Typing / saving allows that second message box to go away. Is the expected behavior that the system should attempt a save when refreshing the session token? Or are we wanting the user to manually click save? I'm thinking that the system should attempt a save since the reason the "unexpected error" box appears is because the user was making an update to the system. 

5) Going to test 508 now, will report back, but wanted to get your thoughts here @thetif + @beparticular about how the behavior and look of (3) and (4)?

cthulhuplus (approved):
I logged in, waited 10 minutes, was warned I was going to be logged out, was then logged out.. I logged back in, waited 10 minutes, was warned I was going to be logged out, clicked the continue button and wasn't logged out, waited 10 more minutes, was warned again that I was going to be logged out, and was then logged out.

web/src/styles/scss/layout.scss
```diff
@@ -70,3 +70,20 @@ footer {
     margin-top: 0;
   }
 }
+
+.alert--session-expiring {
+  position: sticky;
+  top: -200px;
```
tbolt, line 76: we will want to refactor our layout so we don't have to use these hacks

CSwartzHMA (approved):
Was able to click around without saving and not get kicked out for over 15 minutes. Able to save changes after 10 minutes without seeing an error message. When idle for 14 minutes, the countdown timeout message appeared and I was able to clear it and save subsequent changes. When idle for 9 minutes, the countdown timeout session appeared. At the end of the countdown I was logged out. Seems to be working as expected.

web/src/styles/scss/layout.scss
thetif: completely agree

jeromeleecms (approved):
Approving this with the understanding that the continue button is working and the bugs are now logged in #272 (which will be resolved next week). Enabling the extension of the cookie solves a broader problem with being able to use staging at the moment.

PR #2739, opened 2020-12-28
Store PGDATA in Docker Volume, instead of a bind-mount

radavis:
When cloning repo, `pgdata` folder must be present. Otherwise, `docker-compose up` fails.

@cthulhuplus pointed out we should probably be using a docker volume to store this information, instead.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -50,7 +50,7 @@ workflows:
       - owasp zap scan:
           filters:
             tags:
-              only: /.*/
+              ignore: ignore-owasp
       - validate openapi:
           filters:
             tags:
--- a/.gitignore
+++ b/.gitignore
@@ -8,7 +8,7 @@ node_modules
 .nova
 
 npm-debug.log*
-pgdata
+pgdata/*
 __files
 api/test-data/files
 notes.md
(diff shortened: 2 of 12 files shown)


cthulhuplus:
I believe that the desired functionality that we want is for Docker to create the pgdata directory on the host and populate it with contents from the Docker container. If that is the case, I believe what we want to do is to not use a bind mount, but to use a volume:
https://docs.docker.com/storage/bind-mounts/
https://docs.docker.com/storage/volumes/
If that is correct I would suggest updating the postgres section of the docker-compose to something like:
```
services:
  db:
    image: postgres:10
    environment:
      - POSTGRES_DB=hitech_apd
      - POSTGRES_PASSWORD=cms
      - PGDATA=/pgdata
    volumes:
      - pgdata:/pgdata
    expose:
      - 5432
    ports:
      - 54325:5432

volumes:
  pgdata:
```

radavis:
Thanks, @cthulhuplus. I'm unsure why we were using a bind-mount, here. A volume seems to make more sense.

cthulhuplus (approved):
LGTM

mgwalker:
I added the bind mount a long time ago so the postgres data would survive rebuilding from docker-compose, and I did it the first way I found on stackoverflow or something. I'm not surprised it was wrong and I'm thrilled to see y'all making it better. 😄

PR #2757, opened 2021-01-07
Tbolt/2445 ffy warning

### LINKED ISSUE #2445: Warn users about lost data before they uncheck a FFY
**Change Log**:

- Chrome modal pop up with text "Unchecking Federal Fiscal Year (FFY) "Y" (Y being the FFY unchecked) will permanently delete any FFY "Y" specific data in the current APD." (This would refer to whichever FFY they have unchecked. Only one would be able to be unchecked at a time, so multiple FFYs are not needed in this alert.)
- Designate "Cancel" the default highlighted button
- "Cancel" and "Delete" as button text

**Mock up**

Original design task #2416

### This task is done when...
 - [ ] a Chrome modal pop up appears when a user unchecks a FY 
 - [ ] the modal is wired into the deletion process

tbolt:
Resolves #2445 

- When unchecking a FFY year a confirmation dialog is shown 


web/src/containers/ApdSummary.js
```diff
@@ -32,11 +32,20 @@ const ApdSummary = ({
   yearOptions
 }) => {
   const handleYears = e => {
-    const { value } = e.target;
-    if (years.includes(value)) {
-      removeApdYear(value);
+    const year = e.target.value;
```
thetif, line 35: you could do something like
`const { value: year, checked } = e.target;`

thetif (approved):

jeromeleecms (commented):
In the mockup, there was a title indicating "Warning" -- can that be added instead of what's highlighted below: 

image

tbolt:
@jeromeleecms I used the same pattern that is used when you delete an activity. It's using a default web browser API [Window.confirm](https://developer.mozilla.org/en-US/docs/Web/API/Window/confirm). There's no way to not have the highlighted part with this API. I could add text after the highlighted part to have  "Warning" then a line break then the rest of the text. Otherwise we would need to use a custom modal or another custom solution.

radavis (approved):
Looks great!

jeromeleecms (approved):
Thanks @tbolt -- if we have a precedent for this, that's fine. Maybe down the road we can look at a more unified approach to warnings/modals... food for thought.

beparticular (approved):

PR #967, opened 2018-08-23
Updates unit test for FFY Budget Table

jeromeleecms:
Per discussion on Slack, this unit test corrects the Unit Test for the FFY Budget Table Totals column

### This pull request changes...
- FFY totals by quarter should now add up to the correct value (instead of double their correct value)

### This pull request is ready to merge when...
- [ ] FFY Budget Table calculations are corrected

### This feature is done when...
- [ ] Dev Confirms Change to Unit Test is Accurate/Complete

PR #2829, opened 2021-01-29
vulnerability fix

thetif:
Updated the file processor we are using to actually manipulate the file and determine it's type.

- uses sharp image processor
- sets a max width and max height for images and will adjust their size if they are over either one
- converts all file types to png, which allows us to accept more image types


beparticular:
@thetif what is the max width/height?

thetif:
@beparticular the plugin I'm using allows you to resize the image. So I set the max width and height to be the size of a page, so it will resize images if they are too big for the page. Altering images is also a good way to make sure that the images are legit images and makes it more secure.

beparticular (approved):

jeromeleecms (approved):
This looks good - just needs a code review!

api/util/fileValidation.js
```diff
@@ -1,21 +1,40 @@
-const FileType = require('file-type');
+const sharp = require('sharp');
+
+const MAX_WIDTH = 700;
+const MAX_HEIGHT = 1100;
```
radavis, line 4: Any reason for choosing this particular pixel width and height? Just curious.

Will constraining the width and height this way change the aspect ratio of uploaded images?

Should we specify a max DPI, as well?

radavis (approved):
Code looks good to me. Added some questions/concerns to consider.

api/util/fileValidation.js
thetif: I picked those sizes because they fit inside a standard 8.5 x 11 page. It maintains aspect ratio. I just added one more setting in case it's too wide and too long.

We should be ok with dpi according to the dev who created the plugin "TIFF is pretty much the only bitmap format that cares about pixel density, thanks mainly to its use in the print industry, hence the tiff operation exposing xres and yres. JPEG doesn't deal with density, but some metadata added to JPEG can include values that might be interpreted as such - please see #967 for a related future possible enhancement."

Issue #2550, opened 2020-10-02
API returns 5xx HTTP status on validation errors


radavis:
See this [comment](https://github.com/CMSgov/eAPD/pull/2474#pullrequestreview-501258974)

### This task is done when...
 - [ ] validation errors from the api return 4xx HTTP status codes. https://restapitutorial.com/lessons/httpmethods.html
 - [ ] front-end is updated appropriately

radavis:
Consider returning a 404 error, or 422 from the API when validations fail.

https://restapitutorial.com/lessons/httpmethods.html

https://www.bennadel.com/blog/2434-http-status-codes-for-invalid-data-400-vs-422.htm

Issue #2962, opened 2021-03-16
[Dev] Create table to hold State Admin certifications


thetif:
State Admins have to be certified by the State Medicare Director (SMD). An email will be sent to every SMD on a yearly basis to certify or recertify Admins. These certifications should be stored in a database table. When a State Admin logs in, the system should confirm that they are still certified in this table. If they are no longer certified, their role should be downgraded to State Staff. A cron job should run every night to reconcile the roles against the certification table. Create an audit table for tracking changes to the certification table.

Create a state_admin_certifications table that holds the certified user's Okta username, state, the date of certification, and the SMD.
Create a state_admin_certification_audits table that tracks changes made to a certification and by whom.
Update the getUser method to check the state_admin_certifications table, if the user has the role of State Admin. If their certification is still valid, then keep their role as State Admin. If their certification has expired, then update their role to State Staff.

### This task is done when…
 - [x] there is a state_admin_certifications to track certification statuses
 - [x] there is a state_admin_certification_audits to track changes to certifications
 - [ ] ~the backend downgrades users with expired certification~
 - [ ] ~there is a cron job that runs nightly to reconcile the affiliations table and the certifications table~
Cancelling the cron job as this can all be handled with DB calls in 3029 when issuing the JWT.

PR #3015, opened 2021-04-01
updating error codes

thetif:
issue #2550 

Attempting to clean up our error codes to make them more in-line with the issue.

- updated errorHandler to throw 400 errors by default and handle different ways that it could be called
- created standardized error messages for common error codes
- all catches in the endpoints should call next and let the errorHandler middleware handle them
- removed all 500 error codes, except for the case of the database throwing a connection error
- all 401 errors should be a user not logged in and trying to perform an action that needs authentications
- all 403 errors should be a user that is logged in but doesn't have permission to perform and action
- all 415 errors are file related
- all other errors are 400 (we can discuss this some more and update it accordingly in another ticket)

- this is probably best verified by looking at the results of the endpoint tests and the openapi document

~- [ ] The experience passes a basic manual accessibility audit (keyboard nav, screenreader, text scaling) OR an exemption is documented~

~- [ ] Design has approved the experience~


tbolt (changes requested):
These updates are great. Nice work.

Will approve once the tests are all passing 👍

api/routes/apds/post.test.js
```diff
@@ -56,24 +58,38 @@ tap.test('apds POST endpoint', async endpointTest => {
...
+      getStateProfile.throws(error);
+      await handler(req, res, next);
 
-      test.ok(res.status.calledWith(500), 'HTTP status set to 500');
-      test.ok(res.end.calledOnce, 'response is terminated');
+      test.ok(next.calledWith(error), 'HTTP status set to 400');
     });
 
-    // Why is a 500 status being returned when the frontend sends malformed data?
+    // Why is a 400 status being returned when the frontend sends malformed data?
```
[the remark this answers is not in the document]
thetif: haha, obviously I was just doing a find

api/db/auth.js
```diff
@@ -89,7 +89,7 @@ const getUserAffiliatedStates = async (userId, { db = knex } = {}) =>
  * @returns {Object} { stateId: activities }
  */
 const getUserPermissionsForStates = async (userId, { db = knex } = {}) => {
-  const roles = await getRolesAndActivities();
+  const roles = (await getRolesAndActivities()) || [];
```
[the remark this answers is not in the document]
thetif: I added the empty arrays because Knex returns undefined if nothing is found in a query. So if I return an empty array, then the method that is using it, populateUser, won't fail if the user doesn't have any affiliations.

tbolt (approved):
👍 looks good. It's hard to keep 401s/403s straight and not mix them up ha

api/api.js
```diff
@@ -7,12 +7,14 @@ const cors = require('cors');
 const bodyParser = require('body-parser');
 const compression = require('compression');
 const uuid = require('uuid/v1');
+// const swaggerUi = require('swagger-ui-express');
```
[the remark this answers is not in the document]
thetif: Oh oops, I meant to take that out fully. Actually thought it might work since it was actually the send messages that ZAP didn't like.

thetif:
401 - who are you? - (not authenticated)
403 - nice try, Kevin - (authenticated, but without authorization to access)

cthulhuplus (approved):
LGTM

api/routes/apds/get.js
```diff
@@ -79,7 +85,10 @@ module.exports = (
       }
 
       logger.verbose({ id: req.id, message: 'apd does not exist' });
-      return res.status(400).end();
+      return res
+        .status(400)
+        .send({ error: 'apd does not exist' })
```
knollfear, line 90: should apd does not exist be a 404?

api/routes/openAPI/helpers.js
```diff
@@ -20,20 +9,32 @@ const arrayOf = schema => ({
   items: schema
 });
 
+const errorToken = jsonResponse({
+  type: 'object',
+  properties: {
+    error: {
+      type: 'string',
+      description:
+        'An string token indicating the error, which could be translated into a user-readable string for display by the client'
```
knollfear, line 18:
```suggestion
        'A string token indicating the error, which could be translated into a user-readable string for display by the client'
```

api/routes/users/get.js
```diff
@@ -10,7 +10,7 @@ module.exports = (app, { getAllUsers = ga, getUserByID = gu } = {}) => {
 
   app.get('/users/:id', can('view-users'), async (req, res, next) => {
     await getUserByID(req.params.id)
-      .then(user => (user ? res.send(user) : res.status(404).end()))
+      .then(user => (user ? res.send(user) : res.status(400).end()))
```
knollfear, line 13: This feels like it should have stayed as a 404

knollfear (commented):
This looks good.  I think there are a few minor changes that I don't quite understand.

api/routes/users/get.js
tbolt: Good point. It could be either right?

knollfear: If I read the code right, we sent a request to the DB for User with ID of 5 and the DB returned something falsey, that seems like 404.  If "can('view-users')" came back as false then I think 403 is the right answer.

thetif: @tbolt @knollfear 
So the team consensus is if we are searching for something in the database and it's not found, we return a 404? If someone is missing required input or sends in incorrect input is that a 400 or a 422?

eStar386:
Per issue #2962 design doesn't have to approve backend changes.

eStar386:
Tested with @knollfear we confirmed that the endpoints are functioning as specified.  Accessibility testing was not performed because it's a backend change.

eStar386 (approved):

NAretakis (approved):
This is good, so nice to get these updated.

PR #3022, opened 2021-04-05
Knollfear/2960 use okta username

knollfear:

- adds a column for the okta "login" value.  I believe this will fundamentally match to the EUA ID.
- fills in the username column for our existing users via both migrate and seed
- adds a down command to package.json for the API that will undo 1 migration

If you have a local copy that is basically running already then "docker-compose exec api npm run migrate" will get you where you need to be.

If you want to test the seed process you can go into your local database and run "update auth_affiliations set username = null;" and then "docker-compose exec api npm run seed".  

Alternatively, you can stop docker (docker-compose down), prune your volumes (docker volume prune) bring it all back up (docker-compose up) and finally run the migration and seed commands


thetif (approved):
Looks good

api/db/users.js
```diff
@@ -101,21 +101,35 @@ const getAllUsers = async ({
...
     client = oktaClient,
     populate = populateUser,
-    additionalValues
+    additionalValues,
+    db
   } = {}
 ) => {
-  const oktaUser = await client.getUser(id);
 
+  let oktaUser
```
tbolt, line 114: I think this can be a `const` since it's never reassigned to another type other than `object`. I'm not sure what all client.getUser can return though.

api/migrations/20210330134530_authAffiliationOktaUsername.js
```diff
@@ -0,0 +1,38 @@
...
+  });
+
+  await knex.schema.table('auth_affiliations', table => {
+    table.string('username').comment('username of user from authentication service');
+  });
+
+  const toUpdate = await knex('auth_affiliations').whereNull('username')
+
+  for (let i=0; i<toUpdate.length; i+= 1){
+    // eslint-disable-next-line no-await-in-loop
```
tbolt, line 16: I was curious why eslint doesn't like await in loops. Seems like an optimization one probably not necessary at this time, but good to know- https://eslint.org/docs/rules/no-await-in-loop

tbolt (approved):
Code looks good. Added some small comments.

api/db/users.js
knollfear: If I am honest, this is where I find the difference between const and let highly confusing.  In my mind const should be used when something will not ever change.  const ACTIVITY = "ACTIVITY", or const people = getPeople() make sense, because once I get that thing, I don't intend to change it and I signal that with a const initializer.  When I see `const people = getPeople()
people.push(newPerson)
` 
I feel like const is misplaced and should be let so that the next dev knows this will in fact change so they need to follow it through the code to know what it will be.

thetif: I think this is a valid use of `let` because you are setting it differently depending on the status of the boolean and you are doing something else in addition, otherwise you could set it using a ternary operation. You could avoid this by making this section a separate function and returning oktaUser from it.

beparticular:
Not sure there is anything I can verify on this. :)

eStar386 (approved):
Verified according to the steps for verification, the username has a value for existing users, a column with the login value.

PR #3034, opened 2021-04-12
Tbolt/2988 Use better error message when logging in with MFA and an expired password

### LINKED ISSUE #2988: [Dev]
When a user who has MFA enabled tries to login with an expired password, they do not immediately see the expired password warning. They are taken to the MFA verification page and then they get the error that the one-time password is incorrect. This needs to be updated to reflect the correct error. From looking at the flow of calls between Okta and our app, it will not tell you your password is expired until after you enter your one time password. This is annoying, but unavoidable, we should at least make sure the error message matches the issue.

Bug - 

### Steps to Replicate
1. Select a user enrolled into MFA who's password is expired
2. Enter username and password
3. Enter one-time password

### Expected outcome
- receive an error that you password is expired

### Actual outcome
- receive and error that your one-time password is wrong

### This task is done when…
This could include:
 - [ ] the Verify Your Identity Page shows the correct error

tbolt:
Resolves #2988 

- Updates the error message when logging in with MFA and an expired password. Previously if you had an expired password and logged in (with MFA enabled) it would give an error along the lines of "mfa code entered is invalid" when in fact the password was expired. Now it will display an error message "password has expired."

1. Checkout PR build
2. Use the test account `expiredmfa` to verify that after entering a valid MFA code that a password expired alert is shown
3. Optional - Test other login failure scenarios


thetif (approved):
Looks good

eStar386 (approved):
Passes, Test Step login with the users from keybase under Login Errors and/or one of the following a statecontractor, fedadmin,or self-created user accounts.

beparticular (approved):

jeromeleecms (approved):
User experience question - when I click on Okta, the page opens in the same window. Is this the expected behavior? Are we expecting for the user Click on Okta link -> Finish updating Okta password -> Go back to eAPD manually **vs** Click Okta link (opens in new tab, logs user out back to sign in screen) -> Finish updating Okta password-> Close Tab-> Be presented with the eAPD Login screen and login with the new password? Thoughts?

Overall, this works as intended, but wanted to get some insight about a good user flow (or if this is the best practice with this type of password reset action).

tbolt:
@jeromeleecms I discussed this a bit with @beparticular and I think it could go either way. Generally the default `<a>` behavior is to do a page navigation. Since it's the default it allows users freedom to choose how they will open the link. They can click/tap it as-is, right click and open in a new tab, use a hotkey/keyboard, middle-mouse click, etc. Plus it's added to the history stack so back/forward will work as expected. If we set it to open in a new tab we lose all of that and we may be going against their preferred browsing style. Personally I would like it to open in a new tab but I don't want to force my preferences on everyone. That's why in principle I usually recommend sticking with the default behavior. This might be a good case where we intentionally prescribe a certain experience though.

The two PDF documents opening in a new tab I think is a good exception because we're actually downloading the files and the browsers have built-in PDF viewers that can view them. I think we could even enhance those to be more representative that "these are PDF documents"

jeromeleecms:
Awesome - thanks for the thoughtful approach to implementation and helping me get up to speed with some of the background on this. Nice to see this finished and merged.

PR #3067, opened 2021-04-19
updated seeding script to handle no seeding needed

thetif:
The backend was failing because it was trying to insert an empty array, so I added a check so it only inserts if there is something to insert.

- added empty array testing

### DIFF
--- a/api/seeds/shared/roles-and-activities.js
+++ b/api/seeds/shared/roles-and-activities.js
@@ -32,8 +32,13 @@ const insertAndGetIDs = async (knex, tableName, values) => {
       return { name };
     });
 
-  await knex(tableName).insert(insert);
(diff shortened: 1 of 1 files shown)


api/seeds/shared/roles-and-activities.js
```diff
@@ -32,8 +32,13 @@ const insertAndGetIDs = async (knex, tableName, values) => {
       return { name };
     });
 
-  await knex(tableName).insert(insert);
-  logger.info(`Completed seeding the ${chalk.cyan(tableName)} table`);
+  if (insert && insert.length > 0) {
```
knollfear, line 35: if (insert && insert.length){

is the same I think.  This is might be a bit more readable though.

knollfear (approved):

knollfear (approved):

PR #3130, opened 2021-05-14
Knollfear/3029 eapd jwt

knollfear:

- exchanges an okta token for one that is generated by our system and includes our claims.

Login and use the system as normal.  There should be no difference to the user
When the user completes the log in they should receive a token from the me/jwToken endpoint (you can observe this in the network console)
The token stored in cookies/gov.cms.eapd.api-token should be a url encoded json object that includes the JWT.  It will look something like this  gov.cms.eapd.api-token:"{%22accessToken%22:%22<ACTUAL TOKEN>%22}"  everything between the %22 can be copy and pasted into jwt.io to verify it is our token and not okta's
The token is also present in all xhr requests in the authorization header.  This can also be extracted and checked against jwt.io.

- [ ] Develop has updated automated tests (and all tests are passing)
- [ ] QA has manually tested and created bugs for any issues found (beyond minor fixes)


thetif:
@knollfear you weren't here when we fixed this the first time, but we have to handle being able to retrieve images from the backend. I was doing it using a cookie version of the jwt so that it would be passed along on regular fetches too (not just axios calls). If you know of a better way to handle this that's fine, but we don't want to lose the functionality of images in the APD. I can walk you through how I enabled the images tomorrow if it helps.

Screen Shot 2021-05-18 at 5 39 15 PM

tbolt (approved):
Nothing sticks out to me but I'm not that familiar with jwts. Might be good to get an additional review on this.

knollfear:
> @knollfear you weren't here when we fixed this the first time, but we have to handle being able to retrieve images from the backend. I was doing it using a cookie version of the jwt so that it would be passed along on regular fetches too (not just axios calls). If you know of a better way to handle this that's fine, but we don't want to lose the functionality of images in the APD. I can walk you through how I enabled the images tomorrow if it helps.
> 
> Screen Shot 2021-05-18 at 5 39 15 PM

I looked into this and I think it is not a problem with my code exactly.  If I start an incognito session (or sufficiently clean up my cookies) and only log into my localhost version everything is fine and normal like I would expect.  However, if I then log into the staging site I am issued a cookie with the same name ("gov.cms.eapd.api-token"), but on the path "/apds/" instead of just "/".  Then when I switch back to my local host version the apd path cookie is preferred to the one with just "/".  However, if you just clear your cookies it all goes back to normal.  I think this is not a real issue for users as they will not be running the app on local host or switching between major versions.

thetif (approved):
I was able to pull this down and locally it worked after I deleted my cookies and cleared my cache. It also worked when I used incognito mode on the PR. We'll have to make sure it's not an issue in staging and prod once it's merged. But otherwise, I think it's ok.

eStar386:
Approved passes regression and accessibility testing. 
https://docs.google.com/spreadsheets/d/1Rt4CjcFh4TFphMPtQnEC-fSI6yYrK83k0Dg8o7K9fwM/edit?usp=sharing

NAretakis (approved):
Good to go, let's rock and roll.

beparticular:
@knollfear when I test, this is what I see. Not sure what it would look like if it was Okta's token. Is this correct? If so, lemme know and I'll approve. 

Screen Shot 2021-06-10 at 4 55 49 PM

knollfear:
@beparticular   That looks exactly right.  This is what we know about this user and what permissions they have.

beparticular (approved):

PR #3165, opened 2021-05-26
using  JWT_SECRET to sign JWT

### LINKED ISSUE #3128: [Dev] Create secret for signing JWT's
Now that we are issuing JWT's for authentication and authorization they need to be signed with a truly random value.

### This task is done when...
 - [ ] api/auth/jwtUtils.js => getSecret() returns a value from the environment (or other external source)
 - [ ] documentation exists for finding (and setting?) the variable for local environments.   Deployed environments can keep the secret secret for security reasons.
 - [ ] all auth continues to work

knollfear:
resolves #3128

**Description-**

This pull request converts the secret used for signing the JWT to use one from the environment instead of hard coded.

- previously issued JWT's will be invalid (unless you set your secret to 'ABCDEFGHIJKLMNOPQRSTUVWXYZ123456789'
- you MUST stop your docker containers
- you MUST then set a local variable called JWT_SECRET (export JWT_SECRET=<YOUR_SECRET>) 
- you should now restart docker
- **** The environment variable is pulled into docker when it starts.  Changing it while docker runs will have no effect.

nothing.  The JWT will perform exactly as before, simply have a more secure key.

   - This should be very isolated.  If the JWT is working, the secret must be correct. 

see manual verification below.  

Also, unit tests were created.

The most direct way to verify this change is:
- log in to the application
- look at your cookies for one with a key of gov.cms.eapd.api-token
- the cookie value is url encoded like this  {%22accessToken%22:%22<TOKEN_IS_HERE>%22}
- copy token to clipboard and go to JWT.io
- add your secret to 'VERIFY SIGNATURE' 
- verify that the signature, the last section if split by periods, has not changed


api/auth/jwtUtils.test.js
```diff
@@ -6,6 +6,8 @@ const jwt = require('jsonwebtoken');
 const sandbox = sinon.createSandbox();
 const mockVerifier = sandbox.stub();
 
+const JWT_SECRET = 'SOME_SECRET_VALUE_SHHHHH'
```
thetif, line 9: you might want to make this 
`const JWT_SECRET = process.env.JWT_SECRET || 'SOME_SECRET_VALUE_SHHHHH';`
in case they have JWT_SECRET set locally

thetif (approved):
looks good, one change to consider

api/auth/jwtUtils.test.js
knollfear: good call.  I made this update.

api/auth/jwtUtils.test.js
```diff
@@ -6,6 +6,8 @@ const jwt = require('jsonwebtoken');
 const sandbox = sinon.createSandbox();
 const mockVerifier = sandbox.stub();
 
+const JWT_SECRET = process.env.JWT_SECRET || 'SOME_SECRET_VALUE_SHHHHH'
```
tbolt, line 9: If JWT_SECRET isn't set from the env and defaults to that string it will be a sort of silent failure/security concern, wont it? Mostly just curious. I've always seen fallback values and never really question it but I wonder if not having a fallback and letting the system choke on startup would be preferred.

tbolt (approved):

PR #3297, opened 2021-07-20
Knollfear/3202 missing ffy alerts

### LINKED ISSUE #3202: [Feature/Bug] Temporary fix to make the user aware that things will be broken if there are no FFYs
Design task #3136

If a user unchecks all of the FFY options, then the export view and other data related to the FFY won't show up correctly.

When a user deletes the last FFY, they should see the FFY warning as usual. Then an alert should appear above the FFY checkboxes that says "At least one FFY must be selected to continue with your APD." (see the middle mock up in the Figma below.

If the user ignores this warning and continues on, then we should continue to warn them about this issue in case they are confused when they see strangeness. Show the alert  from the design system (content mocked up in the Figma file) at the top of every page

**Mock ups**
https://www.figma.com/file/6eVvo7JjvXiovGTR4BoVgK/CMS-eAPD-2020?node-id=3751%3A27887

**What is the goal or purpose of this ticket?**
Inform the user that not having an FFY selected will be bad.

### This task is done when…

This could include:

- [ ] the user it warned about an APD needing an FFY if they do not have any selected

knollfear:
resolves #3202

**Description-**
When a user chooses to have no Federal Fiscal Years (FFY) selected it can cause problems.  This adds an alert to each section letting the user know that they should select one

- Adds an alert to the APD overview if no FFY is selected
- Adds an alert to:
    - Key State Personnel
    - Results of Previous Activities
    - Activities
    - Activities Schedule
    - Proposed Budget
    - Assurances and Compliance
    - Executive Summary
    - Export
     

hopefully nothing else.  This should be an isolated change

unlikely to have side effects

- unchecked all FFY options and saw alert
- navigated to every page and confirmed alert existed
- checked an FFY and saw the alert disappear
- checked every other page and saw that it was gone

See above for repro steps
Note: The auto generated APD seems to have 2020 checked but the UI no longer allows for that field to be unchecked.  Please create a new APD for testing with.


beparticular:
@knollfear minor issues I wanted to ask about 
1) Can we change the text to say ""At least one FFY must be selected to continue with your APD. Select your FFY(s)." to show that multiple FFYs can be selected? I've updated Figma to match this request. 
2) I remember something about the width of the alert on the FFY page being an issue, but since the alert is wider on every other page, is there anything we can do to match the FFY page width to the rest of them? Thx!

web/src/reducers/apd.selectors.js
```diff
@@ -128,3 +128,12 @@ export const selectApdDashboard = createSelector([selectApds], ({ byId }) =>
 );
 
 export const selectLastSavedTimestamp = state => state.apd.data.updated;
+
+export const getYearsAlert = ({
```
tbolt, line 132: Doesn't look like this is used anywhere

tbolt (commented):
I was able to get into a weird spot with this. Here's what I did:
1. Created a new APD and unchecked all FFYs (I saw the modal alert and the subsequent banner alerts)
2. Went to the Export and Submit page, clicked on Continue to Review
3. This page didn't load instead it was just a blank white page (still had eAPD header and footer though)
4. Used the browser back button and I was taken back to the dashboard
5. Here I could re-open the APD but I didn't see any of the alerts

web/src/reducers/apd.selectors.js
knollfear: Nice catch.  This was a different approach I tried and abandoned as a bad idea.

web/src/containers/ApdExport.js
```diff
@@ -28,9 +29,9 @@ const ExportAndSubmit = ({ push: pushRoute, useParams }) => {
           </p>
           <Button
             size="big"
-            variation="primary"
+            variation={years.length > 0 ? "primary": 'danger'}
```
tbolt, line 32: I would suggest instead of showing this as the danger variant make the button disabled so it doesn't signify it can be clicked.

Something like
`disabled={years.length > 0 ? true : false}`

tbolt (approved):
Looks good and works as intended.

web/src/containers/ApdExport.js
knollfear: fixed in incoming push.  I looked at the code, not the docs and didn't realize disabled was a valid option.

thetif:
Looks like it's working as expected, tested by:

|Given|When|Then|Result|
|---|---|---|---|
|seeded APD|removed all FFYs|saw alert on every page and disabled print button|:white_check_mark: PASS|
|seeded APD with FFYs removed|exit to the dashboard and reopen|saw alert on every page and disabled print button|:white_check_mark: PASS|
|seeded APD with FFYs removed|copied url and opened it in another tab|saw alert on every page and disabled print button|:white_check_mark: PASS|
|new APD|removed all FFYs|saw alert on every page and disabled print button|:white_check_mark: PASS|
|new APD with FFYs removed|exit to the dashboard and reopen|saw alert on every page and disabled print button|:white_check_mark: PASS|
|new APD with FFYs removed|copied url and opened it in another tab|saw alert on every page and disabled print button|:white_check_mark: PASS|

beparticular:
@jeromeleecms currently this is showing on the main nav section of each page and working fine. It is not showing inside activities. 

Do you think it's important to also show this somewhere inside an activity? I do believe that there are areas where in the activity a field would be missing for a FFY that the user would be unaware of if they had all their FFYs unchecked.

jeromeleecms:
@beparticular Def. in line with that thinking. Taking it one step beyond that, it makes more sense to me to locate error messages that impact the apps functionality closer to the affected area(s). The way the app is designed, you may not see the error message at all if you use the left hand nav to move around the app.

For example, for "Key Personnel and Program Management" - by clicking on the nav bar, there is no note that you must select a FFY AND the app doesn't give an indication that some additional fields should be present:

image

If FFY's were selected, we would have seen the expanded fields tied to the cost of these key personnel, but as it stands, the user isn't the wiser that something should be present.

Can we use a similar design to when no content has been added to particular areas? For example, when no contractors exist in the activity: 

image

For this section, we would see this (Chrome tools edit version)

image

Or for contractors within the APD section:

image

Less concerned about the specific language that I put in the warning box, but more of a way to have the app communicate as many ways as possible that something is not right. If we want to stage this ticket, I would be OK with:

1) Adding the language at the activity level and moving along...
2) Then in new ticket, add an additional check that warns users that FFY(s) have not been selected and there is something missing from the application for this particular part of the experience.

jeromeleecms:
Per standup - let's just add to activity level and move along. We can solve the broader problem when we implement validation.

tbolt:
@jeromeleecms @beparticular @thetif  This is updated now to have the alert on all activities pages 👍 

Let me know if you run into any issues with it.

beparticular (approved):

jeromeleecms (commented):
@tbolt I don't see this in staging or in any other PRs, but the executive summary seems to have picked up the HTML tags...

image

Is this related to this PR or is this a separate bug?

tbolt:
@jeromeleecms I would file that as a separate bug. I am seeing that in other builds as well.

jeromeleecms (approved):
Rest of the team clarified that the bug that was identified does not have anything to do with this PR, so this is good to go. Bug has a separate ticket.

PR #3399, opened 2021-08-13
3342 - Move Cypress tests

### LINKED ISSUE #3342: [Feature] Research the most efficient, clean, and easy way to spin up the cypress tests
In the current cypress setup, if cypress needs to be able to change the database (e.g. remove affiliations added in tests), then you have to create an endpoint in the api. Determine a cleaner way for cypress to be able to interact with the database without having to mock the api call that adds things to the database, add endpoints to the api, or add database modules to the web container.

**What is the goal or purpose of this ticket?**
Make cypress easy to run without a lot of overhead and without hacks

### This task is done when…

- [ ] the cypress set up is easy for the developers to use and easy to add to CI/CD

knollfear:
resolves #3342

**Description-**

In order to make the cypress tests more robust and not make API changes solely for the sake of testing I have moved the cypress tests out of the web folder into an "integration tests" folder.  It currently only holds cypress tests, but could hold Postman or other kinds of tests.  The folder has its own package.json so you will need to run npm install in that directory before executing the tests.  

- the location of the cypress tests within the code base
- how the cypress tests interact with the database
- how cypress tests are run (still an npm command, but in integrationTests now)
- removed cypress api endpoints
- fixed missing username values after seed script runs

THis should not affect deployed code as it just relocates test code.

- Cypress tests run successfully and repeatably.

(assuming containers are running as normal)
1. cd to integrationTests
2. npm install (maybe npx install too idk)
3. npm run cy:run
4. cypress tests should execute and succeed.
5. note that cypress test runs will reset your local DB to default so just be aware of that.


integrationTests/cypress-setup.sh
```diff
@@ -0,0 +1,8 @@
+#!/usr/bin/env sh
+
+export NODE_ENV=development
+
+docker-compose exec api npm run migrate
```
thetif, line 5: could we do this more like the endpoint-test/endpoint.sh and do a 
```
docker-compose up -d
```
before we run these other commands, just in case the api isn't up

integrationTests/knexfile.js
```diff
@@ -0,0 +1,13 @@
+module.exports = {
+  development: {
+    client: 'postgresql',
+    connection: {
+      host: 'localhost',
+      database: 'hitech_apd',
```
thetif, line 6: could we change the database to be test specific? like hitech_apd_test or hitech_apd_cypress?

web/.eslintrc.json
```diff
@@ -1,6 +1,6 @@
 {
   "parser": "babel-eslint",
-  "extends": ["airbnb", "prettier", "plugin:jsx-a11y/strict", "plugin:cypress/recommended"],
+  "extends": ["airbnb", "prettier", "plugin:jsx-a11y/strict"],
```
thetif, line 3: could we create a .eslintrc.json file for integrationTests with that plugin in it?
```
{
  "parser": "babel-eslint",
  "extends": ["airbnb", "prettier", "plugin:cypress/recommended"],
  "rules": {
    "arrow-body-style": "off",
    "no-param-reassign": [2, { "props": false }],
    "react/forbid-prop-types": ["error", { "forbid": ["any"] }],
    "react/jsx-filename-extension": 0,
    "react/jsx-fragments": ["error", "element"],
    "react/jsx-props-no-spreading": 0,
    "react/no-array-index-key": "warn"
  },
  "env": {
    "browser": true
  }
}
```

integrationTests/package.json
```diff
@@ -0,0 +1,27 @@
+{
+  "name": "cypress_tests",
+  "version": "1.0.0",
+  "description": "Cypres tests for the EAPD tool",
```
thetif, line 4: Cypress

integrationTests/package.json
```diff
@@ -0,0 +1,27 @@
...
+  "main": "index.js",
+  "scripts": {
+    "cy:open": "./cypress-setup.sh open",
+    "cy:run": "./cypress-setup.sh run",
+    "cy:run:ci": "cypress run"
+  },
+  "author": "",
+  "license": "ISC",
+  "dependencies": {
```
thetif, line 13: I think all of these could be devDependencies, right?

web/Dockerfile
```diff
@@ -1,12 +1,9 @@
-FROM cypress/base:14
+FROM node:14
 
-RUN npm i -g npm@^7
+RUN npm i -g npm@^6
```
thetif, line 3: why npm 6 and not npm 7?

thetif (approved):
Tests running in the new place, looks good!

integrationTests/knexfile.js
knollfear: not without restarting the containers.  The API is already up and connected to a DB before we get going.

web/Dockerfile
knollfear: I reverted the whole thing to an older version.  I can move it back to NPM 7.

integrationTests/package.json
knollfear: I guess.  It won't ever really be built so I am not sure if it will make much of a difference.  I will move them all.

integrationTests/cypress-setup.sh
knollfear: sure, I can do that.  It seems to do nothing if the containers are up already.

web/.eslintrc.json
knollfear: I have the linter running. It shows quite a few issues.  I will fix those in a clean commit.

integrationTests/package.json
knollfear: fixed

integrationTests/cypress-setup.sh
knollfear: Doing this causes the containers to come down at the end of the tests which is a different kind of annoying.

integrationTests/.gitignore
```diff
@@ -0,0 +1,3 @@
+cypress/screenshots/*
+cypress/videos/*
+cypress.env.json
```
tbolt, line 3: Missing newline

integrationTests/knexfile.js
tbolt: that is a good idea though but it does seem like it would take a bit of work and/or duplication to create another seeded db + have the api connect to it. maybe a future enhancement?

tbolt (commented):
Code seems good. I am getting a handful of failures with the testing. Most are throwing this error-
```
1) Affiliation
       Logging in with a user that does not have an affiliation
         should let the user cancel selecting an affiliation:
     CypressError: `cy.type()` cannot accept an empty string. You need to actually type something.

https://on.cypress.io/type
```

Screen Shot 2021-08-19 at 10 09 49 AM

api/seeds/shared/set-up-users.js
```diff
@@ -89,7 +89,7 @@ const createUsersToAdd = async (knex, oktaClient) => {
       role_id: stateAdminRoleId,
       status: 'approved',
       updated_by: 'seeds',
-      username: 'em@il.com'
+      username: 'em@il.com',
```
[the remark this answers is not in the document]
knollfear: idk.  I tend to add them because of my python experience.  It causes me problems though since they are not valid in JSON.

tbolt (approved):

api/seeds/shared/set-up-users.js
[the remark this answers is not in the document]
thetif: @tbolt I think you are getting those errors because you haven't added your cypress.env.json file in the integrationTests folder. I have a copy of the file in keybase

PR #3408, opened 2021-08-18
Knollfear/3386 default select values

knollfear:
resolves #3386

**Description-**

**This pull request changes...**
1) In the FFP, Federal-State Split dropdown on the Budget and FFP page
    State split is initially set to 100% state and 0% federal.  If users make a selection then the state will pay less than the full cost.
2) In the non-personnel cost category, in an Activity on the State staff and expenses page
    When a non-personnel cost is added it will have no category and display as "Category not selected" in the list 
3) In the state and federal admin panels when a user is granted access (either new access or restoring revoked access).
I did not stop the form from submitting when no choice was made.  It will just fail and not approve the user.

**This pull request also touches…**

- possible side effects of this change

**This pull request was tested in the follow ways…**

- manually checked that newly created objects show the "make a selection" option.

**Steps to manually verify this change...**

1. add an activity to an APD
2. edit the apd
3. go to "Budget and FFP"
4. look for federal state split to confirm it is 100% state
5. go to "State Staff and Expenses"
6. click "Add State Expense"
7. confirm no category is chosen

1. log in as a state admin
2. go to the state administrator portal
3. on requests or inactive click to grant access
4. no role should be selected
5. on the active tab if you click edit role, a role SHOULD be selected

### This pull request is ready to review when...

- [ ] Automated tests are updated (and all tests are passing)
- [ ] The change has been documented
- [ ] Associated OpenAPI documentation has been updated
- [ ] Changelog is updated as appropriate
- [ ] The experience passes a basic manual accessibility audit (keyboard nav, screenreader, text scaling) OR an exemption is documented

### This pull request can be merged when…

- [ ] Code has been reviewed by someone other than the original author
- [ ] QA has verified the accessibility and functionality related to the change
- [ ] Design has approved the experience
- [ ] Product has approved the experience

PR #3411, opened 2021-08-19
Knollfear/3408 add aud to jwt

knollfear:
resolves #3408

**Description-**
JWT's produced by staging were fooling the front end of production and vice versa.  While the front end can not fully validate the token (it lacks the private key) it can read the payload and ensure that it is for the right audience.  This PR adds an audience to the token with a value of process.env.NODE_ENV.  This value should match between the front and back end.

- adds a better value for aud to the JWT
- checks that the JWT is for the right aud before accepting it.

should just affect the audience value of a token

should not be any I can imagine

- updated test cases to show that the token value snapshot has changed

    **** NOTE THESE ARE NOT VALID TOKENS THEY WERE CREATED BY HAND AND ARE UNSIGNED ****

Using the current version of the EAPD software set your cookie for key gov.cms.eapd.api-token to this:

{%22accessToken%22:%22eyJhbGciOiJIUzI1NiIsInR5cCI6IkpXVCJ9.eyJ1c2VybmFtZSI6InN0YXRlYWRtaW4iLCJpYXQiOjE2MjkzODM3NjEsIm5iZiI6MTYyOTM4Mzc2MSwiZXhwIjoyNjI5NDI2OTYxLCJhdWQiOiJFQVBEIn0.WlVt9mhbUxEdyIAiFQlF_gVTgDKKcPpHTqH8q-vslSs%22}

refresh the page and you will be signed in, but you won't have a state and fetching APD's will fail.

Checkout this branch and refresh again to see that you are no longer logged in.

If the audience value of the token is correct then the front end will still allow you to fake being logged in.  You can prove that with this token (set your cookie value to this):

{%22accessToken%22:%22eyJhbGciOiJIUzI1NiIsInR5cCI6IkpXVCJ9.eyJ1c2VybmFtZSI6InN0YXRlYWRtaW4iLCJpYXQiOjE2MjkzODM3NjEsIm5iZiI6MTYyOTM4Mzc2MSwiZXhwIjoyNjI5NDI2OTYxLCJhdWQiOiJkZXZlbG9wbWVudCJ9.komJMh6sbuvakmvbRrviLtUbQ5N7BMlM0KuHEJRnLVQ%22}

that will let you be logged in because it has the correct value for audience.  Again, the back end will not serve any requests though because the token is invalid.


api/auth/jwtUtils.js
```diff
@@ -84,7 +84,7 @@ const defaultOptions = {
   algorithm: 'HS256',
   expiresIn: '12h',
   notBefore: '0',
-  audience: 'eAPD',
+  audience: process.env.NODE_ENV,
```
thetif, line 87: could we do something like
`eAPD - ${process.env.NODE_ENV}`

jeromeleecms (approved):
if Dev team is good, I'm good with this.

PR #3460, opened 2021-09-03
APD migration from postgres to mongo

### LINKED ISSUE #3112: [Dev] Create script to migrate APDs from Postgres to Mongo
Create a Postgres migration that will get all of the APDs and convert them into Mongo APDs. Test this using a dump from the staging and production databases to ensure that potential errors are caught. Back up database before this issue is merged into staging and released into prod.

### This task is done when…
 - [ ] databases are backed up
 - [ ] a migration script has been created to copy APD data from Postgres and save it to Mongo
 - [ ] migration script has been tested against actual data dumps from staging and production

thetif:
resolves #3112 

**Description-**
Using the mongoose migrator that already exists in the system. I created a migration that would query postgres for all of the current APDs in the database and then convert it to a format that works for mongoose and save them.

- updates the blank migration to migrate all of the APDs from postgres to mongoose

1. download the PR
2. run `docker-compose -f ../docker-compose.yml exec -e MONGO_URL=mongodb://mongo:cms@mongo:27017/eapd api npm run mongoose-migrate`
3. it should run without issues and the APDs you have in postgres should be copied over to mongo


knollfear (commented):
This is a lot of code to read, but I gave it a quick test and found a problem.  When I click on an APD I am taken to a page with the right navigation on the left, but the center part where the APD info should be did not load.  If I switch to another section, and then come back, it is all there, but it needs to be there on initial load.  I also got an error that in web/src/containers/ApdApplication.js apdID is a required number prop, but is now a string.

PR #3957, opened 2022-04-08
Tbolt/3498 Other state expenses validation

### LINKED ISSUE #3498: [Feature] Update Other State Expenses
Remove the red warning when there is no Other State Expenses added, state expenses are not required.

If there are no state expenses, display this message above the Add non-personnel cost button

---

Add in-line validation to Other State Expenses. The validation should appear after the user hits the Add (previously Done) button. If there are validation errors, the errors should display, nothing should be saved, and it should stay on the form until the errors are fixed.

---

Update the view of added Other State Expenses

### This task is done when…
- [ ] add message above the Add non-personnel cost
- [ ] remove red warning
- [ ] add in-line validation
- [ ] update to match Figma

tbolt:
Resolves #3498 

**Description-**
* Updates other state expenses section to use new form validation tools


integrationTests/cypress/integration/02-apd/01-apd-basics.spec.js
```diff
@@ -623,18 +623,76 @@ describe('APD Basics', { tags: ['@apd', '@default'] }, () => {
       cy.findByRole('button', { name: /Cancel/i }).click();
 
       cy.get('.form-and-review-list')
-        .contains('Other state expenses have not been added for this activity.')
+        .contains('Add other state expense(s) for this activity.')
         .should('exist');
 
       cy.findByRole('button', { name: /Add State Expense/i }).click();
 
+
```
thetif, line 631: lots of empty space

integrationTests/cypress/page-objects/activities-state-staff-expenses-page.js
```diff
@@ -133,22 +132,17 @@ class ActivitiesStateStaffExpensesPage {
...
-      .next()
-      .findAllByRole('button', { name: /^Edit/i })
-      .eq(expenseIndex)
-      .click();
-
-    cy.get('[name="category"]').select(category);
-
-    cy.get('[name="desc"]').clear().type(desc, { delay: 1 });
-
+  fillExpense = (years, expenseIndex, category, costs, desc) => {
```
thetif, line 135: should probably switch this to an object so it's easier to call

thetif (approved):

mirano-darren:
Well done!
| Given  | When | Then | Result |
| ------------- | ------------- | ------------- | ------------- |
| other state expenses subform | form is filled out with valid fields | save button is clickable and saves | PASS ✅ |
| other state expenses subform | right after pressing "Add State Expense" and no fields are touched | save button is disabled | PASS ✅ |
| other state expenses subform | tab navigating through the form, not filling anything out | validation error appears on all fields | PASS ✅ |
| an existing other state expenses subform | form is edited, click cancel | changes are not saved  | PASS ✅ |
| an existing other state expenses subform | form is edited, click save| changes are saved  | PASS ✅ |
| an existing other state expenses subform | a field is edited to be invalid | inline validation triggers and save button is disabled| PASS ✅ |

mirano-darren (approved):

jeromeleecms (approved):
Besides the system wide bug on subforms filed for #4047, this works as advertised. Gtg from my end.

beparticular (changes requested):
Steps: Expected: 
Don't select an expense category
Fill in all the other fields, can't save ofc
Go back to the expense category and select an option
Unexpected: the validation error is still there after I've chosen an option. 

If I tab or click out of the field it updates and there is no longer an error, but the expected behavior would be that selecting an option would fix the problem and the error, rather than selecting an option and needing to navigate to another field.

tbolt:
@beparticular My understanding was that the team decided to only trigger validation "onBlur." Which means that even though you select an item from the dropdown you still have the field "focused" and only when you exit (tab or mouse click away from it) that would trigger the validation. We can revisit and make the call to have some exceptions because I agree the current behavior doesn't feel expected. We could switch to doing validation anytime a change is made for all inputs/fields. This would fix the issue you pointed out but also impact all other inputs. So for text inputs, as soon as you typed one character we would validate.

beparticular (approved):

itsonlydio (commented):
The Other state expenses field validations works!

PR #4158, opened 2022-06-21
Tbolt/3445 hit hie mmis validation updates

### LINKED ISSUE #3445: Create validation dependency for HIT|HIE|MMIS fields on APD Overview
These fields will only show their dependency if not completed during the validation check before submission. These only apply to HITECH APDs as these exact fields wouldn't be needed in the future for other APD funding sources. 

*These validations will be triggered by the error check that is currently being tested. #3488

|Field|Content|Consideration|
|---|---|---|
|HIT|Provide a summary of HIT-funded activities.|This should only be required when the APD is a HITECH APD and has a Program Admin Activity 1. Therefore this will currently be required for all APDs.|
|HIE|Provide a summary of HIE-funded activities.|In theory, an author will know that there will be HIE activities in the APD and should initially fill this out. If they don't fill this out, AND an HIE activity is added to this ADP, then this field becomes required. This would only be highlighted as part of the error check at the end of the APD since we don't want to send someone to another page in the middle of creating their activity.|
|MMIS|Provide a summary of MMIS-funded activities.|In theory, an author will know that there will be MMIS activities in the APD and should initially fill this out. If they don't fill this out, AND an MMIS activity is added to this ADP, then this field becomes required. This would only be highlighted as part of the error check at the end of the APD since we don't want to send someone to another page in the middle of creating their activity.|

**Mockup:**. (How error text would fall, underneath the field.)

### This task is done when...
- Overview fields for funding sources have inline validation

tbolt:
Resolves #3445 

### Description
Adds validation to HIT Overview, HIE Overview, and MMIS Overview. These are dependent on which funding types are selected in the activities. For example, if there are no activities with MMIS funding type, then there will not be any validation and the field may be left empty.  By default each APD starts with a Program Administration activity that is funded by HIT so given that, it is always required. 

### Signifcant changes or possible side effects
This PR should not introduce any side effects or changes outside of the scope of the original story. 

### Steps to manually verify this change

1. Login as a user who can view APDs and create a new one
2. Verify that HIT Overview is a required field
3. Verify that HIE Overview and MMIS Overview are not required fields
4. Create a new activity with HIE funding type, verify that HIE Overview is now a required field
5. Repeat the above for MMIS

This pull request is ready to review when
- [x] Pull request has been labeled, if applicable with feature, content, bug, tests, refactor

- [x] Code has been reviewed by someone other than the original author


babel.config.js
```diff
@@ -0,0 +1,16 @@
+const presets = [
```
thetif, line 1: does this work to share babel across all the folders?

thetif (approved):

mirano-darren:
Nice job! Looks good to me
| Given  | When | Then | Result |
| ------------- | ------------- | ------------- | ------------- |
| HIT overview field | left blank/filled out  | if blank: validation message appears; if filled out: validation message does not appear  | PASS ✅ |
| HIE/MMIS overview field  | there are no HIE/MMIS activities  |  validation doesn't trigger  | PASS ✅|
| HIE/MMIS overview field  | there are HIE/MMIS activities created |  validation triggers, and works as intended  | PASS ✅|

beparticular (approved):

babel.config.js
tbolt: I'm not sure, my editor was complaining that I didn't have a babel config at the root of the project though. I added it thinking it wouldn't impact anything, do you think it will?

jeromeleecms (approved):
Looks good!

PR #4263, opened 2022-08-08
Adding Babel to API

### LINKED ISSUE #4259: [Maintenance] Add babel to api
### Description and related issues
We need to update the API package to use babel so that we can start using new modules and we can integrate with our Common package.

### This task is done when…

- [ ] move babel configs from web/e2e/common into the root directory so that they can all be shared
- [ ] add babel configs to override main configs for api

thetif:
Resolves #4259 

### Description
We needed to add babel to the api so that we can use the common package and so that we can upgrade other packages that use es6.

### Significant changes or possible side effects
This should only significantly change the build process. Once the app is built, as you are able to login and see APDs it should be fine

### This pull request is ready to review when

- [x] Pull request has been labeled, if applicable with feature, content, bug, tests, refactor

- [x] Code has been reviewed by someone other than the original author
- [ ] QA has verified the functionality related to the change
- [ ] QA has verified the accessibility using tools such as screen readers and WAVE (accessibility evaluation tool)


api/unit-test.sh
```diff
@@ -3,16 +3,18 @@
 export NODE_ENV=test
 
 docker-compose -f ../docker-compose.endpoint-tests.yml -p api up -d
+sleep 5
```
tbolt, line 6: why does this need to be added?

tbolt (approved):

api/unit-test.sh
thetif: I don't know that it _has_ to be added, but sometimes it seems like the servers aren't fully up before it moves to the next step, so it's kind of insurance

mirano-darren:
@thetif I am getting this error message when running the seed command.
image

thetif:
@mirano-darren can you give it a try now, I think I resolved the issue.

mirano-darren (approved):

cthulhuplus (approved):
DevOps stuff looks good to me

PR #4345, opened 2022-10-05
Tforkner/3745 api budget

### LINKED ISSUE #3745: [Feature] Integrate budget calculations to the backend, store budget in APD
Currently, the APD budgets are being calculated in redux. However, as we move to mongo and multi-user editing. It makes more sense to do this calculations on the backend and store them in mongo with the APD, so that they don't have to be recalculated each time the APD is opened. With multiple users editing an APD at the same time, there is a good chance that the budget might get out of sync too. It will be safer and faster to have the calculations on the backend and update them as necessary.

### This task is done when…
- [ ] budget calculations are moved from redux to the backend
- [ ] updating an APD will trigger the budgets to be recalculated (an attempt should be made to only run the calculations when something changes related to the budgets)
- [ ] budgets are added to the APD schema
- [ ] budgets are returned along with other things that are changed on patch
- [ ] the frontend is updated to get budget information from the APD instead of the budget redux
- [ ] all tests are passing

thetif:
Resolves #3745 
Fixes issue with test failing

- [x] Automated tests are updated (and all tests are passing)


tbolt (approved):
This is fine, but here's how I thought we could fix it for good-
https://github.com/CMSgov/eAPD/pull/4327/files#diff-36bec2cdb3879a7387c4c2556c438301d4db1089192ed6b6fce72dc7cda6041e

tbolt (approved):

PR #4378, opened 2022-10-25
bbrooks/Packer-Validation-Patch-20221025

cthulhuplus:
Resolves #

### Description
Previous fix to Packer validation missed the Launch Darkly variables in the CircleCI Packer commands, this removes them which should result in the validations and AMI builds to function properly again.

### Automated test cases written

| Given | When | Then | Type (jest, tap, cypress) |
| ----- | ---- | ---- | ------------------------- |

### Steps to manually verify this change

1.
2.
3.

- [ ] New automated test cases are documented above
- [ ] Pull request has been labeled, if applicable with feature, content, bug,
      tests, refactor
- [ ] The experience passes a basic manual accessibility audit (keyboard nav,
      screenreader, text scaling) OR an exemption is documented

### This pull request is ready to review when the QA has

- [ ] Verified the functionality related to the change
- [ ] Verified that the change works with Narrator on Windows
- [ ] Verified that the change works with VoiceOver on Mac
- [ ] Verified all updated pages with the WAVE tool
- [ ] Verified tab and keyboard navigation functionality

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -355,7 +355,6 @@ jobs:
               --var preview_mongo_admin_url=$PREVIEW_MONGO_ADMIN_URL --var preview_mongo_url=$PREVIEW_MONGO_URL --var preview_database_url=$PREVIEW_DATABASE_URL \
               --var preview_okta_domain=$PREVIEW_OKTA_DOMAIN --var preview_okta_api_key=$PREVIEW_OKTA_API_KEY --var environment=Preview \
               --var preview_tealium_tag=$PREVIEW_TEALIUM_TAG --var preview_tealium_env=$PREVIEW_TEALIUM_ENV \
-              --var preview_ld_api_key=$PREVIEW_LD_API_KEY --var preview_ld_client_id=$PREVIEW_LD_CLIENT_ID \
               preview-ami.pkr.hcl
   
   packer validation tests:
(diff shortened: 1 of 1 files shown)


.circleci/config.yml
```diff
@@ -355,7 +355,6 @@ jobs:
               --var preview_mongo_admin_url=$PREVIEW_MONGO_ADMIN_URL --var preview_mongo_url=$PREVIEW_MONGO_URL --var preview_database_url=$PREVIEW_DATABASE_URL \
               --var preview_okta_domain=$PREVIEW_OKTA_DOMAIN --var preview_okta_api_key=$PREVIEW_OKTA_API_KEY --var environment=Preview \
               --var preview_tealium_tag=$PREVIEW_TEALIUM_TAG --var preview_tealium_env=$PREVIEW_TEALIUM_ENV \
-              --var preview_ld_api_key=$PREVIEW_LD_API_KEY --var preview_ld_client_id=$PREVIEW_LD_CLIENT_ID \
```
thetif, line 358: why did you need to take these out? I know we won't need LD_CLIENT_ID on the server, but we do need LD_API_KEY

.circleci/config.yml
```diff
@@ -388,8 +387,7 @@ jobs:
...
```
thetif, line 391: same as above

.circleci/config.yml
cthulhuplus: Launch Darkly isn't installed/enabled until provisioning. Launch Darkly is not set up on the AMI, so having the variables is unnecessary. We could add them to the AMI, but it would be a redundant and it still wouldn't be used until provisioning.

PR #4644, opened 2023-03-28
[Maintenance] fixing budget migration

thetif:
Resolves budget migration issue

### Description
I changed up the migration so that it does HITECH and MMIS separately because it seems to be having an issue setting the correct discriminator. I also changed the discriminators in the schemas to use constants.

- [ ] Pull request has been labeled, if applicable with feature, content, bug,
      tests, refactor

### DIFF
--- a/api/migrations/20230207224139_mongoose-adding-funding-category.js
+++ b/api/migrations/20230207224139_mongoose-adding-funding-category.js
@@ -1,6 +1,6 @@
 import loggerFactory from '../logger/index.js';
 import { setup, teardown } from '../db/mongodb.js';
-import { APD, Budget } from '../models/index.js';
+import { APD, getApdModel, getBudgetModel } from '../models/index.js';
 import {
   APD_TYPE,
   FUNDING_CATEGORY_TYPE,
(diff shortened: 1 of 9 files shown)


Sun-Mountain (approved):
🚀

api/migrations/20230207224139_mongoose-adding-funding-category.js
```diff
@@ -133,9 +134,9 @@ export const up = async () => {
           ...apd,
           apdType
         });
-        return getBudgetModel(apdType).replaceOne(
+        return Budget(apdType).replaceOne(
           { _id: apd.budget },
-          { ...budget }
+          { ...budget, __t: BUDGET_TYPE[apdType] }
```
tbolt, line 139: is this why it is/was failing?

tbolt (approved):

api/migrations/20230207224139_mongoose-adding-funding-category.js
thetif: I think so because I think it wasn't properly setting the discrimination so it was only saving what is in the base budget

